PR #10, opened 2018-03-21
Adding return after rejects in a few places. Implementing Ghostery side of GH-885. Fix for GH-871.

zarembsky:


src/classes/EventHandlers.js
```diff
@@ -378,15 +378,21 @@ class EventHandlers {
...
 			tab_host = tabInfo.getTabInfo(tab_id, 'host');
 			fromRedirect = globals.REDIRECT_MAP.has(request_id);
-			block = this._checkBlocking(app_id, cat_id, tab_id, tab_host, page_url, request_id);
+			const retVal = this._checkBlocking(app_id, cat_id, tab_id, tab_host, page_url, request_id);
+			block = retVal.block;
+			/* eslint prefer-destructuring: 0 */
+			if (retVal.ss_unblock) {
+				// The way to pass this flag to Cliqz handlers
+				details.ghosteryWhitelisted = true;
+			}
```
jsignanini: @zarembsky won't this fail when `this.policy.shouldBlock(...)` returns just true or false?

src/classes/EventHandlers.js
```diff
@@ -378,15 +378,21 @@ class EventHandlers {
...
 		let block;
+		let ss_unblock;
 		if (bug_id) {
 			app_id = bugDb.db.bugs[bug_id].aid;
 			cat_id = bugDb.db.apps[app_id].cat;
 			incognito = tabInfo.getTabInfo(tab_id, 'incognito');
 			tab_host = tabInfo.getTabInfo(tab_id, 'host');
 			fromRedirect = globals.REDIRECT_MAP.has(request_id);
-			block = this._checkBlocking(app_id, cat_id, tab_id, tab_host, page_url, request_id);
+			const retVal = this._checkBlocking(app_id, cat_id, tab_id, tab_host, page_url, request_id);
```
christophertino: `const {block, ss_unblock} = this._checkBlocking(app_id, cat_id, tab_id, tab_host, page_url, request_id);`

src/classes/EventHandlers.js
```diff
@@ -767,10 +773,10 @@ class EventHandlers {
...
 	 * @param  {number} 	tab_id			tab id
-	 * @param  {string} tab_host		tab url host
-	 * @param  {string} page_url		full tab url
+	 * @param  {string} 	tab_host		tab url host
+	 * @param  {string} 	page_url		full tab url
 	 * @param  {number} 	request_id		request id
-	 * @return {boolean}
+	 * @return {Object}		{block, ss_unblock}
 	 */
 	_checkBlocking(app_id, cat_id, tab_id, tab_host, page_url, request_id) {
```
christophertino: this should have an object return, correct? looks to just be returning `block` at the moment, instead of `return {block:true, ss_unblock:false}` etc

See line 793

src/classes/Policy.js
```diff
@@ -100,35 +100,42 @@ class Policy {
 	 * @param  {number} tab_id 		tab id
 	 * @param  {string} tab_host 	tab url host
 	 * @param  {string} tab_url 	tab url
-	 * @return {boolean}
+	 * @return {Object} 			{block, ss_unblock}
 	 */
 	shouldBlock(app_id, cat_id, tab_id, tab_host, tab_url) {
 		if (globals.SESSION.paused_blocking) {
-			return false;
+			return { block: false };
```
christophertino: we should return `{block, ss_unblock`} in all cases to prevent `undefined` errors

src/classes/EventHandlers.js
zarembsky: block IS object, as it is returned by shouldBlock, which returns object.

src/classes/Policy.js
zarembsky: You mean if we ever call shouldBlock elsewhere (but we don't)

zarembsky: I'll make the requested changes though.

src/classes/EventHandlers.js
zarembsky: The problem with the code `const {block, ss_unblock}` =  is that block should be visible out of scope if(bug_id) { }. We can get away with extra assignment, like:
const {block1, ss_unblock} = ...
block = block1;, but is it better?

PR #18, opened 2018-03-29
remove verbose webpack output

trickpattyFH20:


package.json
```diff
@@ -10,7 +10,7 @@
   "license": "MPL-2.0",
   "main": "webpack.config.js",
   "scripts": {
-    "build.dev": "webpack --mode development --display-error-details --progress --verbose --env.dev",
+    "build.dev": "webpack --mode development --display-error-details --progress --env.dev",
```
chrmod: how about adding `--verbose` to `build.watch`?

trickpattyFH20: sure sounds good. I'll add it 👍

PR #26, opened 2018-04-03
upgrade webpack and fix watch scss bug

trickpattyFH20:


trickpattyFH20:
lint watching scss files is fixed 👍 @jsignanini

PR #32, opened 2018-04-10
Automated extension builds

sammacbeth:
Automated build of the extension on Jenkins. Contains a Dockerfile in-which we can build the extension and a Jenkinfile to describe the build steps.

The build is done using `moab makezip`, which builds and packages the extension for all platforms. With the firefox build, we can also make a signed xpi for the Cliqz browser.

Next steps: make signed builds for other browsers. We may be able to add linux compatibility to moab for this.


sammacbeth:
Added packaging and signing of the chrome build too. Current it is just signed with the CI key, but I think this should be fine for betas.

PR #45, opened 2018-04-23
workaround for travis builds failing

trickpattyFH20:
workaround for browser-core integrity check


trickpattyFH20:
#40 should be synced with develop after this is merged in order for the build to pass

PR #65, opened 2018-05-17
feature/yarn

trickpattyFH20:
- `nvm install` will now resolve to node LTS
- CI environments now all using node LTS and yarn
- readme updated for yarn

TODO 
- local builds should migrate to yarn
- MOAB should migrate to yarn


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
jsignanini: @trickpattyFH20 let's replace this with the specific latest lts `lts/carbon` to be sure we're all on the same version.

.travis.yml
```diff
@@ -1,15 +1,15 @@
 language: node_js
 node_js:
-  - "9"
+  - "lts/*"
```
jsignanini: @trickpattyFH20 same as above comment here.

PR #91, opened 2018-06-08
Sign the XPI for Android Cliqz

sharath-cliqz:

PR #92, opened 2018-06-08
Sign the XPI for Android Cliqz - FIX

sharath-cliqz:
FIX :: Was picking the file from the wrong location.


chrmod:
follows up https://github.com/ghostery/ghostery-extension/pull/91

PR #97, opened 2018-06-11
Android browser develop

mai-cliqz:


luciancor:
@christophertino can you guys please take a look at this PR? This is the Android Panel which should appear for the Firefox for Android users

christophertino:
@mai-cliqz this looks solid.  I think we could import some of the utils like msg.js and utils.js from the panel folder like we do for setup, rather than having duplicated code.  See [here](https://github.com/ghostery/ghostery-extension/blob/master/app/setup/utils/index.js)

We also need to add the license header to all the new files.  

++ @IAmThePan for review

mai-cliqz:
Thank you very much for reviewing @christophertino. I have applied some changes based on your feedback. Please have a look again.

IAmThePan:
There is a lot here.  Two things that jump out at me are:
1. We recently made some changes to the blocking logic in `app/panel/utils/blocking.js` (d5d26e1b). Make sure those changes are implemented in the Android panel.
2. I think you need to add a line to webpack.config.js `${RM} ./dist/panel_android_ui.js` (around line 48)

mai-cliqz:
@IAmThePan Thank you very much for your time.
1. We discussed together with our UX and Vincent about the changes we made in blocking.js. So we decided not to change anything for now. Anyway the logic on desktop and mobile extension are a bit different.
2. We don't add the line to webpack.config.js because we don't want to remove the file after building. Since the file is needed for the UI to function.

PR #101, opened 2018-06-13
Inter-extension communication with Cliqz

sammacbeth:
This is a PoC of messaging between Ghostery and Cliqz extension, allowing Ghostery to call actions on Cliqz modules.

The first use case is importing settings from the antitracking and adblocker modules in Cliqz browser, allowing for the migration of this functionality from cliqz extension to Ghostery. The Cliqz side of this communication is implemented [here](https://github.com/cliqz/navigation-extension/pull/5942). 

Further work is needed to define the on-boarding flow for this migration, as currently antitracking and adblocker are forced off in Cliqz browser.

### DIFF
--- a/src/utils/cliqzSettingImport.js
+++ b/src/utils/cliqzSettingImport.js
@@ -1,0 +1,70 @@
+import KordInjector from '../classes/ExtMessenger';
+import { log } from './common';
+
+function promiseTimeout(timeout) {
+	return new Promise((resolve, reject) => {
+		setTimeout(reject, timeout);
+	});
+}
+
+function runCliqzSettingsImport(cliqz, conf) {
+	log('run cliqz settings importer');
+	const inject = new KordInjector();
+  inject.init();
+	// inject modules in remote cliqz extension with which we want to communicate
+	const privacyMigration = inject.module('privacy-migration');
+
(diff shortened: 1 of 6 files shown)


src/utils/cliqzSettingImport.js
```diff
@@ -0,0 +1,70 @@
...
+				}
+			});
+
+			// import site whitelists
+			const existingSites = new Set(conf.site_whitelist);
+			const newSites = new Set(modules.map(mod => result[mod].whitelistedSites)
+				.reduce((lst, val) => lst.concat(val), [])
+				.map(s => s.replace(/^(http[s]?:\/\/)?(www\.)?/, ''))
+				.filter(s => !existingSites.has(s)));
+			log('add whitelisted sites', [...newSites]);
```
zarembsky: const site_whitelist;

zarembsky: Should be:
```const site_whitelist;
newSites.forEach((s) => {
       site_whitelist.push(s);
}
conf.site_whitelist = site_whitelist;```

sammacbeth:
Thanks for the comments @zarembsky. I've updated the PR to address these issues.

PR #123, opened 2018-07-05
GH-1239 Tracker count respects SmartBlock block/unblock

IAmThePan:
Make the total Tracker Count respect the SmartBlock block/unblock numbers


IAmThePan:
Actually we do want this PR
@christophertino please approve

PR #165, opened 2018-08-17
fix conf / panel data setting enable_offers and signal sent

trickpattyFH20:


trickpattyFH20:
@jsignanini updated with fix for toggling from hub

PR #170, opened 2018-08-22
Restructuring timeline of loading to avoid hangs.

zarembsky:
* [ x] Have you followed the guidelines in [CONTRIBUTING.md](../CONTRIBUTING.md)?
* [ x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x ] Have you added an explanation of what your changes do?
* [ x] Does your submission pass tests?


zarembsky:
We call account-related methods after all cliqz modules scheduled to be enabled are fully enabled, meaning that their respective on("enabled"...) handlers were called. setPrefs call is moved to the very end, so that we call expensive storage operation just once on startup.

zarembsky:
There are two cases when it may hang. The prerequisites are same for both:  any one of the following modules - offers, humanweb, antitracking or adblocking is disabled in settings of a user account.
Case 1. User is logged in and installing Ghostery.
Case 2. User disables one of cliqz modules on one device; on another device the user is logged in but has not yet open a panel. At this point an update happens.

PR #174, opened 2018-08-24
GH-1012 fixes

IAmThePan:
GH-1012 fixes:
- remove justInstalled query param,
- fix header newline, 
- fix font-family,
- fix boldness


IAmThePan:
This PR also fixes the new updates to the Blocking step outlined in GH-1350

PR #205, opened 2018-10-05
GH-1404 move offer logo to header

trickpattyFH20:


app/panel/components/Rewards.jsx
```diff
@@ -133,6 +133,11 @@ class Rewards extends React.Component {
 	 * @return {JSX} JSX for the Rewards Header
 	 */
 	renderRewardsHeader() {
+		let reward;
+		const id = this.props.location.pathname.split('/detail/rewards/detail/')[1];
```
jsignanini: @trickpattyFH20 calling "index 1" (`[1]`) here could potentially cause a runtime error if `this.props.location.pathname` is not an Array or has length < 1 for whatever reason. Can we do this in a less error prone way?

PR #223, opened 2018-10-26
Uplift ghostery android

luciancor:
@christophertino @jsignanini can you please please take a look at this one? I tried to wrap up all the changes we did for the Ghostery mobile android browser.

### DIFF
--- a/src/classes/Globals.js
+++ b/src/classes/Globals.js
@@ -16,6 +16,15 @@
 import parser from 'ua-parser-js';
 
 const manifest = chrome.runtime.getManifest();
+const MOBILE_APP_UPDATE_URL = 'https://s3.amazonaws.com/cdncliqz/update/android_browser/firefox@ghostery.com/update.json';
+const CLIQZ_UPDATE_URL = 'https://s3.amazonaws.com/cdncliqz/update/browser/firefox@ghostery.com/update.json';
+
+function isUpdateUrl(applications, url) {
+	return applications
+		&& applications.gecko
+		&& applications.gecko.update_url
+		&& applications.gecko.update_url.indexOf(url) === 0;
+}
 /**
  * Structure which holds parameters to be used throughout the code, a.k.a. global values.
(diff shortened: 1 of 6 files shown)


src/classes/Globals.js
```diff
@@ -132,6 +142,10 @@ class Globals {
 			this.BROWSER_INFO.displayName = 'Cliqz';
 			this.BROWSER_INFO.name = 'cliqz';
 			this.BROWSER_INFO.token = 'cl';
+		} else if (this.IS_MOBILE_APP) {
+			this.BROWSER_INFO.displayName = 'Ghostery Mobile';
```
luciancor: hey @valmik-ghostery ! Would this work for you?

valmik-ghostery: Yeah works for me. Just a couple of things.

1 - Not sure about this but you might need to add the token here as well since that's what pulled into the telemetry database.

2 - We need to differentiate between iOS and android if possible. Will this allow us to do that?

luciancor: 1- what token do you mean?
2- we do not run this extension on iOS at all. We might need to add a different way to measure stuff there. Should we implement something from scratch?

luciancor: 1- I added something random there, we can change it to whatever makes more sense for you :) if ```ghosty``` and/or ```gh``` is not good we can definitely change 

2- great! yes, we should have anolysis there so all good ;)

valmikkpatel: Awesome. Thanks!

1 - can you use 'an' for android. That will fit in with what we used earlier. But anything is fine as long as it's not 'ff'.

luciancor: @valmikkpatel definitely! just updated the PR to have the token ```an```

Are the other values fine? ```ghosty``` and ```Ghostery Mobile``` ?

valmik-ghostery: Yes the other values are fine. Thanks!

PR #242, opened 2018-11-09
GIS-81: Display Correct Currency Symbol and Placement

Eden12345:
JIRA Ticket: https://cliqztix.atlassian.net/browse/GIS-81


Eden12345:
Copying José's comment from account-web to this PR as it is also applicable (he will work on how to deliver this JSON file to the extension project before I complete either of them)

@Eden12345 please update PR to match latest stripe/billing component refactor:

1. Remove country list from React components (it now lives in /app/Billing/countries.json)
2. Add currencySymbolAfter changes to said countries.json file
3. Check that GIS-81 fixes still works

Eden12345:
The only duplicate currency we had was the euro, so I moved all of the euro formats besides the most popular one (comma with the € symbol after the amount) to the end of the currencies.json file. That way, if we ever implement country-specific formats again, they will still work with the same file, but for now it will be consistent as we're accessing the currency with a for loop.

Eden12345:
Giving this PR a bump @jsignanini. The only change that needed to be made since your last review was changing the countries.json filename.

Issue #253, opened 2018-11-27
Ghostery v8.2.5 for Firefox prompts users to re-acknowledge permissions


christophertino:
The v8.2.5 release on Firefox is prompting users to re-accept Ghostery's permissions via an upgrade dialog. This is triggered by #234, which allows Ghostery to catch trackers and ads served over WebSockets (`"ws://*/*", "wss://*/*"`).  This is the only permission that was changed and it has not triggered a similar permission dialog in Chrome or Opera. 

There's an open Bugzilla bug to address why Firefox is forcing users to acknowledge permissions again: 
["Adding ftp/ws/wss permissions should not trigger extension permission warnings on extension upgrade"](https://bugzilla.mozilla.org/show_bug.cgi?id=1504018)

dwaal:
This also struck me as odd. Especially since there is no mention in the 8.2.5 changelog talking about requiring new permissions.

christophertino:
Yes, it's pretty disappointing. The changelog mentions adding WebSocket support but not the new permissions because the permission dialog was not triggered in our normal regression testing. Apparently to test these warnings you need to self-sign and host your test builds on an HTTPS server and then jump through some other hoops as outlined [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1411999#c18).

It's also concerning that some of our users aren't receiving the permission dialog in the url bar.  Instead, it's buried in the FF hamburger menu and if you don't see it, the permission request just goes away when you close Firefox. 
![ff_menu](https://user-images.githubusercontent.com/4699516/49228286-013bdd80-f3b9-11e8-8863-c939acdbdfca.png)

Since most of our users are now stuck on 8.2.4 and we're taking a bath on our [review board](https://user-images.githubusercontent.com/4699516/49228878-54626000-f3ba-11e8-8c89-7fa233300492.png), we're going to release an 8.2.6 build with WebSocket support removed on Firefox. Hopefully we can work with Firefox to acknowledge that this is a redundant permission prompt. Users have already agreed to the "allow access to data on all websites" permission on installation (via `"http://*/*", "https://*/*"` permissions).

smalluban:
Feel free to re-open if the issue is still valid. The only platform that is still on the v8 version is Firefox.

PR #255, opened 2018-11-29
GH-1481 Remove WS permissions for Firefox

IAmThePan:
GH-1481: Remove WebSockets so that upgrading Firefox users aren't asked to re-validate our permissions.  Will require MOAB to be updated.

Also updated Browser-Core

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -62,6 +62,7 @@ const {
 	CDN_SUB_DOMAIN, BROWSER_INFO, IS_CLIQZ, DEBUG
 } = globals;
 const IS_EDGE = (BROWSER_INFO.name === 'edge');
+const IS_FIREFOX = (BROWSER_INFO.name === 'firefox');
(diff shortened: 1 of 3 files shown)


src/background.js
```diff
@@ -1416,7 +1417,8 @@ function initializePopup() {
  */
 function addCommonGhosteryAndAntitrackingListeners() {
 	let urlFilters = ['http://*/*', 'https://*/*', 'ws://*/*', 'wss://*/*'];
-	if (IS_EDGE) {
+	if (IS_EDGE || IS_FIREFOX) {
```
christophertino: @IAmThePan looks like we are going to remove `ws` and `wss` altogether for now.  See cliqz pipeline [here](https://github.com/cliqz/navigation-extension/commit/a5e65dc5d478e542c3aa886216b8ffd0dd690440)

christophertino:
Fixes #253

PR #266, opened 2018-12-07
fix tests, upgrade babel, upgrade browser-core

trickpattyFH20:
fix tests, upgrade babel, upgrade browser-core


package.json
```diff
@@ -40,9 +40,12 @@
     "email": "ghostery-dev@ghostery.com"
   },
   "homepage": "https://github.com/ghostery/ghostery-extension#readme",
+  "resolutions": {
+    "babel-core": "^7.0.0-bridge.0"
+  },
   "dependencies": {
     "base64-js": "^1.2.1",
-    "browser-core": "https://s3.amazonaws.com/cdncliqz/update/edge/ghostery/v7.32/7.32.0.dead86b.tgz",
+    "browser-core": "https://s3.amazonaws.com/cdncliqz/update/edge/ghostery/v7.32/7.32.0.dd59bfa.tgz",
```
Eden12345: 👍 This build should have the `insights` module without the `wss`

app/shared-components/SteppedNavigation/__tests__/__snapshots__/SteppedNavigation.test.jsx.snap
```diff
@@ -350,20 +320,16 @@ exports[`app/shared-components/SteppedNavigation Snapshot tests with react-test-
...
           href="/1"
           onClick={[Function]}
-          style={undefined}
         />
         <a
           className="active"
         />
         <a
           aria-current={null}
-          className={undefined}
```
Eden12345: I don't know when all of these undefined props were added to the snapshots, but it's definitely good that they were removed with the newly generated ones.

Eden12345: @trickpattyFH20 — @IAmThePan said that it may be best to revert back to the last snapshots with those undefined props in them. We did have snapshots with undefined props in the past that passed the tests, but I'm not sure if there's a setting that needs to be changed in a Jest config file or something like that.

PR #268, opened 2018-12-10
Parameterize production/non-production builds

sharath-cliqz:
@chrmod Please check :)


chrmod:
Thanks @sharath-cliqz - I will start the debug build now

PR #269, opened 2018-12-10
GH-1336: Send Historical Tracker Stats to Ghostery Tab Extension via Messaging

Eden12345:
JIRA Ticket: https://cliqztix.atlassian.net/browse/GH-1336

**Notes**
This sets up a listener that waits for a message from another extension, and will execute a function that passes the locally stored historical stats (which are aggregated in that function) and some user settings if the extension ID associated with the message matches the Ghostery Tab extension and the name of the request is `'getStatsAndSettings'`. The Ghostery Tab needs this because it displays that information whenever a new tab is opened.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -1399,6 +1399,10 @@ messageCenter.on('enabled', () => {
 	});
 });
 
+/**
+ * Add page listeners for insights stats.
+ * @memberOf Background
+ */
 insights.on('enabled', () => {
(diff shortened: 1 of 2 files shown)


Eden12345:
The Travis CI checks won't pass until the changes in this PR are merged: https://github.com/ghostery/ghostery-extension/pull/266

PR #271, opened 2018-12-10
Fix: Historical Stats Routing

Eden12345:
JIRA Ticket: https://cliqztix.atlassian.net/browse/GH-1427

**Notes**
Previously, the historical stats view would route back to the main view whenever you selected the back arrow in the header. In this branch, if you go to the historical stats view from the subscription menu, then it will now route back to the subscription menu.

### DIFF
--- a/app/panel/components/Header.jsx
+++ b/app/panel/components/Header.jsx
@@ -137,8 +137,9 @@ class Header extends React.Component {
 
 	determineBackPath = () => {
 		const { entries, location } = this.props.history;
+		const subscriptionRegEx = /^(\/subscription)/;
 		if (location.pathname === '/stats' && (entries.length > 1 &&
-		entries[entries.length - 2].pathname.slice(0, 13) === '/subscription')) {
+		subscriptionRegEx.test(entries[entries.length - 2].pathname))) {
 			return 'subscription/info';
 		}
 		return this.props.is_expert ? '/detail/blocking' : '/';
--- a/app/panel/components/Header.jsx
+++ b/app/panel/components/Header.jsx
@@ -135,6 +135,12 @@ class Header extends React.Component {
 	}
 
 	clickLogo = () => {
+		const { entries, location } = this.props.history;
+		if (location.pathname === '/stats' &&
+		entries[entries.length - 2].pathname === '/subscription/info') {
+			this.props.history.push('/subscription/info');
+			return;
+		}
 		this.props.history.push(this.props.is_expert ? '/detail/blocking' : '/');
 	}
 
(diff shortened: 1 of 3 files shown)


app/panel/components/Header.jsx
```diff
@@ -135,6 +135,12 @@ class Header extends React.Component {
 	}
 
 	clickLogo = () => {
+		const { entries, location } = this.props.history;
+		if (location.pathname === '/stats' &&
+		entries[entries.length - 2].pathname === '/subscription/info') {
```
jsignanini: @Eden12345 can we do this in a "safer" way just in case `entries` has a length lower than 2? Could that ever be an issue?

Eden12345: As the app stands it shouldn't be an issue because you have to leave the menu to get to a view where the header has a back option, but I still see your point and will add a check there.

app/panel/components/Header.jsx
```diff
@@ -135,6 +135,12 @@ class Header extends React.Component {
 	}
 
 	clickLogo = () => {
```
jsignanini: @Eden12345 can you please rewrite this function using `<Redirect ... />` from `react-router-dom`? It seems like we should have been doing this way all long, instead of messing with the history.

Eden12345: I used `<Link ... />` instead of `<Redirect ... />` for two reasons:

1) This routing action is associated with a click event rather than UI state change.

2a) `<Redirect ... />` should only be used when you are redirecting out of a component. This routing action is being triggered from the `<Header />` component, which will not be removed from the DOM when the routing action occurs.

2b) If we wanted to use `<Redirect ... />` here, we would need to handle it in the local state of the `<Panel />` component and pass that state as a prop to the component that is being redirected away from. We would also have to pass a class method for the `<Panel />` component that changes this local state as a prop to the `<Header />` component. This feels more cumbersome than using `<Link ... />`.

app/panel/components/Header.jsx
```diff
@@ -134,8 +135,13 @@ class Header extends React.Component {
 		);
 	}
 
-	clickLogo = () => {
-		this.props.history.push(this.props.is_expert ? '/detail/blocking' : '/');
+	determineBackPath = () => {
+		const { entries, location } = this.props.history;
+		if (location.pathname === '/stats' && (entries.length > 1 &&
+		entries[entries.length - 2].pathname.slice(0, 13) === '/subscription')) {
```
jsignanini: @Eden12345 let's make this a regular expression match to make our intentions a bit more clear.

Eden12345: @jsignanini done!

PR #272, opened 2018-12-10
GH-1464: make sure we show offer popup in the tab with the offer triggering url, whether or not that's the active tab

wlycdgrfromflatiron:
This works with the same chunk of code as https://github.com/ghostery/ghostery-extension/pull/270 and is blocked by it.

Select tab in which to display the Ghostery Rewards offer by the Cliqz offer url rather than by active state, to account for the possibility of the user opening said url in a new tab through the context menu (which may not make the tab with that url active).


wlycdgrfromflatiron:
Erroneously requested merge to master instead of develop.

PR #273, opened 2018-12-10
GH-1464: Make sure we show offer popup in the tab with the offer triggering url, whether or not that's the active tab

wlycdgrfromflatiron:
This works with the same chunk of code as #270 and is blocked by it.

Select tab in which to display the Ghostery Rewards offer by the Cliqz offer url rather than by active state, to account for the possibility of the user opening said url in a new tab through the context menu (which may not make the tab with that url active).

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -237,7 +237,8 @@ export function getTabByUrl(callback, url) {
 	chrome.tabs.query({
 		url
 	}, (tabs) => {
-		callback(tabs[0]);
+		const tab = tabs ? tabs[0] : false;
+		callback(tab);
 	});
 }
 
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -233,9 +233,9 @@ export function getActiveTab(callback) {
  *
  * @param {string} url				the tab url to search for
  * @param {function} callback		function to call if at least one matching tab is found
- * @param {function} fallback		function to call if the provided url pattern is invalid or no matching tab is found
+ * @param {function} error			function to call if the provided url pattern is invalid or no matching tab is found
  */
-export function getTabByUrl(url, callback, fallback) {
+export function getTabByUrl(url, callback, error) {
 	chrome.tabs.query(
 		{
 			url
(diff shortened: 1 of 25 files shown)


src/utils/utils.js
```diff
@@ -226,6 +226,21 @@ export function getActiveTab(callback) {
...
+ * @memberOf BackgroundUtils
+ *
+ * @param {function} callback		function to call if tab found
+ * @param {string} url				the tab url to search for
+ */
+export function getTabByUrl(callback, url) {
+	chrome.tabs.query({
+		url
+	}, (tabs) => {
+		callback(tabs[0]);
```
jsignanini: @wlycdgrfromflatiron can you double check what the behavior here is if there are no tabs found with the query search criteria? If so, what would the value of tabs be? We might need to add a check so that we don't run into an "array out of bounds error" or "can't access variable of undefined" or something like that.

wlycdgrfromflatiron: If no tabs were found, tabs would be undefined, causing an error on the bracket access. I modified the code to guard against this in the commit below.

src/utils/utils.js
```diff
@@ -226,6 +226,22 @@ export function getActiveTab(callback) {
...
 }
 
+/**
+ * Query the first tab that matches the url argument
+ * @memberOf BackgroundUtils
+ *
+ * @param {function} callback		function to call if tab found
+ * @param {string} url				the tab url to search for
+ */
+export function getTabByUrl(callback, url) {
```
jsignanini: @wlycdgrfromflatiron please rewrite this function so that we check for `chrome.runtime.lastError` in the callback function so that we can be assured that `tabs` will be an array, and that we handle an array of size `0` to whatever length.

wlycdgrfromflatiron: I've rewritten getTabByUrl to invoke a fallback function argument in case of chrome.runtime.lastErrror or if the query returns an empty array, using the error handling code in getTab as a template.

src/utils/utils.js
```diff
@@ -226,6 +226,37 @@ export function getActiveTab(callback) {
...
+/**
+ * Query the first tab that matches the url argument.
+ * Will throw an error if the url argument is an invalid url pattern.
+ * @memberOf BackgroundUtils
+ *
+ * @param {string} url				the tab url to search for
+ * @param {function} callback		function to call if at least one matching tab is found
+ * @param {function} fallback		function to call if the provided url pattern is invalid or no matching tab is found
+ */
+export function getTabByUrl(url, callback, fallback) {
```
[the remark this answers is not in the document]
wlycdgrfromflatiron: Done

PR #274, opened 2018-12-11
Performance optimisations on critical path

sammacbeth:
Performance improvements to webrequest handling for android build.

 1. Use native url parsing from `URL` class.
 2. Avoid 4+ redundant parses of the tab url in `isWhitelist` checks.
 3. Drop regex matching.

@chrmod

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -1093,8 +1093,8 @@ function initialiseWebRequestPipeline() {
  * @return {boolean}
  */
 function isWhitelisted(state) {
-	const { host } = state.sourceUrl.host;
-	return globals.SESSION.paused_blocking || state.ghosteryWhitelisted || events.policy.whitelisted(host) === 2;
+	const { host } = state.sourceUrlParts.host;
+	return globals.SESSION.paused_blocking || state.ghosteryWhitelisted || events.policy.whitelisted(host);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/background.js
```diff
@@ -1159,7 +1159,12 @@ adblocker.on('enabled', () => {
 				before: ['checkBlocklist']
 			}),
 			adblocker.action('addWhiteListCheck',
-				url => isWhitelisted({ sourceUrl: url }))
+				url => isWhitelisted({
+					sourceUrlParts: {
```
jsignanini: @sammacbeth can we write this as `{ sourceUrlParts: new URL(url) }` instead? A bit cleaner, so we don't do `sourceUrlParts.host.host` later on?

remusao: @jsignanini You are right, it's not really nice, but unfortunately this is needed (at the moment) to mimic some internal data structure that is expected by this step (there is indeed a `host.host` at some place in the code). We will probably try to clean this up to make the structure of this object more consistent.

jsignanini:
@sammacbeth I see you removed regexp matching from `matcher.isBug`... could you please explain?

remusao:
@jsignanini I'm answering since Sam is OOO today. We thought about disabling the regexps for android build because on mobile they are the bottle-neck of the matching (on longish URLs it can take a lot of time) and from some measurements, they do not seem to match very often. So this is more of a hacky-fix/low hanging fruit for now to get some speed boost for the android build until a better long-term solution is found.

PR #307, opened 2019-01-07
Refactor supporter to plus.

jsignanini:


christophertino:
Looks ok to me but would like @IAmThePan to review

zarembsky:
For unverified accounts makes two unnecessary requests to stripe and to themes, _getUserIDIfEmailIsValidated() returns Uncaught in Promise Error.

PR #326, opened 2019-01-23
Stop overriding cliqz ad block module whitelist check

ravjit-cliqz:


sammacbeth:
How do you whitelist ghostery? That should act as a master switch to whitelist across ghostery, anti-tracking and adblocker.

ravjit-cliqz:
The ghostery whitelist is done by adding the domain to `site_whitelist` dictionary. And I think it doesnt affect the adblocker and antitracking(Looks like an bug). I can still see the count and the ads are getting blocked. 
But then again, we can't just disable the adblock per domain, with other security features staying on.

remusao:
We realized that this PR was not merged and is needed to make the whitelisting work: https://github.com/ghostery/ghostery-extension/pull/312 so with this change the whitelisting works fine again.

sammacbeth:
Was this meant to be merged? I though the other PR fixes it?

PR #350, opened 2019-03-06
Throttle browser button counter updates

chrmod:
Counter updates can happen very fast and every update issues a call to chrome process. This change should improve performance - specifically for underpowered devices like mobiles.


christophertino:
@IAmThePan do you remember what issue we had with passing tabID to `button.update`?

chrmod:
btw. we've tested this change on mobile browser and seems to work without any issues. Would be probably better to have an update queue as there are multiple sources of "badge" updates.

IAmThePan:
I don't quite remember but I think there were inconsistencies where the passed TabId wasn't for the currently active tab.
Examples:

- What if you have two windows open?  Will it incorrectly set the badge count to the other window's active tab count?
- What if a tab closes and a Errored request is a tracker, will that incorrectly update the currently active tab.
- What if the TabId changes (a prefetched tab becomes active)?

chrmod:
regardless of correct behavior of `button.update` this PR should introduce no functional change other than throttling. Note the `_throttleButtonUpdate` called without an argument will pass `undefined` value as `tabId`, which should be no different from passing no argument.

PR #354, opened 2019-03-07
Compress all .png images in the Hub

IAmThePan:
I used https://tinypng.com/ to compress all of the .png files in `app/images/hub`.  This reduced the size of the folder from 3.4 MB to 948 KB, and reduced the size of the package from 8.4 MB to 6.1 MB.
You can use the Display Rich Diff to compare the images.  I couldn't tell the difference between any of the original and compressed images.


IAmThePan:
I used Sketch to shrink the image sizes to match the sizes used throughout the extension.  Then I used https://tinypng.com/ to compress the shrunken .png files.  This further reduced the size of the `app/images/hub` to 625 KB and reduces the size of the package to 5.7 MB.
With this commit I am able to notice the reduced image quality, but maybe it is worth the saved bytes.

Issue #376, opened 2019-05-03
Error on blocked redirect page prevents user from following blocked redirects


tomclose:
### Description

Error on blocked redirect page

### Expected Behavior 

When I click on an ad result in a google search I am taken to a ghostery page, which displays a link which I can click on if I want to display the page linked to in the ad result.

### Actual Behavior 

When I click on an ad result in a google search I am taken to a blank page. In the dev console the on the page I see the following errors:

![Screen Shot 2019-05-03 at 11 56 19 AM](https://user-images.githubusercontent.com/29596/57133420-a27cf480-6d9a-11e9-943b-73c5271961cd.png)

### Steps to Reproduce

1. Search for "buy a chair" or similar on google
2. Select one of the results labelled `AD`

### Versions

+ Browser: Google Chrome Version 74.0.3729.131 (Official Build) (64-bit)
+ OS: Mac OS Mojave (10.14.4)
+ Node: N/A
+ NPM: N/A

shana:
I'm having the same issue. `sendMessageInPromise` is undefined in [`blocked_redirect.js` ](https://github.com/ghostery/ghostery-extension/blob/master/app/content-scripts/blocked_redirect.js#L146)

https://github.com/ghostery/ghostery-extension/commit/e8e2bdf4a6af6b29a89b92cd69bd2102be9a32e0 is this commit that broke this.

christophertino:
This should be fixed in v8.3.4 which pushed to all stores today.  Please update and let us know if you still see the issue.

shana:
Yup, I've confirmed it's working in 8.3.4 :+1:

PR #380, opened 2019-05-07
Fix blocked_redirect by fixing promises and images

IAmThePan:
Duplicate of PR #379 but off the Hotfix branch:

Fix Blocked Redirect:
- Fix images from not showing. I did this with as little change to the code as possible. The solution was to just add PNG files to the extension.
- Fix promises and sendMessageInPromise to return correctly. There was an extra set of () that prevented sendMessageInPromise from returning correctly. There was also no return in content-scripts/utils/msg's sendMessageInPromise which prevented the function from correctly returning a promise.

@christophertino Please approve.


christophertino:
This fixes #376

PR #385, opened 2019-05-23
EX-8836: Ensure that filtered Cliqz rewards are retrieved on extension panel open

wlycdgr:
Ensure that filterOffersByRemote gets called on extension panel open even though the panel now communicates with the background exclusively through a port and no longer sends the 'getPanelData' message.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -782,6 +782,8 @@ function onMessageHandler(request, sender, callback) {
 	}
 
 	// HANDLE UNIVERSAL EVENTS HERE (NO ORIGIN LISTED ABOVE)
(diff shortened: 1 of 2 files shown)


wlycdgr:
> looks good, thx)
> 
> One issue (but we can live with this issue):
> 
>     1. user clicks on ghostery's icon -> simple view
> 
>     2. user clicks on detailed view
> 
>     3. user clicks on rewards
>        `rewards.filterOffersByRemote` was not called, but should be called

Addressed - rewards.filterOffersByRemote is now called in this case also.

Also, removed unneeded call to rewards.filterOffersByRemote in response to requests for data from background by the Hub and the Android extension, as neither displays rewards details.

PR #388, opened 2019-05-30
GH-1566 - Smart blocking should check for Click2Play

christophertino:
When a tracker is smart blocked, we should load the C2P template when necessary. In these cases, we only show the "allow always" option since most of these widgets load on user interaction (lazy load) or beyond domComplete, negating the "allow once" option. 

@wlycdgr @Eden12345 please make sure those lodash replacements look ok.


Eden12345:
The underscore implementation looks good to me

app/panel/components/BuildingBlocks/DonutGraph.jsx
```diff
@@ -11,7 +11,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
-import throttle from 'lodash.throttle';
+import _ from 'underscore';
```
jsignanini: @christophertino can we do `import { throttle } from 'underscore';` instead?

app/panel/components/BuildingBlocks/StatsGraph.jsx
```diff
@@ -11,9 +11,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
+import _ from 'underscore';
```
jsignanini: @christophertino same here for `isEqual`.

PR #394, opened 2019-06-11
GH-1616: tracker list breakdown detailed view

wlycdgr:
Show additional Cliqz details for each tracker in the detailed view: number of cookies blocked / fingerprints wiped by Cliqz for this tracker, if any.

~~Does not implement the ad portion of the ticket, as discussed.~~  Now implements the ad portion of the ticket also.

Note that I haven't been able to find a fingerprint hit in the wild to confirm that it is passed through correctly from the background. However, I've confirmed that the fingerprint readout as well as the cookie readout displays correctly when passed dummy data on the front end.


sammacbeth:
I managed to get a fingerprint shown,  looks good!

Screen Shot 2019-06-11 at 13 00 12
Screen Shot 2019-06-11 at 14 04 55

src/classes/FoundBugs.js
```diff
@@ -276,6 +276,16 @@ class FoundBugs {
 		return apps_arr;
 	}
 
+	getAppsById(tab_id) {
```
IAmThePan: Please add a documentation comment for this function.

wlycdgr: Added

wlycdgr:
Ads portion of GH-1616 implemented

wlycdgr:
Fix bug where panel would not display if a bug id returned by getGhosteryStats did not map to an app id in our PanelData trackerList

PR #401, opened 2019-06-21
Improve green upgrade banner layout robustness 

wlycdgr:
...in face of locale string variations

Make banner a pixel or two taller so that the contain behavior of the green upgrade banner background has more room to work with. Add classes that reduce font size for the much wider Japanese and Korean strings. Shorten French translation.


app/panel/components/Summary.jsx
```diff
@@ -684,10 +684,14 @@ class Summary extends React.Component {
...
 		const { is_expert } = this.props;
 
+		const ctaText = t('subscription_upgrade_to');
+
 		const isPlusSubscriber = this._isPlusSubscriber();
 		const upgradeBannerClassNames = ClassNames('UpgradeBanner', {
 			'UpgradeBanner--normal': !is_expert,
 			'UpgradeBanner--small': is_expert,
+			'UpgradeBanner--japanese': ctaText === 'アップグレード先',
+			'UpgradeBanner--korean': ctaText === '다음으로 업그레이드',
```
christophertino: @wlycdgr Rather than string checking, see if you can sniff out the active language with `chrome.i18n.getUILanguage`

https://developer.chrome.com/apps/i18n#method-getUILanguage

wlycdgr: I was thinking about it in my sleep and I thought of a cleaner simpler way to do it. The large height and turning off vertical tiling prevents background tilling while allowing banner background to stretch for wider strings. Checked with Japanese, Russian, English, French, looks ok in all those.

PR #407, opened 2019-07-02
Feature/gh 1596/rewards icon simple view

wlycdgr:
Implement a Rewards icon in the simple Summary view as per GH-1596


_locales/en/messages.json
```diff
@@ -750,9 +750,6 @@
 	"panel_detail_performance_text": {
 		"message": "Feature coming soon."
 	},
-	"panel_detail_rewards_title": {
```
IAmThePan: Did you run the test script?  I'm still seeing this string in some snapshot files

app/panel/components/Summary.jsx
```diff
@@ -219,15 +220,24 @@ class Summary extends React.Component {
 		this.props.history.push(this._isPlusSubscriber() ? '/subscription/info' : `/subscribe/${!!this.props.user}`);
 	}
 
+	/**
+	 * Show the Rewards view
+	 * Used to handle user clicking on the Rewards icon
+	 */
+	showRewardsView() {
```
IAmThePan: Do you use this or do you rely on the `NavButton` link?  I'm noticing a bug when I click the Rewards button but not the SVG.  It takes me to a blank Detail View.

app/panel/components/Summary.jsx
```diff
@@ -219,15 +220,24 @@ class Summary extends React.Component {
...
+		// this.toggleExpert('gotorewards');
+		this.props.actions.toggleExpert();
+	}
+
 	/**
 	 * Toggle between Simple and Detailed Views.
 	 */
-	toggleExpert() {
+	toggleExpert(subview = 'blocking') {
```
IAmThePan: With this default, maybe you don't need to change line 125.

app/panel/components/Summary.jsx
```diff
@@ -686,6 +696,29 @@ class Summary extends React.Component {
...
+			'g-tooltip',
+			{
+				hide: this.props.is_expert,
+			}
+		);
+
+		return (
+			<div className={rewardsIconClassNames} onClick={this.showRewardsView}>
+				<NavButton path="/detail/rewards/list" imagePath="../../app/images/panel/rewards-icon.svg" />
+				{unreadOffersAvailable && <NavButton classNames="Summary__rewardsIcon__star" path="/detail/rewards/list" imagePath="../../app/images/panel/purple-star.svg" />}
```
IAmThePan: I haven't been able to test this because I don't know how to generate a reward...

app/scss/partials/_summary.scss
```diff
@@ -168,14 +168,36 @@
...
+	padding-left: 9px;
+}
+.Summary__rewardsIcon.Summary__rewardsIcon--absolutely-positioned {
+	top: 85px;
+	padding-right: 6px;
+	padding-left: 10px;
+	text-transform: capitalize;
+}
+// position the star SVG relative to the gift box SVG
+.Summary__rewardsIcon__star {
```
IAmThePan: again, I'll check this when changes are made and I generate a reward.

_locales/en/messages.json
IAmThePan: Hmm I'm seeing that the unit tests do pass but I see `Rewards.jsx.snap`.

IAmThePan: The test is skipped so the unit tests pass but the snapshot is still out of date.  Did you skip the test?

app/panel/components/Summary.jsx
```diff
@@ -686,6 +696,29 @@ class Summary extends React.Component {
 		);
 	}
 
+	_renderRewardsIcon() {
```
christophertino: Can we add a quick comment header on this method?

_locales/en/messages.json
wlycdgr: git blame says I did, back in March. It must have been a quick workaround that we needed for the dynamic UI testing that I did not revert.

wlycdgr: Made changes necessary to reenable the Rewards component tests, reenabled them, and updated the snapshots.

app/panel/components/Summary.jsx
wlycdgr: Fixed showRewardsView (renamed to showRewardsList) to correctly hanle click and convert the NavButtons to ReactSVGs to simplify & give showRewardsList sole responsibility.

app/panel/components/Summary.jsx
wlycdgr: Unnecessary parameter removed from toggleExpert call in clickDonut

app/scss/partials/_summary.scss
wlycdgr: You can force the value of unreadOffersAvailable in Summary#_renderRewardsIcon to true to test the appearance/layout of the star. That bool is true wwhen the unread_offer_ids array sent from the background is not empty.

app/panel/components/Summary.jsx
wlycdgr: As per Vinnie's decision about this, the button should display regardless of whether any rewards (new or seen) are available, so long as the user has not turned Rewards off, so it should be enough to check that the icon takes the user to the Rewards list view, and that it does not display when Rewards are off. For testing the purple star appearance / logic, see next comment.

app/panel/components/Summary.jsx
wlycdgr: Done

app/panel/components/Summary.jsx
IAmThePan: Please revert to use the `NavButton`.  That way worked if you had only removed the `onClick` handler.  This new implementation introduces two bugs:
1. There is no `cursor: pointer` on the SVG.  There should be because it makes sense as you can click the SVG and also because the Historical Stats link above has it.
2. You can click the box around the SVG while you can't for Historical Stats.

Please make these as similar as possible by reverting to the old way of using `NavButton` (like the Historical Stats button does) and by removing the `onClick` handler (which again, the Historical Stats button doesn't have).

wlycdgr: I've made the UIs and implementations of the Historical Stats and Rewards navicons consistent, but by converting the Stats navicon to use ReactSVG also instead of reverting to NavButton for the Rewards navicon. I chose this solution because NavButton isn't sufficient to handle clicks on the Rewards navicon, which need to call toggleExpert in addition to pushing a new route onto the history stack. I think letting user click anywhere inside the bordered navicon area is, anyhow, more user-friendly than requiringt them to click on the icon itself. It's also more consistent with the click-on-donut behavior (clicking on empty space inside the donut works as well as clicking the donut itself). The cursor behavior is now consistent between the two navicons also.

PR #410, opened 2019-07-04
Use cliqz-url-parser for URL processing.

sammacbeth:
Updates `utils.processUrl` to use the [cliqz url-parser](https://github.com/cliqz/url-parser) implementation instead of the current node `url.parse`. Our benchmarks show this should be able 10x faster than the node implementation.


src/utils/utils.js
```diff
@@ -168,7 +168,16 @@ export function processFpeUrl(src) {
...
  */
 export function processUrl(src) {
-	if (!src) {
+	try {
+		const res = new URL(src);
+		return {
+			protocol: res.protocol ? res.protocol.substr(0, res.protocol.length - 1) : '',
+			host: res.hostname || '',
+			path: res.pathname ? res.pathname.substr(1) : '',
+			host_with_path: (res.host || '') + (res.pathname || ''),
```
sammacbeth: This was carried over as-is from the previous version, but I wonder if it intentional that you use `host` here instead of `hostname`. The former includes the port number if it is non-standard. Are there patterns that use this information? If not, we could use `hostname` here to be more consistent. @christophertino

jsignanini: I think that since we were creating a somewhat custom object being returned from `processUrl()`, and not aiming to exactly replicate the `URL()` function, we just named it `host` for convenience but I agree that it could lead to confusion.

PR #414, opened 2019-07-17
Feature/gh 1733/legal updates

wlycdgr:
Adds a yes-required, opt-in legal consent checkbox to the Hub and Extension account creation forms.

Hub form: 'Create Account' in the sidebar
Extension form: Hamburger menu -> 'Sign In' -> 'Create Account'


app/hub/Views/CreateAccountView/CreateAccountViewContainer.jsx
```diff
@@ -93,9 +95,17 @@ class CreateAccountViewContainer extends Component {
 	}
 
 	/**
-	 * Update input checkbox values by updating state.
+	 * Update legal consetnt checkbox  value by updating state
 	 */
-	_handleCheckboxChange = () => {
+	_handleLegalConsentCheckboxChange = () => {
+		const legalConsentChecked = !this.state.legalConsentChecked;
+		this.setState({ legalConsentChecked });
```
christophertino: Use callback on `setState()` rather than `this.state`.  See [react/no-access-state-in-setstate](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md)

app/panel/components/CreateAccount.jsx
```diff
@@ -200,6 +210,15 @@ class CreateAccount extends React.Component {
 									</div>
 								</div>
 							</div>
+							<div className="row">
+								<div className="small-12 columns">
+									<div id="create-account-legal-consent-checkbox" className={(legalConsentNotCheckedError ? 'checkbox-error' : '')}>
+										<input id="legalConsentChecked" name="legalConsentChecked" type="checkbox" checked={legalConsentChecked} onChange={this.handleCheckboxChange} />
+										{/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
christophertino: Do we still need this lint exclusion?

wlycdgr: Without it, the jsx-a11y linter plugin complains that "A form label must be associated with a control". It thinks the label is empty because we are using dangerouslySetInnerHTML attribute to set the text instead of inserting the string directly into the tag as with the promotions checkbox. But if we don't use dangerouslySetInnerHTML, the links in the string are displayed as plain text HTML. It seemed to me that this was a rare case where disabling the linter is the lesser of two evils, the other evil being somehow changing the checkbox implementation and probably making it more complicated and inconsistent with the promotions checkbox implementation. What do you think?

app/hub/Views/CreateAccountView/CreateAccountViewContainer.jsx
wlycdgr: Done

PR #415, opened 2019-07-18
GH-1795: Expand Icon showing over Tooltip

tdtnguyen:
* [ ] Have you followed the guidelines in [CONTRIBUTING.md](../CONTRIBUTING.md)?

### DIFF
--- a/app/scss/partials/_tooltip.scss
+++ b/app/scss/partials/_tooltip.scss
@@ -23,7 +23,7 @@
 		background-color: $mystic;
 		text-align: center;
 		position: absolute;
-		z-index: 100;
+		z-index: 20;
 		box-shadow: 0px 2px 2px 0 rgba(0, 0, 0, 0.2);
 		box-sizing: border-box;
 		&:after {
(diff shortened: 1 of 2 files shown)


app/scss/partials/_tooltip.scss
```diff
@@ -23,7 +23,7 @@
 		background-color: $mystic;
 		text-align: center;
 		position: absolute;
-		z-index: 3;
+		z-index: 100;
```
Eden12345: You can check with @vincentlai88 if you'd like, but I'm pretty sure he would want the toast messages to show up in front of tooltips (should that scenario arise).

The current `z-index` for the `callout` class (which controls the styling associated with the toast messages in the `<Panel />` and `<Settings />` components) is 5, which would make it appear behind the tooltip `z-index` depending on the DOM organization. Please make ensure this doesn't happen by increasing the `callout` `z-index` property to be higher than the tooltip's `z-index`.

Furthermore, all of the z-indices in this project are relatively low, with the exception of a couple outliers. I think it would be a good idea to continue with that pattern, so I would suggest a `z-index` of `20` for the tooltip, and `z-index` of `30` for the `callout` class.

PR #417, opened 2019-07-18
GH-1621: Site-Specific Whitelisting for Unknown Trackers via Anti-Tracking

Eden12345:
JIRA Ticket: https://cliqztix.atlassian.net/browse/GH-1621

**Notes**
If you need more test cases, turn off Ad-Blocking and head to any news site like cnn.com

The race condition issue should be fixed along with a few other issues that @IAmThePan found, and I also have moved the strings into the translation library. It should be good to go at this point


app/panel/components/Blocking/Categories.jsx
```diff
@@ -29,28 +29,78 @@ class Categories extends React.Component {
...
+					trackers: antiTracking.unknownTrackers.map((unknownTracker, idx) => {
+						if (unknownTracker.whitelisted) { whitelistedTotal++; }
+						return {
+							name: unknownTracker.name,
+							domains: unknownTracker.domains,
+							whitelisted: unknownTracker.whitelisted,
+							blocked: false,
+							catId: 'anti_tracking_unknown',
+							description: '',
+							id: 100000000 + idx,
```
Eden12345: Remove this line along with any other props that are unnecessary for the `<Tracker />` component for Unknown trackers

Eden12345: **Edit:** We still need a unique ID for this because it's used later as a React `key` prop when rendering all the trackers from the `<Trackers />` component. I'll use the `unknownTracker.name` and concatenate it with the first domain listed in the `unknownTracker.domains` array (if the array is empty, it will just concatenate `"undefined"` onto the string).

PR #418, opened 2019-07-23
Gh 1763/metrics feature/new triggers for broken page pings

wlycdgr:
Watch for a range of user action sequences and report them to the metrics server as possibly indicative of a broken page, with the goal of improving our understanding of which user action sequences are actually most indicative of a broken page.


src/background.js
```diff
@@ -1101,6 +1101,8 @@ function initializeDispatcher() {
 		button.update();
 		utils.flushChromeMemoryCache();
 		cliqz.modules.core.action('refreshAppState');
+		// if user has whitelisted site, suspect broken page
+		metrics.handleBrokenPageTrigger(globals.BROKEN_PAGE_WHITELIST);
```
christophertino: @wlycdgr seems like this would be triggered when adding to and removing from the whitelist. Do we need both cases? If not we should check against `sitePolicy` to look for just the addition portion.

wlycdgr: Updated. Since `sitePolicy` is not currently being sent to the background and would have had to be added anyway, I instead send a `brokenPageMetricsWhitelistSite` bool to express intent more clealry and test for it in `PanelData#set`

PR #426, opened 2019-07-25
Gh 1767/bugfix/target blank links crash opera

wlycdgr:
Fixes Opera crashes in Help, About, Subscribe, and CreateAccount views that resulted from Opera needing to open panel-to-tab links through chrome.tabs.create only (vs. a raw link).

Note that the implementation changes needed to fix the bug in CreateAccount mean that it was necessary to add new strings to messages.json.


wlycdgr:
I've integrated José's changes from https://github.com/ghostery/ghostery-extension/pull/429 that improve the handling of i18n strings with embedded links, cleaned up duplication, and added doc headers. Ready for another look.

app/panel/components/BuildingBlocks/PanelToTabLink.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Implements panel -> new tab links. Used in Help, About, and other panel views
+ * @memberof PanelBuildingBlocks
+ */
+const PanelToTabLink = (props) => {
+	const { href, label } = props;
+
+	return (
+		<a href={href} onClick={openFixedDestinationLinkInNewTab}>{label}</a>
```
jsignanini: @wlycdgr we should be using `this.props.children` here so that we can pass it whatever we want as it's children (there's an example on `app/panel/components/Subscribe.jsx` L37` where this would help.

jsignanini:
@wlycdgr would also be nice if we could rename `openFixedDestinationLinkInNewTab` into something that reflects the fact that it expects a click event as it's parameter.

app/panel/components/BuildingBlocks/PanelToTabLink.jsx
wlycdgr: Refactored to do this.

wlycdgr:
> @wlycdgr would also be nice if we could rename `openFixedDestinationLinkInNewTab` into something that reflects the fact that it expects a click event as it's parameter.

Renamed to handleClickOnNewTabLink

PR #458, opened 2019-10-09
GH-1777 and GH-1776 Plus Promo modals

wlycdgr:
Implements the UI and business logic for the on-install Plus Promo modals described in GH-1776, GH-1777, and GH-1773. Provides a framework and a stub for the upgrade version of the modal described in GH-1813. Does not implement the metrics pings - that has been factored out to GH-1775 and will be a separate PR.

The main parts of the code are:
* a new shared component called PlusPromoModal that renders the initial and upgrade versions of the promo modal inside the Modal shared component
* the accompanying scss
* a new background class called PromoModals that manages modal-related state and answers questions like "is it time to display another promo?". It will be modified and extended to handle the Insights promo modal alongside the Plus ones. It is effectively a namespace for some static methods that query and set conf properties.

The Hub version of the modal is rendered in HomeViewContainer; the panel version is rendered in Panel. The components send messages to the main background script, which delegates their handling to PromoModals. PanelData queries PromoModals to set the properties it sends to Panel which determine whether the modal is rendered.

To faciliate testing, when debug is set in manifest, the modals should display every ~20 seconds instead of every 30 days.

P.S. The number of changed files is large because I snuck in some localization cleanup. A lot of it is test snapshots updated to match the new i18n token names.

P.P.S. Price localization will be implemented by calling `Number.toLocaleString()` on the string "$2", as per discussion with Jose. This will require some design updates from John to make sure the modals look good in every locale, as there's significant variation in how '$2' is localized (see GH-1777 for details). ~~Forgot to mention. There's a discussion about how to handle price localization for the "$2 per month" line in the modal. One suggestion is to try to infer location from IP. Another is to use USD $2 everywhere. Another option would be to use the most likely currency icon and amount combination for each locale.~~


wlycdgr:
@christophertino I've removed the console statement - should be ready to merge.

PR #497, opened 2020-02-10
GH-1477 Wildcard/Regex Whitelisting

benstrumeyer:
* [*] Have you followed the guidelines in [CONTRIBUTING.md](../CONTRIBUTING.md)?
* [*] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [*] Have you added an explanation of what your changes do?
* [*] Does your submission pass tests?
* [*] Did you lint your code prior to submission?

- Add wildcard and regex input for whitelist/blacklist
- Remove some validation since everything is a regex
- Does not include paygating

* Should we use https://www.npmjs.com/package/wildcard for wildcard matching?

Ticket: https://cliqztix.atlassian.net/browse/GH-1477

### DIFF
--- a/app/panel/components/Settings/TrustAndRestrict.jsx
+++ b/app/panel/components/Settings/TrustAndRestrict.jsx
@@ -12,6 +12,7 @@
  */
 
 import React from 'react';
+import safe from 'safe-regex';
 import Sites from './Sites';
 /**
  * @class Implement Trust and Restrict subview presenting the lists
@@ -119,10 +120,12 @@ class TrustAndRestrict extends React.Component {
 		pageHost = pageHost.toLowerCase().replace(/^(http[s]?:\/\/)?(www\.)?/, '');
 
 		// Check for Validity
-		if (pageHost.length >= 2083) {
+		if (pageHost.length >= 2083
+			|| !this.isValidUrlWildcardOrRegex(pageHost)) {
 			this.showWarning(t('white_black_list_error_invalid_url'));
 			return;
 		}
+
 		// Check for Duplicates
 		if (list.includes(pageHost)) {
 			this.showWarning(duplicateWarning);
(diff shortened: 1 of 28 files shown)


app/panel/components/Settings/TrustAndRestrict.jsx
```diff
@@ -86,9 +86,10 @@ class TrustAndRestrict extends React.Component {
 	/**
 	 * Implement adding site to the list of whitelisted or blacklisted sites
 	 * This routine valides entered url and checks if the entered url is a duplicate, or
-	 * if it has been alreday added to the opposite list. Displays appropriate warnings.
+	 * if it has been already added to the opposite list. Displays appropriate warnings.
 	 */
 	addSite() {
+		// TODO: Implement paygating, use this regex for validating free users who don't have regex/wildcard ability
```
Eden12345: Just curious as to why we aren't implementing the paygating in this PR?

benstrumeyer: Chris said don't worry about it for now. I think he wanted me to focus on implementation first.

Eden12345: Ah okay, then let's update the ticket to reflect that — unless @christophertino you would like me to pair up with @benstrumeyer and knock out the paygating real quick? It should be pretty straightforward since we have the account/subscription info in the `AccountReducer`.

Eden12345: @benstrumeyer Looks like @christophertino doesn't want this to be a Plus feature anyway, and he said he'd update the ticket to refelect that.

app/panel/components/Settings/TrustAndRestrict.jsx
```diff
@@ -86,9 +86,10 @@ class TrustAndRestrict extends React.Component {
...
 		// from node-validator
 		const isValidUrlRegex = /^(?!mailto:)(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?:(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]+-?)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]+-?)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))|localhost)(?::\d{2,5})?(?:\/[^\s]*)?$/i;
```
Eden12345: Let's remove this variable if we aren't going to be using it.

src/classes/Policy.js
```diff
@@ -174,6 +177,29 @@ class Policy {
...
 	}
+
+	/**
+	 * Check given url against pattern which might be a regex, or a wildcard
+	 * @param  {string} url		site url
+	 * @param  {string} pattern	regex pattern
+	 * @return {boolean}
+	 */
+	matchesWildcardOrRegex(url, pattern) {
+		let regex = escapeStringRegexp(pattern);
```
Eden12345: I'm not sure if I'm missing something here with the node module you're using, but when do we actually use this value that the var `regex` is initialized with?

Eden12345: Did you mean to pass the var `regex` to the first `RegExp` instantiation and the `wildcardPattern` var initialization? And if so, do you still need the try/catches?

Eden12345:
@benstrumeyer I'm also wondering if we need to add that new dependency since it's such a little amount of code. Do you want to just pull the code out of it and add that code directly into the function where you used the module?

src/classes/Policy.js
benstrumeyer: Yup, that's absolutely what I meant to do. I was also able to get rid of the try/catches because of the escape function you commented below. Thanks for the help.

Eden12345:
Looks good to me now, but requesting a review from @christophertino as I haven't touched this code much.

app/panel/components/Settings/TrustAndRestrict.jsx
```diff
@@ -86,11 +86,9 @@ class TrustAndRestrict extends React.Component {
 	/**
 	 * Implement adding site to the list of whitelisted or blacklisted sites
 	 * This routine valides entered url and checks if the entered url is a duplicate, or
-	 * if it has been alreday added to the opposite list. Displays appropriate warnings.
+	 * if it has been already added to the opposite list. Displays appropriate warnings.
 	 */
 	addSite() {
-		// from node-validator
-		const isValidUrlRegex = /^(?!mailto:)(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?:(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]+-?)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]+-?)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))|localhost)(?::\d{2,5})?(?:\/[^\s]*)?$/i;
```
christophertino: We still need front-end validation here so that we can prevent users from adding urls in unsupported formats. You should check that the url entered is either a valid url or a valid regex, and show an error if not.

benstrumeyer: I checked if it's a valid URL, wildcard, or regex and updated the error messages. Let me know if you'd like to change anything else

PR #509, opened 2020-03-09
GH-1969 Plus In App Promo & GH-1970 Telemetry

benstrumeyer:
* [*] Have you followed the guidelines in [CONTRIBUTING.md](../CONTRIBUTING.md)?
* [*] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [*] Have you added an explanation of what your changes do?
* [*] Does your submission pass tests?
* [*] Did you lint your code prior to submission?

- Makes spring in app promo for plus
- Replaces old midnight promo (same logic, appears once per month)
- "Upgrade to Plus" has a UTM parameter attached
- GBE pings are updated to include spring/leaf themes

Tickets: https://cliqztix.atlassian.net/browse/GH-1969, https://cliqztix.atlassian.net/browse/GH-1970


app/panel/components/Panel.jsx
```diff
@@ -369,6 +384,30 @@ class Panel extends React.Component {
...
 				handleXClick={this._handlePromoXClick}
+				show
+			/>
+		);
+	}
+
+	/**
+	 * @returns {null|JSX}
+	 * @private
+	 * Renders the Insights promo modal if the user is not already an Insights subscriber
```
Eden12345: Looks like this comment needs to be updated

app/panel/components/Panel.jsx
```diff
@@ -369,6 +384,30 @@ class Panel extends React.Component {
...
+		if (this._plusSubscriber() || this._premiumSubscriber()) { return null; }
+
+		sendMessage('promoModals.sawPlusPromo', {});
+		sendMessage('ping', 'promo_modals_show_plus');
+
+		return (
+			<PlusPromoModal
+				handleGoAwayClick={this._handlePromoGoAwayClick}
+				handleSignInClick={this._handlePromoSignInClick}
+				handleSubscribeClick={this._handlePromoSubscribeClick}
```
Eden12345: Won't `this._handlePromoSubscribeClick` open the checkout for Premium rather than plus?

app/panel/components/PlusPromoModal.jsx
```diff
@@ -0,0 +1,86 @@
...
+import ClassNames from 'classnames';
+import Modal from '../../shared-components/Modal';
+import ModalExitButton from './BuildingBlocks/ModalExitButton';
+
+/**
+ * A functional React component for a PlusPromo Modal that may be displayed in the Hub and/or Panel
+ * @return {JSX} JSX for rendering a PlusPromo Modal
+ * @memberof SharedComponents
+ */
+const PlusPromoModal = (props) => {
```
Eden12345: Is there a way we could combine `PlusPromoModal`, `InsightsPromoModal`, and `PremiumPromoModal`, or at least wrap them all in the same outer component that takes care of the functions that are essential to each (eg. `handleGoAwayClick`, `handleSignInClick`, `handleXClick`)?

src/classes/PromoModals.js
```diff
@@ -40,16 +42,18 @@ class PromoModals {
 	 */
 	static whichPromoModalShouldWeDisplay() {
```
Eden12345: Are we never displaying Premium modals anymore, then?

app/panel/components/Panel.jsx
benstrumeyer: Updated comment!

src/classes/PromoModals.js
benstrumeyer: That's the plan, we're also keeping the Premium modal code in there in case we want to switch back at some point

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,195 @@
...
+	};
+
+	_handlePromoXClick = (type) => {
+		this.props.actions.togglePromoModal();
+
+		if (type === INSIGHTS) {
+			sendMessage('ping', 'promo_modals_decline_insights_upgrade');
+		}
+
+		if (type === PLUS) {
```
Eden12345, line 116: An `else if` statement might make more sense here.

app/panel/components/InsightsPromoModalContent.jsx
```diff
@@ -0,0 +1,78 @@
+/**
+ * Insights Promo Modal Component
```
Eden12345, line 2: Is this component not with the other modal content components because it is only used in the panel? Even if that's the case, I would opt for keeping them together in one `modal-content` directory in `shared-components`.

app/panel/components/InsightsPromoModalContent.jsx
```diff
@@ -0,0 +1,78 @@
...
+const InsightsPromoModalContent = (props) => {
+	const {
+		handleGoAwayClick,
+		handleSignInClick,
+		handleTryInsightsClick,
+		XButton,
+	} = props;
+	return (
+		<div className="InsightsModal__content flex-container flex-dir-column align-middle">
+			{XButton}
```
Eden12345, line 25: I'm still confused as to why this outer `<div />` and `<ExitButton />` can't be moved out of each individual modal content component and into the parent `<PromoModal />` component, as you were saying. From what I can tell, the only difference between them is the main class name (which you can determine in `<PromoModal />` with the `classnames` library), and the `<ExitButton />` click handler (which you're already determining in `<PromoModal />`, but then pass down to the child components).

app/shared-components/PremiumPromoModalContent/PremiumPromoModalContent.jsx
```diff
@@ -0,0 +1,129 @@
...
+		location,
+		XButton
+	} = props;
+
+	const contentClassNames = ClassNames(
+		'PremiumPromoModal__content',
+		'flex-container',
+		'flex-dir-column',
+		'align-middle',
+	);
```
Eden12345, line 39: Also, you don't need to use the `classnames` library if none of your class assignments are dependent on logic.

app/panel/components/Panel.jsx
benstrumeyer: fixed!

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: else if does make more sense here

app/panel/components/InsightsPromoModalContent.jsx
benstrumeyer: The `PremiumPromoModal` is the only one that's being used in the hub. But yeah it makes more sense just to bag them all together in there too since it might be an option at one point. Moved!

app/panel/components/InsightsPromoModalContent.jsx
benstrumeyer: Moved the ExitButton out! Making a fragment inside each `PromoModalContent` class messed up the CSS, but the Exit button is inside the PromoModal.jsx component nonetheless!

wlycdgr:
"No thanks, continue with Basic" does not dismiss the popup (at least not in the Hub, on first install)

wlycdgr:
(Also, is the user still supposed to see the Midnight promo modal in the hub?)

app/panel/containers/PlusPromoModalContainer.js
```diff
@@ -0,0 +1,28 @@
...
+ * @memberOf PanelContainers
+ * @param  {Object} state    entire Redux store's state
+ * @param  {Object} ownProps props passed to the connected component
+ * @return {function}        this function returns plain object, which will be merged into Panel view component props
+ * @todo  We are not using ownProps, so we better not specify it explicitly,
+ * in this case it won't be passed by React (see https://github.com/reactjs/react-redux/blob/master/docs/api.md).
+ */
+const mapStateToProps = state => Object.assign({}, state.account, {});
+
+export default withRouter(connect(mapStateToProps)(PlusPromoModal));
```
wlycdgr, line 28: I think we can delete this, right?

app/scss/partials/_modal_exit_button.scss
```diff
@@ -19,9 +19,14 @@
 	width: 26px;
 	height: 26px;
 	border-radius: 15px;
-	border: solid 0.8px #325e97;
 	background-color: #f7f7f7;
 	@include transition(background-color 0.2s);
+	&.grey {
+		border: solid 0.8px #325e97;
+	}
```
wlycdgr, line 26: See my comment on `ModalExitButton` to make sure that this `grey` class gets applied when needed.

app/panel/components/BuildingBlocks/ModalExitButton.jsx
```diff
@@ -22,19 +23,30 @@ import PropTypes from 'prop-types';
...
 	const {
-		toggleModal
+		toggleModal,
+		border
 	} = props;
 
+	const borderClassNames = ClassNames('ModalExitButton__exit flex-container align-middle', {
+		green: border === 'green'
+	});
+
```
wlycdgr, line 33: Instead of `green`, let's just apply the value of `border` - this also makes sure that `grey` is applied in the default case.

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,201 @@
...
+import globals from '../../../src/classes/Globals';
+import ModalExitButton from '../../panel/components/BuildingBlocks/ModalExitButton';
+
+const DOMAIN = globals.DEBUG ? 'ghosterystage' : 'ghostery';
+const INSIGHTS = 'insights';
+const PLUS = 'plus';
+const PREMIUM = 'premium';
+
+/**
+ * A base functional component for Promo Modals
```
wlycdgr, line 32: Update description to reflect that this is a class component, not a functional one

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,201 @@
...
+
+		switch (modal) {
+			case INSIGHTS:
+				sendMessage('ping', 'promo_modals_decline_insights_upgrade');
+				break;
+			case PLUS:
+				sendMessage('ping', 'promo_modals_decline_plus_upgrade');
+				break;
+			case PREMIUM:
+				sendMessage('ping', 'promo_modals_decline_insights_upgrade');
```
wlycdgr, line 56: Should this be `.....decline_premium_upgrade`?

Zooming out, perhaps instead of a switch we can simply send `promo_modals_decline_${modal}_upgrade`?

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,201 @@
...
+		let url;
+		switch (product) {
+			case PLUS:
+				// Should we send a plus ping here?
+				// sendMessage('ping', 'promo_modals_plus_upgrade_cta');
+				url = `https://checkout.${DOMAIN}.com/plus?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				break;
+			case PREMIUM:
+				url = `https://ghostery.com/thanks-for-downloading-midnight?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				break;
```
wlycdgr, line 85: Should we send `promo_modals_[modal_type]_upgrade_cta` pings for plus and for premium like we do for insights?

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,201 @@
...
+	};
+
+	_handlePromoXClick = (type) => {
+		this.props.actions.togglePromoModal();
+
+		if (type === INSIGHTS) {
+			sendMessage('ping', 'promo_modals_decline_insights_upgrade');
+		} else if (type === PLUS) {
+			sendMessage('ping', 'promo_modals_decline_plus_upgrade');
+		}
```
wlycdgr, line 117: Do we need to send a decline ping for Premium too?

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,201 @@
...
+						handleSignInClick={this._handlePromoSignInClick}
+						loggedIn={loggedIn}
+					/>
+				);
+			case PREMIUM:
+				return (
+					<PremiumPromoModalContent
+						handleGoAwayClick={() => this._handlePromoGoAwayClick(PREMIUM)}
+						handleTryMidnightClick={() => this._handlePromoTryProductClick(PREMIUM, 'in_app')}
+						handleGetPlusClick={() => this._handlePromoTryProductClick(PLUS, 'in_app')}
```
wlycdgr, line 157: Should first arg be `PREMIUM`?

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -0,0 +1,201 @@
...
+						handleTryInsightsClick={() => this._handlePromoTryProductClick(INSIGHTS, 'in_app_upgrade')}
+						handleSignInClick={this._handlePromoSignInClick}
+						{...this.props}
+					/>
+				);
+			case PLUS:
+				return (
+					<PlusPromoModalContent
+						handleGoAwayClick={() => this._handlePromoGoAwayClick(PLUS)}
+						handleTryPlusClick={() => this._handlePromoTryProductClick(PLUS, 'in_app_spring2020')}
```
wlycdgr, line 147: @Eden12345 @benstrumeyer It would be very good if we could avoid hardcoding a specific campaign into this generic promo modal rendering code. As is, we'll have to do a whole release if we want to end this campaign or run a different one. We should be fetching the campaign identifier(s) from a remote url that Teresa & co can update independently. If we think that's outside the scope of this release, I think we should make a ticket for it and try to get it into the next release.

src/classes/Metrics.js
```diff
@@ -510,10 +510,16 @@ class Metrics {
 	 */
 	_getThemeValue() {
 		const { current_theme } = conf;
-		if (current_theme === 'midnight-theme') {
-			return 1;
+		switch (current_theme) {
+			case 'midnight-theme':
```
wlycdgr, line 514: Nothing to change here & now, but just a note that we should remember to update this bit to `dark-blue-theme` in the final theme framework PR

app/panel/containers/PlusPromoModalContainer.js
benstrumeyer: Definitely can delete it

app/scss/partials/_modal_exit_button.scss
benstrumeyer: Good idea, we can just apply the value of border here

app/panel/components/BuildingBlocks/ModalExitButton.jsx
benstrumeyer: Perfect

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: Updated comment to a class component

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: It looks like we currently don't have a metrics ping case for this so I reached out to Teresa and she said to hold off on adding extra pings for now since wants to sync with JJ first. I refactored the plus and insights pings to use the template literal notation for now.

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: Holding off on adding extra pings for now per teresa

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: Same as above

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: I think the handleGetPlusClick on the premium promo modal should lead to the checkout page for plus if I'm not mistaken? I wound up reusing the `_handlePromoTryProductClick()` function because it doesn't send UTM params

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: I pushed for this idea in the extension sprint meeting and it's my understanding that there's a Firefox clause preventing us from hooking up the extension to a remote URL that updates UTM params on the fly. We can sidebar this with chris since teresa was very interested in the idea.

src/classes/Metrics.js
benstrumeyer: I spoke with Frank a while ago and I think we're only visually updating the theme to dark-blue-theme and not the theme variables themselves so we can prevent breakage for users on older versions of the extension that request a midnight-theme from the account project which would no longer exist. We do however have a ticket in the backlog to do this properly

app/shared-components/PromoModal/PromoModal.jsx
wlycdgr: Ah, dang. Sounds like that Firefox policy is a dealbreaker. Too bad.

app/shared-components/PromoModal/PromoModal.jsx
wlycdgr: My bad, I forgot that the Premium modal has an option to get Plus instead. Looks good.

PR #518, opened 2020-03-26
GH-1983: Handle opt-in settings for legacy users on FF

christophertino:


src/background.js
```diff
@@ -1487,13 +1489,12 @@ function initializeVersioning() {
 				conf.enable_smart_block = false;
 			}
 
-			// Are we upgrading from Ghostery 8 prior to 8.2?
-			if ((+prevVersion[0] === 8) && (prevVersion[1] < 2)) {
-				globals.JUST_UPGRADED_FROM_8_1 = true;
+			// Check if the earliest version is < 8.4.2
+			if (earliestVersion[0] <= 8 && earliestVersion[1] <= 4 && earliestVersion[2] < 2) {
```
jsignanini: @christophertino to be extra safe we should add a length check on the `earliestVersion` array so it doesn't crash if it's `< 3`.

PR #510, opened 2020-03-12
GH-1967 Themes Panel Update

benstrumeyer:
* [*] Have you followed the guidelines in [CONTRIBUTING.md](../CONTRIBUTING.md)?
* [*] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [*] Have you added an explanation of what your changes do?
* [*] Does your submission pass tests?
* [*] Did you lint your code prior to submission?

- Change theme toggle to radio buttons inside panel
- Visually changes "Midnight Theme" to "Dark Blue Theme"
  - Variable name changes will be reflected in GH-1974 and GIS-545
- Add palm/leaf theme radio buttons that currently do nothing

Ticket: https://cliqztix.atlassian.net/browse/GH-1967

PR #525, opened 2020-04-14
GH-1975 Theme Framework, Leaf & Palm Themes (FINAL)

benstrumeyer:

- **Can be merged!**
- This PR points to production servers and has all of the necessary extension changes to display the leaf and palm themes that are built off the theme framework which lives inside the account project
- GH-1967 Themes Panel Update #510 has been merged into this branch, but still needs to be tested
- GH-1969 Plus In App Promo & GH-1970 Telemetry that displays modals will be a separate PR


app/panel/utils/utils.js
```diff
@@ -218,10 +218,11 @@ export function setTheme(doc, name, account) {
 	// First remove all other style elements which may be there
```
wlycdgr, line 218: Remove or correct this comment (and also let's move the previous one at the start of the function down to the appropriate place)

app/panel/utils/utils.js
```diff
@@ -218,10 +218,11 @@ export function setTheme(doc, name, account) {
...
 	const styleList = doc.head.getElementsByTagName('style');
 	// Other kinds of loops are not supported equally across browsers
+	let themeStyle = null;
 	for (let i = 0; i < styleList.length; i++) {
 		const style = styleList[i];
 		if (style.title.startsWith(styleTitlePrefix)) {
-			doc.head.removeChild(style);
+			themeStyle = style;
 		}
 	}
```
wlycdgr, line 227: We can remove this comment and replace with the tidier `forEach`, which is now supported in all our supported browser versions

app/panel/utils/utils.js
```diff
@@ -232,12 +233,21 @@ export function setTheme(doc, name, account) {
...
-		document.head.appendChild(themeStyle);
+		if (!themeStyle) {
+			themeStyle = doc.createElement('style');
+			themeStyle.id = name;
+			themeStyle.title = `${styleTitlePrefix}`;
+			themeStyle.textContent = css;
+			doc.head.appendChild(themeStyle);
+		} else {
+			themeStyle.textContent = css;
+		}
```
wlycdgr, line 244: Simplify by removing the `else` and putting `themeStyle.textContent = css` below the conditional block

app/panel/utils/utils.js
```diff
@@ -232,12 +233,21 @@ export function setTheme(doc, name, account) {
...
+	} else {
+		for (let i = 0; i < styleList.length; i++) {
+			const style = styleList[i];
+			if (style.title.startsWith(styleTitlePrefix)) {
+				doc.head.removeChild(style);
+			}
+		}
```
wlycdgr, line 251: Here too we can now use `forEach`

app/panel/utils/utils.js
wlycdgr, line 244: We can simplify by removing the `else` and moving `themeStyle.textContent` below the `if` block

fcjr: @wlycdgr line 241 requires themeStyle it cant be rearranged like this.

wlycdgr: @fcjr sorry if I hallucinated you commenting on this, but I think we can set `textContent` after appending, no?

Eden12345:
@wlycdgr I suggested that we add the background PNG for one of the themes to the extension in a comment a few PRs ago, but when you put it that way (especially in regards to future theme releases), I agree with you.

@benstrumeyer Would it be a huge lift to switch back to the way you had it when you were passing the PNG with the themed Sass file?

app/panel/utils/utils.js
benstrumeyer: Updated comment to "Get style tag for the active theme", and moved the first comment

app/panel/utils/utils.js
benstrumeyer: Good to know, changed to a forEach loop

app/panel/utils/utils.js
benstrumeyer: Changed to a forEach loop

app/panel/utils/utils.js
benstrumeyer: Simplified, I left the `themeStyle.textContent = css` line inside the if block as is in case there's any funky stuff going on with that content not being applied before it gets appended

app/panel/utils/utils.js
benstrumeyer: I stand corrected, we're unable to use a regular forEach loop because getElementsByTagName returns an HTML collection. I've changed it to `Array,prototype.forEach.call(styleList, (style) =>...` as suggested by the following stackoverflow thread unless you think keeping a basic for loop is cleaner. Thoughts?

https://stackoverflow.com/questions/39797101/why-cant-i-use-array-foreach-on-a-collection-of-javascript-elements

app/panel/utils/utils.js
benstrumeyer: Same issue as above

app/panel/utils/utils.js
wlycdgr: Ah dang, right, sorry. We can use `Array.from(styleList).forEach`

PR #527, opened 2020-04-17
GH-1979 Allow user selection of Ad Block filter lists

### LINKED ISSUE #504: The New Yorker: Annoying Banner
### Description
The New Yorker has an annoying banner that takes up half the page when you click on an article asking the user to subscribe. Please remove.

### Expected Behavior 

Force Ghostery to hide the banner.

### Steps to Reproduce

1. Go to `https://www.newyorker.com/`
2. Click on any article.
3. https://postimg.cc/yWGFVxdt

### Versions
Applies to any browser, but here is my configuration:
+ Browser: Firefox 73.0.1 (current stable release); Ghostery is set to block everything in all categories provided.
+ OS: macOS 10.15.3
+ Node:
+ NPM:

christophertino:
Adds a new component in the Settings panel with Ad Blocker list selection options.  

Fixes #504


app/panel/components/Settings/SettingsMenu.jsx
```diff
@@ -14,6 +14,10 @@
 import React from 'react';
 import { NavLink } from 'react-router-dom';
 import ClassNames from 'classnames';
+import globals from '../../../../src/classes/Globals';
+
+const IS_CLIQZ = (globals.BROWSER_INFO.name === 'cliqz');
```
wlycdgr, line 19: Why not use `globals.IS_CLIQZ`?

christophertino: good call, that's in a few other places too.  let me remove.

yokoffing:
Thank you for finally implementing this! I feel more confident in recommending Ghostery now that annoyances are being addressed.

PR #546, opened 2020-05-08
GH-2048: Display premium icons & subscription info & GH-2036: Unlock Plus features on GBE/Midnight for all Premium & Plus users

benstrumeyer:

GH-2048
- Display respective premium/plus icons for subscribed users
- Upgrade account api to v2.1.0
- Display subscription info for premium if a user has both a premium and plus subscription
- Add premium icon in menu
- The colored premium badge icon in the designs is not used as intended since we are upselling plus to users without a subscription

GH-2036
- Allow premium users to view the subscription panel when clicking the subscription badge
- Allow premium users to view the historical stats graph

Tickets:
- https://cliqztix.atlassian.net/browse/GH-2048
- https://cliqztix.atlassian.net/browse/GH-2036


src/background.js
```diff
@@ -824,19 +824,7 @@ function onMessageHandler(request, sender, callback) {
...
-				}
-				const subscriptionData = sub.reduce((acc, curr) => {
-					let a = acc;
-					if (curr.productName.includes('Plus')) {
-						a = curr;
-					}
-					return a;
-				}, {});
-				callback({ subscriptionData });
+				callback({ subscriptionData: customer });
```
wlycdgr, line 827: `Account#getUserSubscriptionData` now returns a subscription object if the user is plus/premium, or the customer object if user is neither. So it looks like we have changed what we pass to the callback if the user is not a subscriber. Is this as intended?

app/panel/components/Summary.jsx
```diff
@@ -312,6 +312,11 @@ class Summary extends React.Component {
 		return user && user.subscriptionsPlus;
 	}
 
+	_isPremiumSubscriber() {
+		const { user } = this.props;
+		return user && user.scopes && user.scopes.includes('subscriptions:premium');
+	}
+
 	_pageHost() {
```
wlycdgr, line 320: Mimicking the `subscriptionsPlus` approach allows us to abstract away this implementation detail here as well

app/panel/components/HeaderMenu.jsx
```diff
@@ -230,13 +234,27 @@ class HeaderMenu extends React.Component {
...
-									</g>
-								</svg>
-								<span>{ t('ghostery_plus') }</span>
+								{/* Upselling plus for users who are not logged in */}
```
wlycdgr, line 237: This comment doesn't seem to match what's right below - couldn't user still be logged in and a plus subscriber even if their `subscriberType` isn't premium?

app/panel/components/Header.jsx
```diff
@@ -145,12 +145,13 @@ class Header extends React.Component {
...
-	clickUpgradeBannerOrGoldPlusIcon = () => {
+	clickUpgradeBannerOrSubscriberBadgeIcon = () => {
 		// TODO check whether this is the message we want to be sending now
 		sendMessage('ping', 'plus_panel_from_badge');
 		const { user } = this.props;
-		const subscriber = user && user.subscriptionsPlus;
-		this.props.history.push(subscriber ? '/subscription/info' : `/subscribe/${!!user}`);
+		const plusSubscriber = user && user.subscriptionsPlus;
+		const premiumSubscriber = user && user.scopes && user.scopes.includes('subscriptions:premium');
+		this.props.history.push(plusSubscriber || premiumSubscriber ? '/subscription/info' : `/subscribe/${!!user}`);
```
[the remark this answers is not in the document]
Eden12345: +1, this is how we handle subscriptions on the frontend in Midnight as well.

app/panel/components/Header.jsx
```diff
@@ -176,11 +177,12 @@ class Header extends React.Component {
...
 		});
-		const subscriber = user && user.subscriptionsPlus;
+		const plusSubscriber = user && user.subscriptionsPlus;
+		const premiumSubscriber = user && user.scopes && user.scopes.includes('subscriptions:premium');
 		const accountLogolink = this.generateAccountLogo();
 		const badgeClasses = ClassNames('columns', 'shrink', {
-			'non-subscriber-badge': !subscriber,
-			'gold-subscriber-badge': subscriber
+			'non-subscriber-badge': !(plusSubscriber || premiumSubscriber),
+			'gold-subscriber-badge': plusSubscriber || premiumSubscriber
```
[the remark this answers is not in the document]
benstrumeyer: Renamed to subscriber badge!

app/panel/components/Header.jsx
```diff
@@ -215,11 +217,17 @@ class Header extends React.Component {
...
+		} else if (plusSubscriber) {
+			subscriberType = 'plus';
+		}
 		const plusUpgradeBannerOrSubscriberBadgeLogolink = (
-			<div className={badgeClasses} onClick={this.clickUpgradeBannerOrGoldPlusIcon}>
+			<div className={badgeClasses} onClick={this.clickUpgradeBannerOrSubscriberBadgeIcon}>
 				{
-					(subscriber && <ReactSVG src="/app/images/panel/gold-plus-icon-expanded-view.svg" />) ||
-					<ReactSVG src="/app/images/panel/green-upgrade-banner-expanded-view.svg" />
+					((premiumSubscriber || plusSubscriber) && <ReactSVG src={`/app/images/panel/${subscriberType}-badge-icon-expanded-view.svg`} />)
```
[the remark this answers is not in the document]
benstrumeyer: That makes a lot of sense--will keep in mind for future image pathing

app/panel/components/HeaderMenu.jsx
benstrumeyer: Yup, that's true. Changed to "Upselling plus for all users who are not premium subscribers" which much better represents the behavior

src/background.js
benstrumeyer: Ah. I was trying to reduce code duplication between account#getUserSubscriptionData and the background, but it looks like I got a little too fancy since that function is used in several places in the background class. Changed function back to it's original behavior

app/panel/components/Header.jsx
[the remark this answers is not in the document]
benstrumeyer: Ah yeah, it makes way more sense to have these consistent. I added a subscriptionsPremium property on the user object and got rid of subscriptionType

app/panel/components/HeaderMenu.jsx
```diff
@@ -154,17 +154,21 @@ class HeaderMenu extends React.Component {
 	clickSubscriber = () => {
 		sendMessage('ping', 'plus_panel_from_menu');
 		this.props.toggleDropdown();
-		this.props.history.push(this.props.subscriber ? '/subscription/info' : `/subscribe/${this.props.loggedIn}`);
+		this.props.history.push(this.props.subscriberType ? '/subscription/info' : `/subscribe/${this.props.loggedIn}`);
```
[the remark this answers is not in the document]
benstrumeyer: Changed to check if a user has a premium or plus subscription!

app/panel/components/Summary.jsx
benstrumeyer: The `_isPremiumSubscriber()` now mimics the `_isPlusSubscriber` function

src/classes/Account.js
```diff
@@ -163,19 +163,23 @@ class Account {
...
+				// Display premium info if user has both premium and plus subscriptions
+				const premiumSubscription = sub.find(subscription => subscription.productName.includes('Ghostery Premium'));
+				if (premiumSubscription) {
+					this._setSubscriptionData(premiumSubscription);
+				}
+
+				const plusSubscription = sub.find(subscription => subscription.productName.includes('Ghostery Plus'));
+				if (plusSubscription) {
+					this._setSubscriptionData(plusSubscription);
+				}
```
wlycdgr, line 182: Even though `account-web` call this function indirectly via the `accountPage.getuserSubscriptionData` message, it does not appear to actually use or even capture the return value. So the only (indirect, via the `account.getUserSubscriptionData` message handler in `background.js`) consumer of the return value is the GBE panel, via the `Account` component & reducer. So this is the only data flow out of the function that we need to be concerned with.

In this data flow, the message handler in `background.js` throws away most of the returned `customer` object, extracting the `Ghostery Premium` or `Ghostery Plus` entry from the `customer.subscriptions` array and sending just that back to the `Account` component in `app`, which is just what that component expects.

So, particularly since the extraction is duplicated here and in `background.js`, we can instead return just the `premiumSubscription`/`plusSubscription` object from this function directy, or else an empty object if neither subscription type is found.

Then, the handler in `background.js` can be simplified to

```
	if (name === 'account.getUserSubscriptionData') {
		account.getUserSubscriptionData()
			.then(subscriptionData => callback({ subscriptionData }))
			.catch((err) => {
				log('Error getting user subscription data:', err);
				callback({ errors: _getJSONAPIErrorsObject(err) });
			});
		return true;
	}
```

@christophertino  - would appreciate your input on whether this sounds like a safe change to make and also on whether you think we need to update what is being passed to the front end to reflect that the user could now have multiple subscriptions active simultaneously, or whether it's sufficient to pass the highest-tier subscription, allowing us to leave the `subscriptionData` Redux object and logic unchanged on the front end for now.

src/background.js
```diff
@@ -829,13 +829,13 @@ function onMessageHandler(request, sender, callback) {
...
-					return a;
-				}, {});
+
+				let subscriptionData = sub.find(subscription => subscription.productName.includes('Ghostery Premium'));
+				if (subscriptionData) {
+					callback({ subscriptionData });
+				}
+
+				subscriptionData = sub.find(subscription => subscription.productName.includes('Ghostery Plus'));
 				callback({ subscriptionData });
```
wlycdgr, line 839: See my comment on `Account#getUserSubscriptionData`. Note that even if we don't make that change, we need to make sure we pass an empty object to the callback, and not `undefined` (the fail return value of `Array.prototype.find`), in the event that neither subscription is found.

src/background.js
wlycdgr: See my further comments below :)

src/classes/Account.js
wlycdgr: Also, if we don't do this change, we need to make sure we don't clobber the Premium subscription set with the Plus subscription set in case a user somehow has both.

wlycdgr: Based on feedback from @christophertino , let's do a blend of what I suggested and what we have now. Instead of returning the whole `customer` object or just the highest tier sub, let's return an array with all the subscription objects (or an empty array, in case the user has neither sub type). So the function guarantees it returns an array. Then in the handler in `background`, we can extract the highest tier sub object from that array and return that, or else an empty object if the array is empty.

app/panel/components/HeaderMenu.jsx
```diff
@@ -230,13 +243,27 @@ class HeaderMenu extends React.Component {
...
-									</g>
-								</svg>
-								<span>{ t('ghostery_plus') }</span>
+								{/* Upselling plus for all users who are not premium subscribers */}
```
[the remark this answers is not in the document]
benstrumeyer: Ah okay. Will make comments as clear and descriptive as possible from now on

src/classes/Account.js
benstrumeyer: Hey, thanks for digging into this and helping me understand why we can make this change. Love how you found a way to keep the return type of `account#getUserSubscriptionData` consistent as an array and allowed the respective handlers to be flexible

src/background.js
benstrumeyer: Changed to return only the highest tier subscription, and made sure to return an empty object instead of `undefined`!

app/panel/components/HeaderMenu.jsx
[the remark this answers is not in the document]
wlycdgr: Awesome, thank you for your patience with my nittiness about this

src/classes/Account.js
wlycdgr: Sure thing, this was fun to think about

src/background.js
```diff
@@ -823,20 +823,15 @@ function onMessageHandler(request, sender, callback) {
...
-				let subscriptionData = sub.find(subscription => subscription.productName.includes('Ghostery Premium'));
-				if (subscriptionData) {
-					callback({ subscriptionData });
-				}
+				const plusSubscription = subscriptions.find(subscription => subscription.productName.includes('Ghostery Plus'));
+				if (plusSubscription) callback({ subscriptionData: plusSubscription });
 
-				subscriptionData = sub.find(subscription => subscription.productName.includes('Ghostery Plus'));
-				callback({ subscriptionData });
+				callback({});
```
wlycdgr: Invoking the callback doesn't cause the function to return, so we need to add explicit return statements after the Premium and Plus callbacks. Looks good otherwise!

src/classes/Account.js
```diff
@@ -158,6 +158,9 @@ class Account {
 			})
 	)
 
+	/**
+	 * @return {array}	All subscriptions the user has, empty if none
+	*/
```
wlycdgr, line 163: Good idea to add this!

src/classes/Account.js
```diff
@@ -170,17 +173,20 @@ class Account {
...
+					subscriptions.push(premiumSubscription);
 				}
 
 				const plusSubscription = sub.find(subscription => subscription.productName.includes('Ghostery Plus'));
 				if (plusSubscription) {
+					subscriptions.push(plusSubscription);
 					this._setSubscriptionData(plusSubscription);
 				}
-				return customer;
+				return subscriptions;
```
wlycdgr: Looking much better, just a couple more things: 
* We can simplify/dedupe this by usng `forEach` to iterate once over the `sub` array, checking whether the `productName` of each element is in a `productNames` array we define, and pushing the element to the `subscriptions` array if so.
* We need to make sure we don't clobber Premium subsciption data with Plus subscription data in `_setSubscriptionData` in case the user has both and the Plus sub is found after the Premium one.

src/background.js
benstrumeyer: Added explicit returns!

src/classes/Account.js
wlycdgr: Let's not worry about the `forEach` change for now

app/panel/components/Detail.jsx
```diff
@@ -60,13 +60,14 @@ class Detail extends React.Component {
...
+			condensed: is_expanded,
 		});
 
-		const activeTab = this.props.history.location.pathname.includes('rewards') ? 'rewards' : 'blocking';
+		const activeTab = history.location.pathname.includes('rewards') ? 'rewards' : 'blocking';
 		const contentDetailsClassNames = ClassNames({
-			expanded: this.props.is_expanded,
+			expanded: is_expanded,
 			rewardsView: activeTab === 'rewards',
 		});
```
wlycdgr, line 73: I'm feelin' all the destructurin'

app/panel/components/Panel.jsx
```diff
@@ -217,7 +217,7 @@ class Panel extends React.Component {
 	_premiumSubscriber = () => {
 		const { loggedIn, user } = this.props;
 
-		return loggedIn && (user && user.subscriptionsPremium);
+		return loggedIn && (user && user.premiumAccess);
 	}
 
 	/**
```
wlycdgr: Can we rename this and the  matching `_plusSubscriber` helper to reflect that they now return access status rather than subscription status?

app/panel/components/Stats.jsx
```diff
@@ -517,7 +517,7 @@ class Stats extends React.Component {
 		return selectionData;
 	}
 
-	_isPlus = props => props.user && props.user.subscriptionsPlus;
+	_isPlus = props => props.user && props.user.plusAccess;
```
wlycdgr, line 520: Similarly, let's rename this helper to `_hasPlusAccess`

app/panel/components/Stats.jsx
```diff
@@ -329,7 +329,7 @@ class Stats extends React.Component {
 			monthlyAverageData: clearData,
 			dailyAverageData: clearData,
 			showResetModal: false,
-			showPitchModal: (!this.props.user || !this.props.user.subscriptionsPlus),
+			showPitchModal: (!this.props.user || !this.props.user.plusAccess),
```
wlycdgr, line 332: While you're in there, can you destructure `this.props.user` like you've done elsewhere?

app/panel/components/Summary.jsx
```diff
@@ -752,34 +752,34 @@ class Summary extends React.Component {
 	 * @return {JSX} JSX for rendering the plus upgrade banner or subscriber icon
 	 */
 	_renderPlusUpgradeBannerOrSubscriberIcon() {
-		const { is_expert, current_theme } = this.props;
+		const { is_expert, current_theme, user } = this.props;
 
-		const isPremiumSubscriber = this._isPremiumSubscriber();
-		const isPlusSubscriber = this._isPlusSubscriber();
+		const hasPremiumAccess = user && user.premiumAccess;
+		const hasPlusAccess = user && user.plusAccess;
```
wlycdgr, line 758: Man I'm SO ready for optional chaining to be a thing

src/classes/Account.js
```diff
@@ -493,7 +494,6 @@ class Account {
 			conf.paid_subscription = true;
 			dispatcher.trigger('conf.save.paid_subscription');
 		}
-		conf.account.subscriptionData = data || null;
```
wlycdgr, line 484: I think we still need to save this, no?

src/classes/PanelData.js
```diff
@@ -264,8 +264,9 @@ class PanelData {
 	_getCurrentAccount() {
 		const currentAccount = conf.account;
 		if (currentAccount && currentAccount.user) {
-			currentAccount.user.subscriptionsPlus = account.hasScopesUnverified(['subscriptions:plus']);
-			currentAccount.user.subscriptionsPremium = account.hasScopesUnverified(['subscriptions:premium']);
+			currentAccount.user.plusAccess = account.hasScopesUnverified(['subscriptions:plus'])
+													|| account.hasScopesUnverified(['subscriptions:premium']);
+			currentAccount.user.premiumAccess = account.hasScopesUnverified(['subscriptions:premium']);
```
wlycdgr, line 269: I still wanna move this into `Account` but let's worry about that after. We can sneak it into one of the related tickets once this is in testing.

app/panel/components/Panel.jsx
benstrumeyer: Changed to `_hasPlusAccess()` and `_hasPremiumAccess()`!

app/panel/components/Stats.jsx
benstrumeyer: Renamed!

app/panel/components/Stats.jsx
benstrumeyer: Done!

app/panel/components/Summary.jsx
benstrumeyer: Ooh, I've never heard of optional chaining. Looks interesting

src/classes/Account.js
benstrumeyer: Thank god for PR reviews, my bad!

src/classes/PanelData.js
benstrumeyer: I'll make a note!

app/panel/components/Panel.jsx
```diff
@@ -237,17 +237,17 @@ class Panel extends React.Component {
...
 
-		const isPlus = this._plusSubscriber();
+		const hasPlusAccess = this.hasPlusAccess();
 
 		return (
 			<PromoModalContainer
 				type={PREMIUM}
 				location="panel"
-				isPlus={isPlus}
+				isPlus={hasPlusAccess}
```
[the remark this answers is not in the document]
benstrumeyer: Gotcha, I will be more thorough!

PR #551, opened 2020-05-18
GH-2070: Main screen promotion box

benstrumeyer:

- Create main screen promo box as per design specs in GH-2070
- Break responsively where the tutorial/customize setup div breaks for fluidity instead of 768 x 1040 as mentioned in design specs
- Merge in GH-2048 (which has already been reviewed) to take advantage of `user.premiumAccess` property
- CTA button navigates to non-existant `/update` route, will update when it is created

Ticket: https://cliqztix.atlassian.net/browse/GH-2070


app/hub/Views/HomeView/HomeView.scss
```diff
@@ -73,14 +73,25 @@
...
+	display: flex;
 	margin-top: 22px;
 	padding: 22px 12px;
 	background-color: #f0f3f4;
 	border-radius: 4px;
+
+	&.purple-border {
+		box-shadow: 0 0 8px 5px #920094;
+		border: solid 1.6px #920094;
+	}
```
wlycdgr, line 86: Let's use the `$dark-magenta` variable defined in `_colors.scss`. If you feel like converting the rest of this file to use the other predefined `_colors.scss` variables while you're at it, I certainly won't stop you.

wlycdgr: I am not sure why this stylesheet wasn't already using the variables - they should be available as they get pulled in through `settings_hub` in `hub.scss`.

app/hub/Views/HomeView/HomeView.scss
```diff
@@ -73,14 +73,25 @@
...
 }
-.HomeView__plusFeature {
-	padding: 0 36px;
+.HomeView__upgradeIcon {
+	background-image: upgradeShieldIcon(#DAF4FF, #4a4a4a);
+	margin: 0 42px;
+	height: 50px;
+	width: 50px;
+	background-repeat: no-repeat;
```
wlycdgr, line 93: I think you can cut the `background-repeat` prop if you just bump the `width` down by 1 pixel

app/hub/Views/HomeView/HomeView.scss
```diff
@@ -73,14 +73,25 @@
...
+	background-position: center center;
```
wlycdgr, line 94: Looks like you can cut `background-position` too.

src/utils/api.js
```diff
@@ -154,17 +154,17 @@ class Api {
 
 	get = (type, id, include = '') => {
 		if (!id) { return Promise.reject(new Error('id is missing')); }
-		return this._sendAuthenticatedRequest('GET', `/api/v2/${type}/${id}?${include ? `include=${include}` : ''}`);
+		return this._sendAuthenticatedRequest('GET', `/api/v2.1.0/${type}/${id}?${include ? `include=${include}` : ''}`);
```
wlycdgr, line 157: @christophertino Should we pull the API version in from a global that we set by calling account at startup so that we can upgrade the version without having to do an extension release?

wlycdgr: Leaving as is

app/hub/Views/HomeView/HomeView.scss
benstrumeyer: Definitely should be using color variables--updated!

app/hub/Views/HomeView/HomeView.scss
benstrumeyer: Good solution, i've seen that used a few places around the extension, but didn't know why

app/hub/Views/HomeView/HomeView.scss
benstrumeyer: Cut!

PR #569, opened 2020-06-16
Cross platform extension

sammacbeth:
Preparing for unified extension build for desktop and android:

 - [x] Dynamic loading of modules based on platform
 - [x] Include browser-core build with -lite modules included.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -16,7 +16,7 @@
  */
 import { debounce, every, size } from 'underscore';
 import moment from 'moment/min/moment-with-locales.min';
-import cliqz from './classes/Cliqz';
+import cliqz, { HUMANWEB_MODULE, HPN_MODULE } from './classes/Cliqz';
 // object class
 import Events from './classes/EventHandlers';
 import Policy from './classes/Policy';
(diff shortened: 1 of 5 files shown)


src/background.js
```diff
@@ -1340,11 +1341,8 @@ function getDataForGhosteryTab(callback) {
  */
 function initializePopup() {
 	if (BROWSER_INFO.os === 'android') {
-		chrome.browserAction.onClicked.addListener((tab) => {
-			chrome.tabs.create({
-				url: chrome.extension.getURL(`app/templates/panel_android.html?tabId=${tab.id}`),
```
christophertino, line 1345: @sammacbeth Check out https://github.com/ghostery/ghostery-extension/blob/498e44b0ae3109c8809353af6bf2506278c4b173/app/panel-android/components/Panel.jsx#L48.  The React app uses that `tabId` value to fetch data for the panel view. 

Using `setPopup` will require bumping our min FF Android version to 57+, which is fine.

sammacbeth: I was wondering why the panel seems to be working despite this, and it seems that there is code to handle the tabId parameter not being provided: https://github.com/ghostery/ghostery-extension/blob/498e44b0ae3109c8809353af6bf2506278c4b173/src/background.js#L705-L712

This code to get the active tab could be moved into the panel code to ensure it knows which tab it refers to in case an undefined tabId introduces other issues.

christophertino: Good catch.  @IAmThePan we can most likely remove that `tabId` lookup in Panel.jsx (see above)

IAmThePan: Noted.  However, including the `tabId` in the URL enables Anri to open the panel (android or regular) in a webpage rather than the popup so he can do regression tests.
I will incorporate these changes in my `feature/update-android` branch.  I will keep the code to open the `tabId` __without__ the query parameter, but I will make sure the query parameter _can be_ included if the use (tester, or developer) wants easy access to the panel.

IAmThePan: @christophertino ^^

PR #587, opened 2020-07-27
Updated Firefox Android extension

christophertino:
Improved mobile panel view for Firefox Android and Ghostery Android Browser.


src/background.js
```diff
@@ -1731,8 +1759,12 @@ function initializeGhosteryModules() {
 			// we need to do this after running scheduledTasks for the first time
 			// because of an A/B test that determines which promo variant is shown in the Hub on install
 			if (globals.JUST_INSTALLED) {
-				const route = (conf.hub_promo_variant === 'upgrade' || conf.hub_promo_variant === 'not_yet_set') ? '' : '#home';
-				const showPremiumPromoModal = conf.hub_promo_variant === 'midnight';
+				let route = (conf.hub_promo_variant === 'upgrade' || conf.hub_promo_variant === 'not_yet_set') ? '' : '#home';
+				let showPremiumPromoModal = conf.hub_promo_variant === 'midnight';
+				if (IS_ANDROID) {
```
christophertino, line 1764: @wlycdgr Do you want me to add these to the ternary / declaration statements above?

src/classes/Globals.js
```diff
@@ -153,6 +153,21 @@ class Globals {
 		const version = parseInt(ua.browser.version.toString(), 10); // convert to string for Chrome
 		const platform = ua.os.name.toLowerCase();
 
+		// Check for Ghostery Android Browser
+		if (typeof chrome.runtime.getBrowserInfo === 'function') {
```
christophertino, line 157: It may be better to make this it's own async function that we can call in specific cases. @wlycdgr thoughts?

src/classes/PanelData.js
```diff
@@ -71,7 +76,7 @@ class PanelData {
 		this._panelPort = port;
 		this._mountedComponents.panel = true;
 
-		getActiveTab((tab) => {
+		function tabCallback(tab) {
```
christophertino, line 79: @wlycdgr Double check this is ok

src/background.js
wlycdgr: I think the `route` ternary would get too convoluted, but `const showPremiumPromoModal = (conf.hub_promo_variant === 'midnight' && !IS_ANDROID)` looks clear enough.

src/classes/PanelData.js
wlycdgr: Looks good/same when `paramTabId` is undefined and we end up calling `getActiveTab` as before. How come we need to check for the presence of that param - is it not the same as the active tab on Android?

christophertino: this was to allow loading of the panel in a browser tab for automation testing

PR #594, opened 2020-08-11
GH-2040: Update Insights in-app promo CTA link

caseyoppenheim:
* [ ] Have you followed the guidelines in [CONTRIBUTING.md](https://github.com/ghostery/ghostery-extension/blob/master/CONTRIBUTING.md)?


app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -76,7 +76,7 @@ class PromoModal extends React.Component {
 				break;
 			case INSIGHTS:
 				sendMessage('ping', 'promo_modals_insights_upgrade_cta');
-				url = `https://checkout.${DOMAIN}.com/insights?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://www.ghostery.${DOMAIN}.com/insights/?utm_source=gbe&utm_campaign=in_app_upgrade=${utm_campaign}`;
```
fcjr, line 79: this will expand to `https:///www.ghostery.ghostery.com/insights/?utm_source=gbe&utm_campaign=in_app_upgrade=<utms>` take a look at the value of DOMAIN from line 26 of this file. as this URL does not change depending on environment we should probably not be using any conditional domains for this URL

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -76,7 +76,7 @@ class PromoModal extends React.Component {
 				break;
 			case INSIGHTS:
 				sendMessage('ping', 'promo_modals_insights_upgrade_cta');
-				url = `https://checkout.${DOMAIN}.com/insights?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://www.ghostery.com/insights/?utm_source=gbe&utm_campaign=in_app_upgrade=${utm_campaign}`;
```
fcjr, line 79: utm_campaign is a [query string](https://en.wikipedia.org/wiki/Query_string) so it is a string=string pair, so given that this function is always called with utm_campaign input as `in_app_upgrade` the current value here of utm_campaign would be `in_app_upgrade=in_app_upgrade`, i think what you want to do here is just set the value of the `utm_campaign` query param to the passed `utm_campaign` value

wlycdgr: ^ what Frank said

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -76,7 +76,7 @@ class PromoModal extends React.Component {
 				break;
 			case INSIGHTS:
 				sendMessage('ping', 'promo_modals_insights_upgrade_cta');
-				url = `https://checkout.${DOMAIN}.com/insights?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://www.ghostery.com/insights/?utm_source=gbe&utm_campaign=${utm_campaign}`;
```
fcjr, line 79: looks like we do actually have a staging website now, rather than doubling dev work lets wait till support for this is done by @benstrumeyer in #595 to update this PR.

fcjr:
Closing in favor of #595

PR #595, opened 2020-08-12
GH-2086, GH-2084, GH-2040: Update intro hub Get Plus, Midnight & Insights Modal CTA link

benstrumeyer:

- Intro Hub Get Plus should go to the plus description page
- Intro Hub's Ghostery Midnight modal should link to the landing page
- Intro Hub's Ghostery Insights modal should link to the landing page
 
Tickets: 
- https://ghostery.atlassian.net/browse/GH-2083
- https://ghostery.atlassian.net/browse/GH-2084


fcjr:
> We should use `www.${DOMAIN}.com` here as well to point at the staging website.
> 
> @wlycdgr This probably exists in lots of places.

While we are at it we should update the promo component to get this domain from `Globals` instead setting it at the top of this file.

wlycdgr:
> We should use `www.${DOMAIN}.com` here as well to point at the staging website.
> 
> @wlycdgr This probably exists in lots of places.

I'll check other places. I originally implemented it that way in this component so hopefully it will not be so many.

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -69,14 +69,14 @@ class PromoModal extends React.Component {
...
+				url = `https://www.${GHOSTERY_ROOT_DOMAIN}/products/plus?utm_source=gbe&utm_campaign=${utm_campaign}`;
 				break;
 			case PREMIUM:
-				url = `https://ghostery.com/thanks-for-downloading-midnight?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://www.${GHOSTERY_ROOT_DOMAIN}/midnight?utm_source=gbe&utm_campaign=${utm_campaign}`;
 				break;
 			case INSIGHTS:
 				sendMessage('ping', 'promo_modals_insights_upgrade_cta');
-				url = `https://checkout.${DOMAIN}.com/insights?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://checkout.${GHOSTERY_ROOT_DOMAIN}/insights?utm_source=gbe&utm_campaign=${utm_campaign}`;
```
fcjr, line 79: please use `CHECKOUT_BASE_URL` here

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -69,14 +69,14 @@ class PromoModal extends React.Component {
 		let url;
 		switch (product) {
 			case PLUS:
-				url = `https://checkout.${DOMAIN}.com/plus?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://www.${GHOSTERY_ROOT_DOMAIN}/products/plus?utm_source=gbe&utm_campaign=${utm_campaign}`;
 				break;
 			case PREMIUM:
-				url = `https://ghostery.com/thanks-for-downloading-midnight?utm_source=gbe&utm_campaign=${utm_campaign}`;
+				url = `https://www.${GHOSTERY_ROOT_DOMAIN}/midnight?utm_source=gbe&utm_campaign=${utm_campaign}`;
```
fcjr, line 75: it would also be nice to add a `GHOSTERY_WEBSITE_BASE_URL` global here to avoid managing the protocol and subdomain throughout the code.

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: Good catch! Changed to use `CHECKOUT_BASE_URL`

christophertino: in that case we should use `GHOSTERY_BASE_URL` for the others

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -23,7 +23,7 @@ import { sendMessage } from '../../panel/utils/msg';
 import globals from '../../../src/classes/Globals';
 import ModalExitButton from '../../panel/components/BuildingBlocks/ModalExitButton';
 
-const DOMAIN = globals.DEBUG ? 'ghosterystage' : 'ghostery';
+const { GHOSTERY_ROOT_DOMAIN } = globals;
```
fcjr, line 26: you need to expand `CHECKOUT_BASE_URL` here to use it below

wlycdgr: @fcjr @benstrumeyer we actually should also use `GHOSTERY_ROOT_DOMAIN` / `GHOSTERY_ROOT_URL` here, as per https://github.com/ghostery/ghostery-extension/pull/594

fcjr: @wlycdgr We might as well close #594 and just update this link in this PR as well then.

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: Great idea, added that variable and used it here

app/shared-components/PromoModal/PromoModal.jsx
```diff
@@ -23,7 +23,7 @@ import { sendMessage } from '../../panel/utils/msg';
 import globals from '../../../src/classes/Globals';
 import ModalExitButton from '../../panel/components/BuildingBlocks/ModalExitButton';
 
-const { GHOSTERY_ROOT_DOMAIN } = globals;
+const { GHOSTERY_WEBSITE_BASE_URL, CHECKOUT_BASE_URL } = globals;
```
wlycdgr, line 26: Let's use the `GHOSTERY_BASE_URL` value which we already have and which is the same as `GHOSTERY_WEBSITE_BASE_URL`

app/shared-components/PromoModal/PromoModal.jsx
wlycdgr: @fcjr good idea

benstrumeyer: I missed that! Added, and also updated the insights link, as per #594

app/shared-components/PromoModal/PromoModal.jsx
benstrumeyer: Good catch, I didn't realize we had this already

fcjr:
looks good to me, however one thing with using `GHOSTERY_BASE_URL` is that both these links will _always_ result in an immediate redirect `ghostery.com -> www.ghostery.com`. Not sure if it matters but wanted to point it out incase @wlycdgr or @christophertino have any concerns with that

wlycdgr:
> looks good to me, however one thing with using `GHOSTERY_BASE_URL` is that both these links will _always_ result in an immediate redirect `ghostery.com -> www.ghostery.com`. Not sure if it matters but wanted to point it out incase @wlycdgr or @christophertino have any concerns with that

In Chrome I don't see a user-visible redirect at all; in Firefox, there is one in the address bar, but there's no flicker, so I have no issue with it from a UX perspective anyway. I did notice a bug in Firefox though where the panel does not close after the Midnight landing page opens! I'll make a ticket.

christophertino:
It shouldn't matter since the request is [redirected](https://github.com/ghostery/ghostery-website/blob/master/wp-content/themes/ghostery/deploy/ansible/templates/nginx.conf.j2#L11) in nginx.  If we wanted to be really verbose we could make another global variant as a `www` url, but I don't think it's necessary.

PR #601, opened 2020-09-02
Fix A/B test timing bug

wlycdgr:
`conf.save.enable_human-web` is set during startup, which is intercepted by the dispatcher, which calls `setupABTest`, which calls `setupHubPromoABTest`. Because the call to `setupABTest` in the `dispatcher.on` handler is not then()-chained to an `abtest.fetch` call, `setupHubPromoABTest` may get called before the A/B tests have been fetched for the first time. The absence of either `hub_midnight` or `hub_plain` on the still-empty `abtest.tests` object would then have been incorrectly interpreted to mean that the user should be placed in the remaining test bucket.

I initially fixed this by putting the `setupABTest` call in the `dispatcher.on` handler in a then() clause of an `abtest.fetch` call, but I prefer the current fix because it avoids making an extra call to the A/B test server on every startup.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -1118,7 +1118,12 @@ function getAntitrackingTestConfig() {
  * @return {Object} 	Hub promotion configuration parameters
  */
 function setupHubPromoABTest() {
-	if (conf.hub_promo_variant !== 'not_yet_set') return;
+	if (
+		!abtest.hasBeenFetched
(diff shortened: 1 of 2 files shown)


fcjr:
Looks to be working as expected

fcjr (approved):

PR #604, opened 2020-09-17
GH-2171: Add Missing Page Title to Upgrade Plan Hub Page

fcjr:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,23 +3,23 @@
 # the last matching pattern has the most precedence.
 
 # Core Ghostery team
-*       @ghostery/ghostery
+*       @ghostery/extension
 
 # CMP / Metrics / WebRequest
-/src/classes/ABTest.js @jsignanini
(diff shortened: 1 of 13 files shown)


benstrumeyer (approved):
LGTM

wlycdgr:
Nice spot

fcjr:
> Nice spot. If we want to avoid setting this on every render without reverting to using a class component, we can use the `useEffect` hook:
> 
> ```
> useEffect(() => {
>   document.title = 'Ghostery is cool';
> }, []);
> ```
> 
> The empty array second argument makes it so that the effect is only run once instead of after every render: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
> 
> (No new packages need to be installed to use hooks, we should just be able to do `import React, { useEffect } from 'react'` since we are using React 16.13 and they've been supported since 16.8)

Nice! Looks to be working for me, updated to use this.

PR #622, opened 2020-10-22
GH-2188 - Adjustments for Ghostery Browser

christophertino:
- Disable intro hub
- Disable Rewards
- Enable metrics
 See GH-2188


app/panel/components/Help.jsx
```diff
@@ -23,9 +26,11 @@ const Help = () => (
 		<div className="row">
 			<div className="small-12 columns">
 				<h1>{t('panel_help_panel_header')}</h1>
-				<div className="support-section">
-					<a href="#" onClick={openHubPage}>{t('panel_help_setup')}</a>
-				</div>
+				{ BROWSER_INFO.name !== 'ghostery_desktop' && (
```
sammacbeth, line 29: So we completely disable access to the hub page on the browser? Would it not be useful for some users who may want to access it later?

src/background.js
```diff
@@ -1642,9 +1648,9 @@ function initializeGhosteryModules() {
 				conf.enable_ad_block = !adblocker.isDisabled;
 				conf.enable_anti_tracking = !antitracking.isDisabled;
 				conf.enable_human_web = !humanweb.isDisabled;
-				conf.enable_offers = !offers.isDisabled && !IS_ANDROID;
+				conf.enable_offers = !offers.isDisabled && !IS_ANDROID && BROWSER_INFO.name !== 'ghostery_desktop';
```
sammacbeth, line 1651: Is there a potential race here? How early is this block evaluated in the startup? The browser name will not be correct until after `chrome.runtime.getBrowserInfo` has resolved.

src/classes/ConfData.js
```diff
@@ -111,8 +111,8 @@ class ConfData {
 			_initProperty('enable_click2play', true);
 			_initProperty('enable_click2play_social', true);
 			_initProperty('enable_human_web', !IS_CLIQZ && !IS_FIREFOX);
-			_initProperty('enable_metrics', false);
-			_initProperty('enable_offers', !IS_CLIQZ && !IS_FIREFOX && !IS_ANDROID);
+			_initProperty('enable_metrics', BROWSER_INFO.name === 'ghostery_desktop');
```
sammacbeth, line 114: Again, could this be a race?

app/panel/components/Help.jsx
christophertino: We would need to adjust opt-in default settings for the hub but it's doable.

ang557 (commented):
I am being hacked this has been going on for 4 months now there hacking me off your website with coding

src/background.js
christophertino: This evaluates pretty far down the init chain and is nested inside `cliqz.start()`. The line directly below uses a similar check for android and has been working ok. But I admit I don't love this implementation. Open to suggestions.

src/classes/ConfData.js
christophertino: Same as above, it does evaluate in order but the possibility for a race condition technically exists.

src/background.js
sammacbeth: We could attach a `Globals.browserInfoReady` Promise in `Globals` when we call `buildBrowserInfo`, which we can then just `await` at these points that need to be sure that the browser check has been done.

app/panel/components/Help.jsx
christophertino: Fixed

manifest.json
```diff
@@ -3,10 +3,11 @@
 	"author": "Ghostery",
 	"name": "__MSG_name__",
 	"short_name": "Ghostery",
-	"version": "8.5.3",
-	"version_name": "8.5.3",
+	"version": "8.5.4",
+	"version_name": "8.5.4",
 	"default_locale": "en",
 	"description": "__MSG_short_description__",
+	"debug": true,
```
sammacbeth, line 10: Should this be here?

christophertino: yea it's fine for `develop`. It gets removed by the builder.

sammacbeth (approved):

PR #632, opened 2020-11-18
GH-2222 - Handle updated Chrome new tab page url

christophertino:
The Chrome new tab url is now chrome://new-tab-page. Prior to Chrome 75 the url was https://www.google.com/_/chrome/newtab.  For Purplebox and Click2Play, we only inject on http/s pages.  We still check for `_/chrome/newtab` as a fallback for legacy users.


wlycdgr (approved):
LGTM

PR #633, opened 2020-11-19
Desktop browser telemetry

sammacbeth:
Adds extra telemetry for the desktop browser. These are fetched over inter-extension messaging from the ghostery search extension (see https://github.com/ghostery/ghostery-search-extension/pull/30) and then appended to metrics requests.


wlycdgr (approved):
Looks good to me

christophertino (approved):

PR #647, opened 2020-12-22
GIS-866: Search Communication

fcjr:
Exposes the refreshToken for use in the ghostery search extension


src/classes/SearchMessager.js
```diff
@@ -0,0 +1,51 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import ExtMessenger from './ExtMessenger';
+import account from './Account';
+import { log } from '../utils/common';
+
+/**
+ * Class for handling cross-extension messaging.
```
wlycdgr, line 19: Can you throw a @since into the JSDoc comment?

wlycdgr: ....What the heck is going on with the @ since formatting, ?lol?

wlycdgr (commented):
LGTM, just the one JSDoc thing would be great

wlycdgr (approved):
LGTM

src/classes/SearchMessager.js
christophertino: You tagged that github user. I updated the comment.

src/background.js
```diff
@@ -66,6 +67,7 @@ const {
 	CDN_BASE_URL, BROWSER_INFO, IS_CLIQZ
 } = globals;
 const IS_EDGE = (BROWSER_INFO.name === 'edge');
+const IS_GHOSTERY_BROWSER = (BROWSER_INFO.name === 'ghostery_desktop');
```
christophertino, line 70: This is asynchronous. Check out `globals.BROWSER_INFO_READY` instead.

christophertino (changes requested):

src/background.js
fcjr: Ah thanks, updated to await BROWSER_INFO_READY, however I see the browser info is being used before verifying that it is populated in a few other places which we should address in another PR -- we should also probably have BROWSER_INFO_READY short circuit if once a successful call to buildBrowserInfo is completed as currently every time we await `globals.BROWSER_INFO_READY` we are unnecessarily recalculating these values.

christophertino: Let's remove this line completely since it's evaluated at runtime and unreliable. The other browser_info checks should be working as is. The async check only applies to the desktop browser and Android browser to allow us to differentiate from stock FF.

src/background.js
```diff
@@ -1648,6 +1650,18 @@ function initializeGhosteryModules() {
 	});
 }
 
+/**
+ * Initialize Search Message Handler on Ghostery Browser.
+ * @memberOf Background
+ */
+async function initializeSearchMessageHandler() {
+	await globals.BROWSER_INFO_READY; // ensure browser info is set
+	if (IS_GHOSTERY_BROWSER) {
```
christophertino, line 1659: change to `BROWSER_INFO.name === 'ghostery_desktop'`

christophertino (changes requested):

src/utils/api.js
```diff
@@ -27,6 +27,10 @@ class Api {
 		}
 	}
 
+	refreshToken() {
```
jsignanini, line 30: @fcjr I'm thinking renaming this to something like `refreshTokenAndHandleIsRefreshing` to make it super clear what the purpose of this new function is

jsignanini (changes requested):
@fcjr see comment

src/utils/api.js
fcjr: That sounds reasonable, I'll make the change. My thought with just naming it `refreshToken` was that as an external user of the api class the inner "isRefreshing" state is an implementation detail that external users of the function don't really need to know about, but since this should only be used in special cases anyways there shouldn't be a harm in making it explicit

jsignanini: @fcjr actually you do make a good point as it’s only used internally in the other “private” function. Ok you can just leave it as is then. Thanks.

jsignanini (approved):

christophertino (approved):

src/utils/api.js
```diff
@@ -27,6 +27,10 @@ class Api {
...
+		return this._refreshToken().then(this.isRefreshing = false); // eslint-disable-line no-return-assign
```
chrmod, line 31: `this.isRefreshing` is set to false immediately  after `this._refreshToken` is being called - it does not wait for a promise to resolve.

chrmod, line 31:
```suggestion
		return this._refreshToken().then(() => { this.isRefreshing = false; });
```

chrmod (changes requested):

src/utils/api.js
chrmod: Also, why `this.isRefreshing` is not set to `false` by `_refreshToken()` ?

chrmod (changes requested):

src/utils/api.js
```diff
@@ -27,23 +27,41 @@ class Api {
...
+			window.dispatchEvent(new CustomEvent(this.tokenRefreshedEventType, {
+				detail: res,
+			}));
+			return res;
+		}).catch((err) => {
+			this.isRefreshing = false;
+			window.dispatchEvent(new CustomEvent(this.tokenRefreshedEventType, {
+				detail: { err },
+			}));
+			throw err;
```
jsignanini, line 64: @fcjr could you move the two `this.isRefreshing = false` to just one in a `.then()` after the `.catch()`?

jsignanini (approved):

chrmod (approved):

Issue #44, opened 2018-04-22
Ghostery breaks onBeforeRequest redirects made by other extensions


lidel:
### Description

[IPFS Companion](https://github.com/ipfs/ipfs-companion) is a browser extension that enables users to detect content-addressed resources on websites and redirect them to a local IPFS2HTTP gateway (localhost or LAN). 

It seems that when Ghostery is enabled and user loads a HTTPS website with content-addressed IPFS resources, Ghostery breaks the gateway redirect by forcing HTTPS even if user chose to use  a localhost or LAN gateway exposed on HTTP. 

More details: https://github.com/ipfs-shipyard/ipfs-companion/issues/466

### Expected Behavior 

Ghostery should detect and respect internal redirects (HTTP 307) made by other extensions. 

Note that [HTTPS Everywhere](https://addons.mozilla.org/en-US/firefox/addon/https-everywhere/) does not break the same website.

### Actual Behavior 

Ghostery always upgrades URL of HTTP 307 to HTTPS, breaking IPFS websites:
> ![2018-04-22-021752_877x973_scrot](https://user-images.githubusercontent.com/157609/39090006-6a0ef7dc-45d3-11e8-9527-813218f1fb4e.png)

### Steps to Reproduce

1.  Install [ipfs-companion](https://github.com/ipfs/ipfs-companion/#install) and Ghostery
2. Open Preferences of ipfs-companion
3. Set up a working "Custom Gateway" to a HTTP port and IP of a [go-ipfs](https://github.com/ipfs/go-ipfs) node running on your localhost or in  LAN.
4. Open HTTPS website with content-addressed resources, eg. https://d.tube/#!/v/goingforhealthy/go0fyd9m
5. Open Console, note network errors caused by URLs starting with `https://` instead of redirected `http://`  ([like in this screenshot](https://user-images.githubusercontent.com/157609/39090006-6a0ef7dc-45d3-11e8-9527-813218f1fb4e.png))

### Versions

+ IPFS Companion: 2.2.1
+ Browserify:  8.1.2
+ Browser: Chrome and Firefox

lidel:
Yup, `localhost` and `127.0.0.1` are safe to whitelist (browsers don't even show a mixed-content warning for `127.0.0.1` eg. [Firefox since v55](https://bugzilla.mozilla.org/show_bug.cgi?id=903966)).

PR #675, opened 2021-02-02
Onboarding spreadsheet bugs #38 (GH-2259), #44 (GH-2306), and #45 (GH-2305)

benstrumeyer:

- Bug 38, Make Glow logo larger in search selection modal
- Bug 44, Center align ghostery glow icon with the box
- Bug 45, Select plus plan and expand selection if user is basic and has chosen ghostery search


wlycdgr (commented):
Changes look good ~ Couple more things:

* I made a render helper for the checkmark items in the option cards and started refactoring the code to use it - if you could finish that up that would be great
* Let's factor the `isBasic`, `isPlus`, and `isPremium` checks out to helpers. First cos they're duplicated in `render` and `setDefaultPlan`, but also we reduce chance of error nicely with the logic for these checks being a bit convoluted.
* Speaking of `setDefaultPlan` - as far as I can tell from a combination of memory and double-checking just now, `setTimeout` is non-blocking, so with the code as it was, `render` would run before `setDefaultPlan` anyway (you can see this in action by setting `readyToRender` to `true` from the get-go and throwing in a couple consologs). To make sure we actually don't render until after `setDefaultPlan` has run after the timeout period, I added this `readyToRender` state variable that defaults to `false` and short-circuits `render` until it's set to `true at the end of `setDefaultPlan`. Nothing to do here at the moment really, just wanted to point this out (ultimately though it'd be good to update the `AccountReducer` so that we have a way of distinguishing between 'no user' and 'user data not fetched yet').
* I took the liberty of making some additional adjustments to search selection modal logo sizes to better match the Zeplins, since that was on me to begin with.

wlycdgr:
> * I made a render helper for the checkmark items in the option cards and started refactoring the code to use it - if you could finish that up that would be great

Done

wlycdgr:
> * Let's factor the `isBasic`, `isPlus`, and `isPremium` checks out to helpers. First cos they're duplicated in `render` and `setDefaultPlan`, but also we reduce chance of error nicely with the logic for these checks being a bit convoluted.

Done

wlycdgr:
> * Speaking of `setDefaultPlan` - as far as I can tell from a combination of memory and double-checking just now, `setTimeout` is non-blocking, so with the code as it was, `render` would run before `setDefaultPlan` anyway (you can see this in action by setting `readyToRender` to `true` from the get-go and throwing in a couple consologs). To make sure we actually don't render until after `setDefaultPlan` has run after the timeout period, I added this `readyToRender` state variable that defaults to `false` and short-circuits `render` until it's set to `true` at the end of `setDefaultPlan`.  Nothing to do here at the moment really, just wanted to point this out (ultimately though it'd be good to update the `AccountReducer` so that we have a way of distinguishing between 'no user' and 'user data not fetched yet').

Reworked to call `componentDidMount` to avoid the hack of using `setTimeout` to `setState` "in" the `constructor`

wlycdgr:
> * I took the liberty of making some additional adjustments to search selection modal logo sizes to better match the Zeplins, since that was on me to begin with. Let me know if you see that I screwed anything up and I'll fix.

Double checked that this looks ok, but a fresh pair of eyes would be appreciated

leuryr (approved):
LGTM

christophertino (approved):

PR #678, opened 2021-02-04
GH-2213 bugfix: Broken Next button

wlycdgr:
Fixes bug where Next button would break if the internal name of the user's chosen search returned by `browser.search.get()` did not match any of our search engine name constants.

I added some protection to reduce the risk of unexpected values breaking our code, and added a new `dawn_setup_number` to use for cases when we don't recognize the chosen search.

Finally, I made a best guess for what the Search Encrypt internal name will turn out to be to improve the chance that we won't have to make a code change for this. If we start getting a lot of "unrecognized" pings once that engine is added on the Dawn side, we'll want to look here first.

* [X] Have you followed the guidelines in [CONTRIBUTING.md](https://github.com/ghostery/ghostery-extension/blob/master/CONTRIBUTING.md)?


benstrumeyer (approved):
LGTM, approved

PR #680, opened 2021-02-06
Allow host name fuzzy matching for Compatibility URLs

christophertino:
Updates `fuzzyUrlMatcher()` to allow host name fuzzy matches from the Compatibility DB. A compatibility entry for `*atlassian.net` should allow `ghostery.atlassian.net`.  This will also allow host and path fuzzy matches, i.e. `*atlassian.net/page*`.


jsignanini:
@christophertino my only suggestion here is that we make this more strict so that fuzzy matching only works after a `.`, because on the example above, an entry of `*atlassian.net` would match with `www.anythingatlassian.net` which would not be the desired outcome. So it should only work as `*.atlassian.net`.

zarembsky (changes requested):
Agree with @jsignanini. I'll update Midnight PR accordingly.

fcjr:
@jsignanini if we keep the syntax as is and instead just add the rules as `*.atlassian.net` is there much harm? Then if for some reason we wanted to add a rule based on domain partials in the future we have the flexibility.

jsignanini:
@afranco07 I'd rather be safe than sorry, patterns could be incorrectly introduced quite easily, and I don't really see a valid use-case for a catch-all that doesn't get restricted to sub-domains.

zarembsky (approved):
Looks good. Can be merged once Travis CI is over.

PR #681, opened 2021-02-08
GH-2258: Preserve user onboarding settings

leuryr:
Keeps user settings for Blocking and Default Search when the user returns to those screens with Back.

- Adds `blockSetupSeen` and `searchSetupSeen` to Redux in order to track if the user has previously visited the settings page and the old selections should be displayed, in addition to respective actions to update them.
- Changes some of the Redux keys in order to prevent any issues when pulling in state through props that have the same key names

---


app/dawn-hub/Views/OnboardingViews/Step3_ChooseDefaultSearchView/ChooseDefaultSearchView.jsx
```diff
@@ -78,6 +78,24 @@ class ChooseDefaultSearchView extends Component {
...
+			const searchChoices = [
+				SEARCH_GHOSTERY,
+				SEARCH_BING,
+				SEARCH_YAHOO,
+				SEARCH_STARTPAGE,
+			];
+			const isOther = searchChoices.reduce((accum, choice) => {
+				if (accum === false) return false;
+				return choice !== defaultSearch;
+			}, false);
```
wlycdgr, line 95: You can simplify to
`const isOther = searchChoices.includes(defaultSearch)`

(Also I think your version will always return `false`? Since `accum` is initialized to `false` and so the condition in line 93 is always satisfied. Bigger picture, with the simpler `includes` version it's much easier to be confident that the code does what you intend)

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
```diff
@@ -30,13 +30,70 @@ class BlockSettingsView extends Component {
...
+					decodedPolicy = 2;
+					break;
+				case 'BLOCKING_POLICY_NOTHING':
+					decodedPolicy = 3;
+					break;
+				default:
+					break;
+			}
+		}
+		return decodedPolicy;
```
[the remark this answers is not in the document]
wlycdgr: Also, while we're digging around in this file, could you factor out rendering of the `BlockSettingsView_question` `li`s to a helper like the `renderAnswerBlock` helper I made? It would be a nice bit of cleanup since they are all the same except for a couple strings

app/dawn-hub/Views/OnboardingViews/Step3_ChooseDefaultSearchView/ChooseDefaultSearchView.jsx
```diff
@@ -78,6 +78,24 @@ class ChooseDefaultSearchView extends Component {
...
+			const prevChosenSearch = isOther ? SEARCH_OTHER : defaultSearch;
+			this.setState({ chosenSearch: prevChosenSearch, otherSearchSelected: isOther ? prevChosenSearch : null });
+		}
```
wlycdgr, line 98: This will set `otherSearchSelected` to `SEARCH_OTHER` if `prevChosenSearch` is `SEARCH_OTHER`. That's not quite right: `otherSearchSelected` should store which _specific_ other search is selected, when `chosenSearch` is `SEARCH_OTHER` (e.g., `SEARCH_GIBIRU`, etc)

app/shared-hub/reducers/AntiSuiteReducer.js
```diff
@@ -17,21 +17,27 @@ import {
...
 } from '../constants/AntiSuiteConstants';
 
-const initialState = {};
+const initialState = {
+	antiSuite: {
+		enable_ad_block: false,
+		enable_anti_tracking: false,
+		enable_smart_block: false
+	}
+};
```
wlycdgr, line 26: I am realizing I have a strong suspicion we don't need the extra `antiSuite` and `setupLifecycle` wrappers at all after all, and that the code would be better and simpler without them....but I can fix that up at the end, don't worry about it unless you want to

wlycdgr (commented):
Lemme know if you have any questions on the comments! Two other quick things for the `BlockSettingsView`:

* "Recommended Choices" checkbox should be checked if the user's choices match what is recommended
* "Recommended Choices" checkbox should become unchecked if the user changes their choice such that they no longer exactly match the recommended settings (this one isn't specific to this PR but would just be a nice bit of cleanup while we're at it)

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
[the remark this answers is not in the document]
leuryr: > Better yet, can we make do without doing these conversions in the first place? I think we are asking for trouble if `kindsOfTrackers` can sometimes be a string and sometimes a number. What if we just always use the string values? That's what `background.js` expects so no conversion would be necessary - only a modification to what values we are testing against in the checkbox rendering code (unless I'm missing somethin'?)

While it looks like we still need the number representation for when we send over the `setup_number` string from onboarding, using the constants everywhere else would mean that we only need to convert to a number for that case in particular, so all the other code can be removed/refactored. Was able to drop quite a few lines with this change.

app/dawn-hub/Views/OnboardingViews/Step3_ChooseDefaultSearchView/ChooseDefaultSearchView.jsx
leuryr: Yea, you're right. Definitely messed up on the logic here, and `.includes()` is a much simpler implementation. I changed the check to `const isOther = !searchChoices.includes(defaultSearch)` as `.includes(defaultSearch)` returning `true` would mean that `isOther` should be `false`.

app/dawn-hub/Views/OnboardingViews/Step3_ChooseDefaultSearchView/ChooseDefaultSearchView.jsx
leuryr: Yep messed up here as well. This should be setting `otherSearchSelected` as the `defaultSearch` value if `isOther` is true, which would carry the specific other search that the user had selected.

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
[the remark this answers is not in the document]
wlycdgr: Sweet!

app/dawn-hub/Views/OnboardingViews/Step3_ChooseDefaultSearchView/ChooseDefaultSearchView.jsx
wlycdgr: Great, thanks man!

app/dawn-hub/Views/OnboardingViews/Step3_ChooseDefaultSearchView/ChooseDefaultSearchView.jsx
wlycdgr: Woot woot, ty ty

app/shared-hub/reducers/AntiSuiteReducer.js
wlycdgr: (Based on our convo, you're right, let's leave it as is)

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
```diff
@@ -39,57 +44,56 @@ class BlockSettingsView extends Component {
...
+				decodedPolicy = 1;
+				break;
+			case BLOCKING_POLICY_RECOMMENDED:
+				decodedPolicy = 2;
+				break;
+			case BLOCKING_POLICY_NOTHING:
+				decodedPolicy = 3;
+				break;
+			default:
+				break;
```
wlycdgr: If I'm not mistaken, we can dispense with the need for even this simplified converter function by setting the values of the `BLOCKING_POLICY_*` to be these numbers right in `share-hub\constants\BlockingPolicyConstants`, no?

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
```diff
@@ -39,57 +44,56 @@ class BlockSettingsView extends Component {
...
+			enable_ad_block,
+			enable_smart_block,
+			kindsOfTrackers
+		} = this.state;
+
+		if (!recommendedChoices && kindsOfTrackers === BLOCKING_POLICY_RECOMMENDED && enable_ad_block === true && enable_anti_tracking === true && enable_smart_block === true) {
+			// eslint-disable-next-line react/no-did-update-set-state
+			this.setState({ recommendedChoices: true });
+		} else if (recommendedChoices && (kindsOfTrackers !== BLOCKING_POLICY_RECOMMENDED || enable_ad_block !== true || enable_anti_tracking !== true || enable_smart_block !== true)) {
+			// eslint-disable-next-line react/no-did-update-set-state
```
wlycdgr, line 78: It would be good if we can avoid needing to disable this linting rule. Updating state in `componentDidUpdate` will trigger another render right after one has been completed, which according to the linter docs justifying the rule may lead to layout thrashing.

`recommendedChoices` can be derived from the other bits of state anyway, so it would be cleaner to dispense with it. when we only store state values that can't be derived from other state values, and derive others from these as needed, we eliminate the risk of a derived piece of state being incorrectly out of sync due to forgetting to update it or a bug in the update logic.

my suggestion:
* get rid of `recommendedChoices`
* replace it with a `_recommendedChoices()` method that tells you whether the current combo of selections matches the recommended choices and use that method to determine whether the RC checkbox should be checked in `render()`
* get rid of `componentDidUpdate()` - with this setup, you don't need it anymore. whenever a user changes a setting, the corresponding bits of state gets updated, `render()` runs, and `_recommendedChoices()` spits out the appropriate checkbox value

wlycdgr (commented):
Just a couple more changes!

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
leuryr: Thank you for this comment, definitely better this way! I've updated this to using function `recommendedChoicesActive()` for determining whether the checkbox should be checked.

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
leuryr: Since it looks like `background.js` is expecting a string, and another part of the app in the old hub is also sending it this way, we might have to make more changes than initially intended to implement this change. I think it might be best to leave it as is for now, and revisit it later in a wider refactor? I agree that it's worth looking into and keeping in mind.

wlycdgr: Sounds good

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
wlycdgr: Great, thanks!

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
```diff
@@ -230,7 +225,7 @@ class BlockSettingsView extends Component {
 						<div className="BlockSettingsView_checkboxBlock">
 							<ToggleCheckbox
 								className="BlockSettingsView_checkbox"
-								checked={recommendedChoices}
+								checked={this.recommendedChoicesActive()}
 								onChange={() => this.toggleRecommendedChoices(!recommendedChoices)}
 							/>
 							<div className="BlockSettingsView_checkboxLabel" onClick={() => this.toggleRecommendedChoices(!recommendedChoices)}>{t('ghostery_dawn_onboarding_recommended_choices')}</div>
```
wlycdgr, line 231: (Sorry if this wasn't initially clear) - let's follow through on this idea throughout the component and get rid of `recommendedChoices` completely. We can call `this.recommendedChoicesActive()` one time at the beginning of `render` and save the result to avoid repeating the calculation three times

wlycdgr (changes requested):

app/dawn-hub/Views/OnboardingViews/Step2_BlockSettingsView/BlockSettingsView.jsx
leuryr: Ah ok, yea that makes sense!

wlycdgr (approved):
LGTM!

christophertino (approved):

PR #694, opened 2021-02-26
GH-2289: Add sub status to CMP fetch URL

leuryr:
Adds the `ss` param to the CMP fetch URL. Feeling a bit conflicted about having the logic for this inside the `_buildUrl` method, though it is technically part of building the URL. Open to moving it into a separate function.

---


wlycdgr:
Looks like it works as expected, but I share your mixed feelings about having the logic be inside the `_buildURL` method. I suggest factoring it out to a static method of either the `CMP` or the `Metrics` class. While you're at it, could you convert `CMP#_buildURL` to use the `_buildQueryPair` method you made in `Metrics`? It can be (and should be, since it can be) converted to a static method, I think. 

Probably most importantly, we should see if we can rename the key from `ss` to something else, to avoid the confusion of a key that means one thing in the context of a call to the CMP server and something else entirely in the context of a call to the metrics server

wlycdgr (commented):
See comment (but disregard the part about the parameter naming)

leuryr:
As we discussed, I moved the query string building to a util function that we can use for both the `CMP` & `Metrics` classes (with a small addition for when we need a `?` at the beginning of the returned string instead of `&`), and I factored getting the sub status out from `CMP_buildURL` into its own static method.

christophertino (approved):

PR #715, opened 2021-07-12
ASDF to ensure nodejs version

chrmod:


christophertino:
Should we also bump `.nvmrc`?

chrmod:
oh, I got it now. 

I have not notice `.nvmrc` as was looking for `.node-version` - `ASDF` support both formats so there is no need for `.tool-versions` file.

Closing the PR

PR #721, opened 2021-09-08
Remove Ghostery Dawn onboarding

chrmod:
Dawn onboarding got simplified and is moved to own addon https://github.com/ghostery/ghostery-dawn-onboarding-extension


christophertino:
For future reference when editing the language files, you can just remove the tokens from the `en` file and push to Transifex. They'll remove them from the other languages and you can pull them down.  We rely on Transifex to be the source of truth for the translations.

src/classes/Metrics.js
```diff
@@ -270,8 +270,6 @@ class Metrics {
 			buildQueryPair('at', conf.enable_anti_tracking ? '1' : '0') +
 			// The deepest setup page reached by user during setup
 			buildQueryPair('ss', Metrics._getSetupStep(type).toString()) +
-			// The number of times the user has gone through setup in the regular hub, or the answer selection on each page of the dawn-hub
-			buildQueryPair('sl', Metrics._getSetupNumber(type).toString()) +
```
christophertino, line 274: I think we should leave this in but change back to 
`buildQueryPair('sl', conf.setup_number.toString()) +`

christophertino (changes requested):
run `yarn test.snapshot` to fix those lint errors

src/classes/Metrics.js
chrmod: I wasn't sure about that one, was going to check AWS Glue if I would not break the schema. Thanks for help Chris!

PR #727, opened 2021-11-18
Fix site breakage related to the Tracker-Tally CSS

philipp-classen:
The Tracker-Tally can break sites that run JavaScript that interacts with the injected css code. A recent example is
"https://mycovidrecord.health.nz/", which fails with:

```
Uncaught DOMException: CSSStyleSheet.cssRules getter: Not allowed to access cross-origin stylesheet
```

These older examples could be duplicates as well:
* https://github.com/niklasvh/html2canvas/issues/2197
* https://stackoverflow.com/a/62475347/783510

We can delay the css injection to "document_end" instead of "document_start". It may still happen, but it should make it
less likely with little downsides. In the concrete example, I can no longer reproduce with the change.

* [ ] Have you followed the guidelines in [CONTRIBUTING.md](https://github.com/ghostery/ghostery-extension/blob/master/CONTRIBUTING.md)?

### DIFF
--- a/src/classes/ConfData.js
+++ b/src/classes/ConfData.js
@@ -135,7 +135,7 @@ class ConfData {
 			_initProperty('paid_subscription', false);
 			_initProperty('settings_last_imported', 0);
 			_initProperty('settings_last_exported', 0);
-			_initProperty('show_alert', !IS_ANDROID);
+			_initProperty('show_alert', false); // Tracker-Tally
 			_initProperty('show_badge', true);
 			_initProperty('show_cmp', true);
 			_initProperty('show_tracker_urls', true);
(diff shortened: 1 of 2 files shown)


philipp-classen:
Also, disabling the feature for new installations (users can enable it through the UI: settings -> Tracker Tally).

chrmod (approved):

varun-raj:
This is still happening for our webapp. We use rails + webpacker.
Is there any ways to prevent the site break?

philipp-classen:
@varun-raj We will push a release soon (most likely tomorrow). Apology for the delay, but once it is live, it should fix the problems.

PR #730, opened 2021-12-02
Smart Blocking: Allow blocking of first party tracking

chrmod:
In order to deal with cname cloaking we have to be able to block first party trackers. That way we can block Google Beacons.

### DIFF
--- a/src/classes/EventHandlers.js
+++ b/src/classes/EventHandlers.js
@@ -326,7 +326,6 @@ class EventHandlers {
 
 		// TODO fuse this into a single call to improve performance
 		const page_url = tabInfo.getTabInfo(tab_id, 'url');
-		const page_domain = tabInfo.getTabInfo(tab_id, 'domain');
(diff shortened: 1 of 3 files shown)


src/classes/EventHandlers.js
```diff
@@ -326,7 +326,6 @@ class EventHandlers {
 
 		// TODO fuse this into a single call to improve performance
```
philipp-classen, line 327: I think, after the removal of page_domain that TODO is also obsolete (only one call to tabInfo.getTabInfo left).

philipp-classen (approved):
Merging for now, as the change makes sense. It has an unintended side-effect of disabling the later steps of the community-based adblocking (the removed code path did not set `eventMutable.ghosteryWhitelisted = true`).

As the former behavior is more accidental, merging and finding a proper fix for the latter (it is an independent issue) makes sense IMHO.

PR #764, opened 2022-04-08
Fix Chrome memory leak

### LINKED ISSUE #481: High memory usage
### Description
Ghostery is using a lot of RAM. 

### Expected Behavior 
Ghostery should have a reasonable usage. I can accept 100 MB maximum 150 MB.

### Actual Behavior 
In my case it is using about 450 MB and the best value was 326 MB.  

### Steps to Reproduce
1. Chrome latest version on Windows 10 Pro. No other extensions except Ghostery
2. No URL in Address bar just Chrome home page.
3. Shift + Esc to check the memory usage. CPU usage is around 45% (4 cores).

I will disable Ghostery for a while. My machine is only 4 GB of RAM and more than 10% is too much. Usage of CPU is also an issue.

chrmod:
fix #481 

This PR replaces native `DOMParser` with a library `linkedom`. This have to be done to mitigate Chromium memory leak.

Here screenshots from debugging tools for two extensions running in same browser
right: `linkedom`, left: native DOMParser
Screenshot 2022-04-08 at 17 21 30

Chrome Task manager after some time:
Screenshot 2022-04-08 at 17 34 20

We see that `linkedom` has higher memory footprint, but it does not leak.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "hybrids": "^7.0.0",
     "jquery": "3.5.0",
     "json-api-normalizer": "^1.0.4",
+    "linkedom": "^0.14.5",
     "moment": "^2.29.1",
     "prop-types": "^15.6.2",
     "query-string": "^6.13.1",
--- a/src/classes/Cliqz.js
+++ b/src/classes/Cliqz.js
@@ -1,5 +1,5 @@
 /**
- * Cliqz Import Class
+ * Ghostery Common Import Class
  *
  * Ghostery Browser Extension
  * https://www.ghostery.com/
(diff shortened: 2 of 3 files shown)


chrmod:
@philipp-classen please test `linkedom` replacing `jsdom` in `ghostery-common`. We can only make this change here when all the tests will be passing

philipp-classen:
> @philipp-classen please test `linkedom` replacing `jsdom` in `ghostery-common`. We can only make this change here when all the tests will be passing

This one is already open: https://github.com/ghostery/common/pull/45

smalluban (approved):
LGTM.

philipp-classen (approved):

PR #765, opened 2022-04-11
Update adblocker and package clean up

smalluban:
We decided to split the version bump from the update. I will leave the branch name for simplicity, so don't mind it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,12 @@
 {
   "name": "ghostery-extension",
-  "version": "0.0.0",
-  "title": "Ghostery",
+  "private": true,
   "author": {
-    "name": "Ghostery, Inc",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "ghostery-extension",
-  "version": "0.0.0",
-  "title": "Ghostery",
+  "private": true,
+  "version": "8.6.3",
```
chrmod, line 4: lets not change the version here, no need to bump it as we never publish

chrmod (changes requested):

philipp-classen (approved):

chrmod (approved):

PR #769, opened 2022-04-19
Link account registration to website

chrmod:
Instead of account registration form directly in the addon, we now link to the website to ensure all users have exactly same registration flow


smalluban (approved):
Fixed regquested missing link, thanks. LGTM.

PR #772, opened 2022-04-19
version bump

chrmod:


smalluban (approved):
For the notice, I read in MDN docs, that if `version_name` is the same as `version` it can be omitted.

chrmod:
> For the notice, I read in MDN docs, that if `version_name` is the same as `version` it can be omitted.

Good to know. Lets remove it next time we touch it.

PR #808, opened 2022-07-06
Bugfix: cannot login to ghostery account in Firefox with same site isolation

chrmod:
Firefox privacy features like [Same Site Isolation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies#first-party_isolation) and [State partitioning](https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning) change the way we are able to use cookies to maintain a persistent user session in the Ghostery Browser Extension.

This PR, moves away from setting the session cookies implicitly by `login` and `refresh_token` requests in favor of doing it explicitly. It also uses `Authorization Bearer` HTTP header instead of a cookie to authorize all API requests. 

We should test this in:
* [ ] Chrome - to ensure there is no breakage
* [ ] Firefox 
* [ ] Firefox with Strict mode privacy settings (in `about:preferences#privacy` select `Strict`)
* [ ] Firefox with first party isolation enabled (set `privacy.firstparty.isolate` to `true` in `about:config`)

### DIFF
--- a/src/utils/api.js
+++ b/src/utils/api.js
@@ -42,15 +42,18 @@ class Api {
 	}
 
 	async _sendReq(method, path, body) {
-		const { value: access_token } = await cookiesGet({ name: 'access_token' });
-		const csrfToken = await this._getCsrfCookie();
+		const	accessTokenCookie = await cookiesGet({ name: 'access_token' });
+		const	csrfTokenCookie = await cookiesGet({ name: 'csrf_token' });
+		if (!accessTokenCookie || !csrfTokenCookie) {
+			throw new Error('no cookie');
+		}
 		return fetch(`${this.config.ACCOUNT_SERVER}${path}`, {
 			method,
 			headers: {
 				'Content-Type': Api.JSONAPI_CONTENT_TYPE,
 				'Content-Length': Buffer.byteLength(JSON.stringify(body)),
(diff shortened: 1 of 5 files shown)


src/utils/api.js
```diff
@@ -35,24 +35,36 @@ class Api {
...
-			}));
+	async _sendReq(method, path, body) {
+		const	accessTokenCookie = await cookiesGet({ name: 'access_token' });
+		const	csrfTokenCookie = await cookiesGet({ name: 'csrf_token' });
+		return fetch(`${this.config.ACCOUNT_SERVER}${path}`, {
+			method,
+			headers: {
+				'Content-Type': Api.JSONAPI_CONTENT_TYPE,
+				'Content-Length': Buffer.byteLength(JSON.stringify(body)),
+				Authorization: `Bearer ${accessTokenCookie ? accessTokenCookie.value : ''}`,
```
philipp-classen, line 62: Wouldn't it be better to leave skip the Authorization header if there is no access token? From what I understand, it is standardized to be
```
Authorization: <auth-scheme> <authorization-parameters>
```

```
Authorization: Bearer
```
is telling I'm authorized and my (base64-decoded) token is empty

PR #817, opened 2022-07-28
Migrate from yarn to npm + node upgrade

chrmod:


webpack.config.js
```diff
@@ -17,6 +17,11 @@ const webpack = require('webpack');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const WebpackShellPlugin = require('webpack-shell-plugin');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+const sass = require('sass');
+const crypto = require('crypto');
+
+const crypto_orig_createHash = crypto.createHash;
+crypto.createHash = algorithm => crypto_orig_createHash(algorithm == 'md4' ? 'sha256' : algorithm);
```
philipp-classen, line 24: md4 or should it be rather md5? Do you remember why this was needed?

package.json
```diff
@@ -128,5 +127,8 @@
     "webpack": "^4.46.0",
     "webpack-cli": "^3.3.12",
     "webpack-shell-plugin": "^0.5.0"
+  },
+  "engines": {
+    "yarn": "YARN NO LONGER USED - use npm instead."
```
smalluban, line 132: is this a good practice, when the engine is going to be deprecated? I suppose that this field should take a version number, not a warning massage.

webpack.config.js
smalluban, line 24: Linter does not scan this file, but my editor suggests to change `==` to `===` ;)

smalluban (changes requested):
* `yarn.lock` leftover
* Not sure about the `engine` field
* some simple clean up
* No idea why we monkey patch the crypto

Nevertheless, I was able to build the project and run it in Chrome. I did not notice CSS issues (but the mixin is in the android folder, so I suppose that those styles are only applied on that platform).

webpack.config.js
chrmod: This is sadly something that I don't have a grasp on. But in short, webpack uses some hash functions that are no longer provided by the crypto package. A proper fix would be to update webpack, but that solution requires way more effort.

chrmod: I've added some comments to make it clear

philipp-classen (dismissed):

PR #820, opened 2022-08-01
MV3: simple telemetry

chrmod:
This PR extracts Metrcis module from `extension-manifest-v2` and shares it with `extenison-manifest-v3`. The usage on `extension-manifest-v3` for the time being is minimal as we want to test if the system works correctly.

The additional module `browser-infor` was extracted from `Globals`, again, to be used across Manifest v2 and Manifest v3 extensions. The module gained ability to detect Safari.


extension-manifest-v2/src/background.js
```diff
@@ -1538,6 +1540,20 @@ async function initializeAccount() {
 	}
 }
 
+async function recordUTMs() {
+	try {
+		if (globals.JUST_INSTALLED) {
+			const utms = await metrics.detectUTMs();
+			prefsSet(utms);
```
philipp-classen, line 1547: await prefsSet(utms);

It is arguably easier to reason about.

extension-manifest-v3/src/background/telemetry.js
```diff
@@ -0,0 +1,51 @@
...
+    conf: options,
+    log,
+    storage: metrics,
+    saveStorage: async (_storage) => {
+      await store.set(metrics, _storage);
+    },
+  });
+
+  try {
+    recordUTMs(telemetry);
```
philipp-classen, line 38: await missing

libs/node_modules/ua-parser-js/package.js
```diff
@@ -0,0 +1,12 @@
+Package.describe({
```
philipp-classen, line 1: Is it intended that the sources of dependencies (here `ua-parser-js`) are checked in under `node_modules`?

chrmod: nope, thank you for noticing!

extension-manifest-v2/package.json
```diff
@@ -48,8 +48,9 @@
   },
   "homepage": "https://github.com/ghostery/ghostery-extension#readme",
   "dependencies": {
-    "@ghostery/ui": "../ui",
     "@cliqz/url-parser": "^1.1.4",
+    "@ghostery/libs": "file:../libs",
```
philipp-classen, line 52: Is there a reason why "libs" and "ui" use different syntax ("file:../libs" vs "../ui")? Should it also be "file:../ui"? (I'm not familiar with local paths in package.json)

extension-manifest-v3/src/background/telemetry.js
```diff
@@ -0,0 +1,53 @@
...
+    storage: JSON.parse(JSON.stringify(metrics)),
+    saveStorage: async (_storage) => {
+      await store.set(metrics, _storage);
+    },
+  });
+
+  const JUST_INSTALLED = metrics.install_all === 0;
+
+  try {
+    await recordUTMs(telemetry);
```
philipp-classen, line 41: Do you need to pass "JUST_INSTALLED" as second parameter?

extension-manifest-v3/src/store/metrics.js
```diff
@@ -0,0 +1,41 @@
...
+
+const Metrics = {
+  ...VELOCITIES.reduce((all, current) => ({ ...all, [current]: [0] }), {}),
+  [store.connect]: {
+    async get() {
+      const { metrics = {} } = await chrome.storage.local.get(['metrics']);
+      VELOCITIES.forEach((velocity) => {
+        if (
+          (metrics[velocity] &&
+            metrics[velocity].length === 0 &&
```
philipp-classen, line 18: "length !== 0"? (Next check is accessing the first element)
Or should there be "||" instead of "&&"?

extension-manifest-v2/package.json
chrmod: there is no difference AFAICT

extension-manifest-v3/src/background/telemetry.js
chrmod: good catch, thanks

chrmod:
To make rollup and webpack by happy about `ua-parser-js` being a `@ghostery/libs` dependency, I had to patch the `ua-parser-js`. Here is the issue reported back to that package https://github.com/faisalman/ua-parser-js/issues/591

extension-manifest-v3/src/store/metrics.js
```diff
@@ -0,0 +1,41 @@
...
+  ...VELOCITIES.reduce((all, current) => ({ ...all, [current]: [0] }), {}),
+  [store.connect]: {
+    async get() {
+      const { metrics = {} } = await chrome.storage.local.get(['metrics']);
+      VELOCITIES.forEach((velocity) => {
+        if (
+          (metrics[velocity] &&
+            metrics[velocity].length === 1 &&
+            metrics[velocity][0] === 0) ||
+          !metrics[velocity]
```
chrmod, line 20: @smalluban what's the proper way to declare an array of numbers? `[0]` does that job but initializes the array with a leading `0`, which has to be removed in here.

extension-manifest-v3/src/background/telemetry.js
```diff
@@ -0,0 +1,53 @@
...
+  const options = await store.resolve(store.get(Options));
+  const metrics = await store.resolve(store.get(Metrics));
+
+  const telemetry = new Telemetry({
+    METRICS_BASE_URL: 'https://d.ghostery.com',
+    EXTENSION_VERSION: chrome.runtime.getManifest().version,
+    conf: options,
+    log,
+    // object copy as hybrics arrays are not extensible
+    storage: JSON.parse(JSON.stringify(metrics)),
```
chrmod, line 32: array properties of the model objects are immutable, thus I had to clone the object here. @smalluban how this should be used?

extension-manifest-v3/src/background/telemetry.js
```diff
@@ -0,0 +1,53 @@
...
+    saveStorage: async (_storage) => {
+      await store.set(metrics, _storage);
```
chrmod, line 34: @smalluban this does not work. (`_storage` is a plain object not a model). Store complains with some error that model is out of date or something similar

libs/patches/ua-parser-js+1.0.2.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/node_modules/ua-parser-js/package.json b/node_modules/ua-parser-js/package.json
```
chrmod, line 1: this patch is here as Rollup `preserveModules` adds a commonjs runtime for modules that has dynamic exports (like `ua-parser-js` which does `modules.exports = ` in IIFE). This could work, but rollup puts its commonjs runtime together with a module stub in `_virtaul` folder which breaks Chromium as folder names starting with `_` are prohibited in that engine.

extension-manifest-v3/src/store/metrics.js
smalluban: Unfortunately, It is an architectural limitation. Eighter the store can use values from the definition as defaults or not. It is more common, that you want to do the first. A good example is in the documentation [here](https://hybrids.js.org/#/store/model?id=primitives-or-nested-objects-internal).

extension-manifest-v3/src/background/telemetry.js
smalluban: By the design, the model structure is frozen, and cannot be changed after initialization. It is not a bad thing, as you discovered a weak spot of the `Telemetry` class - it should not mutate arguments passed to it :)
From the store perspective, the feature helps to make changes to the model by mistake thinking that it will update the model.

extension-manifest-v3/src/background/telemetry.js
smalluban: the metrics reference is set only once, so if you update the store again (which is likely), the model is stale (its new version is the current one). For the singleton definitions, you can use the definition as a reference for the `store.set()` method.

libs/patches/ua-parser-js+1.0.2.patch
smalluban: Ugly thing, but I get it. As we talked about, we should remove that package from dependencies in separate PR.

extension-manifest-v3/src/background/telemetry/index.js
```diff
@@ -0,0 +1,53 @@
+import { Metrics as Telemetry } from '@ghostery/libs';
+import { store } from 'hybrids';
+
+import Options from '/store/options.js';
+import UTMs from '/store/utms.js';
+
+import { loadStorage, saveStorage } from './storage';
```
smalluban, line 7: The `index.js` is still a quite small file, so I don't see a reason to take out those two functions into a separate file. Then you can back to the `background/telemetry.js` - which would contain the whole feature.

extension-manifest-v3/src/store/utms.js
```diff
@@ -0,0 +1,18 @@
+import { store } from 'hybrids';
```
smalluban, line 1: Is it something that will be used in hybrids components contexts at all? I was thinking about not using the store for both introduced models.

smalluban (changes requested):

extension-manifest-v3/src/background/telemetry/index.js
chrmod: done

extension-manifest-v3/src/store/utms.js
chrmod: ok, make sense

philipp-classen (approved):

PR #823, opened 2022-08-10
version bump

chrmod:


extension-manifest-v3/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@ghostery/extension-manifest-v3",
   "private": true,
-  "version": "9.5.0",
+  "version": "9.6.0",
```
smalluban, line 4: 9.5.0 is already upcoming version. It was bumped with onboarding, but not yet released.

smalluban (changes requested):

smalluban:
Generally, don't bump versions in package by hand, as it is now, you may forget about lock files. The npm has command for that.

philipp-classen (approved):

extension-manifest-v3/package.json
chrmod: fixed

smalluban (approved):

PR #825, opened 2022-08-10
Handling of unknown categories

chrmod:


extension-manifest-v3/src/pages/panel/views/detailed.js
```diff
@@ -28,7 +28,7 @@ export default define({
       ${store.ready(stats) &&
       html`
         ${Object.keys(stats.byCategory)
-          .sort((a, b) => order.indexOf(a[0]) - order.indexOf(b[0]))
```
chrmod, line 31: looks like we had a bug in here

ui/src/utils/categories.js
```diff
@@ -89,3 +94,24 @@ export const labels = {
...
+
+export const getCategoryColor = (category) => {
+  return colors[getCategoryKey(category)];
+};
+
+export const getCategoryLabel = (category) => {
+  return labels[getCategoryKey(category)];
+};
+
+export const sortCategories = (_resolveCategoryName) => {
```
smalluban, line 110: You can avoid underscore prefix and just do:
```suggestion
export const sortCategories = (resolveCategoryName = ((a) => a)) => {
```

smalluban (changes requested):
Categories are shared between v2 and v3, so you should also run `npm run locales` in v2, to add new category to its localization files.

extension-manifest-v3/src/pages/panel/views/detailed.js
smalluban: I assume that added `sortCategories` function solves the problem, isn't it? ;)

smalluban (approved):
It's fine now, but I made a typo in the suggestion, so there are to many braces, you must remove one.

PR #827, opened 2022-08-11
Shared prettier config

chrmod:


smalluban (changes requested):
Regarding failed tests, I think you should be able to just put the configuration in the root path of the repository, isn't it?

smalluban (approved):

PR #829, opened 2022-08-11
Remove adblocker settings

### LINKED ISSUE #811: Streamline Adblocker settings
Right now the blocking settings are manager by two mechanisms: tracker list and adblocker lists. This can sometimes be confusing thus we aim to simplify the way users can customize their choices. 

TODO:
* [ ] Remove "Ad-block Lists" from settings
* [ ] flip all users to use full list for adblocking, that is to block Ads, Trackers and Popups
* [ ] Adblocker toggle should affect all blocking 
* [ ] General settings -> Trackers -> Update now; should force update adblocker lists

chrmod:
fix #811


extension-manifest-v2/src/background.js
```diff
@@ -756,7 +756,8 @@ function onMessageHandler(request, sender, callback) {
 		return true;
 	}
 	if (name === 'update_database') {
-		updateDBs().then((result) => {
+		updateDBs().then(async (result) => {
+			await common.modules.adblocker.background.adblocker.update();
```
philipp-classen, line 760: Do we need to recover from errors here? The old code also did not try (if updateDBs failed the callback does not fire). Perhaps it is not needed here, so I'm fine with merging as is. Just a comment.

philipp-classen (approved):

Issue #806, opened 2022-06-28
Error in console: Uncaught (in promise)


Kreash:
On any site, an error appears in the console:

`Uncaught (in promise) 
{message: 'A listener indicated an asynchronous response by r…age channel closed before a response was received'}
message: "A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received"`

It interferes quite a lot when working in the console

### Steps to Reproduce

To check, you can go to https://www.google.com/, open the console and click on an empty space.
Several errors will appear almost immediately.

### Versions

+ Browser: Chrome 102.0.5005.115

Kreash:
there is also an open question on stackoverflow about this error:

https://stackoverflow.com/questions/72494154/a-listener-indicated-an-asynchronous-response-by-returning-true-but-the-messag

philipp-classen:
I get it too with a fresh installation of Ghostery on Chromium.

```
A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received 
```

philipp-classen:
Can also be reproduced on the Ghostery homepage (https://www.ghostery.com), so it is not caused by any of the scripts on the Google search page. From the Stackoverflow post, it might have started in Chrome 102 or 103.

Most likely it comes from the adblocker when it injects its code through content_script into the main page. Somewhere we are not properly handling all errors. I don't think it will affect functionality for average users, but it can be annoying for frontend developers.

Unfortunately, it is not straightforward to figure out where the error handling is missing.  But error messages share that common part; at least that is a start:

```
   | Promise.then (async) |   |  
-- | -- | -- | --
   | (anonymous) | @ | content_script_bundle.js:1495
   | handleNewFeatures | @ | content_script_bundle.js:273
   | handleUpdatedNodes | @ | content_script_bundle.js:289
   | (anonymous) | @ | content_script_bundle.js:232
```

Kreash:
I first encountered this error on my ad-free project with the Ghostery extension enabled.
And at the same time, Ghostery did not block anything.
Maybe this information will help somehow

philipp-classen:
My feeling is that the code that is triggering it, is in the adblocker.

Based on the observation that [`handleUpdatedNodes`](https://github.com/ghostery/adblocker/blob/5fb9df7afb511a14db18cf75f2a0fb91f927ff3b/packages/adblocker-content/adblocker.ts#L218) shows up in the log, and because I don't get the error when adblocking in Ghostery is disabled.

[...]

PR #830, opened 2022-08-11
Apply patch for webextension-polyfill to silence Chrome warning

philipp-classen:
Apply patch https://github.com/mozilla/webextension-polyfill/pull/385 from the webextension-polyfill (and removes the minified file). The patch silences the following warning that started to appear in recent Chrome versions:

```
A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received
```

refs #806


chrmod (approved):

chrmod (changes requested):
Please add a note about it in changelog

chrmod (approved):

PR #849, opened 2022-09-01
Trackers preview in UI

### LINKED ISSUE #837: [iOS] [Tracker Wheel] Missing Tracker Wheel right to the search results.
Missing Tracker Wheel right to the search results. 

### Expected Behavior
There should be a possibility to see Tracker Wheel in the search results. And then, there should be a possibility to click it and open a popup with more information.

### Actual Behavior
When the option "Disable Tracker Wheel in browser toolbar" is disabled, Tracker Wheel should be visible in the search results.
(Screenshots were added at the bottom.)

### Steps to Reproduce
1. Install "Ghostery - Privacy Ad Blocker".
2. Turn it on. Open Configuration (Settings-> Safari-> Extensions). 
3. Open Extensions settings. Leave "Disable Tracker Wheel in browser toolbar" disabled.
4. Open Safari.
5. Search for example: test, BMW, facebook.
6. Try to find Tracker Wheel.

### Versions
+ Browser: Safari
+ OS: iOS 15.4.1
+ Version of Ghostery - Privacy Ad Blocker: 2.5.1 (51) 

For example:
obraz
obraz
obraz

smalluban:
* Updates wtm packages to 0.4.2 which fixes #837 
* Moves trackers-preview Ui into the `ui/src/modules/trackers-preview` module
* Update v2 and v3 extensions to use new version of the package
* Fixes hidpi issue with canvas drawing


.prettierignore
```diff
@@ -0,0 +1 @@
+extension-manifest-v2/**/*
```
smalluban, line 1: Prettier in v2 is in conflict with eslint rules. I don't know a better option for now than disabling it there.

extension-manifest-v2/app/trackers-preview/trackers-preview.js
```diff
@@ -0,0 +1,38 @@
...
+
+import { define, html } from 'hybrids';
+
+const domain = new URLSearchParams(window.location.search).get('domain');
+const stats = getStats(domain);
+
+updateIframeHeight();
+
+export default define({
+	tag: 'gh-trackers-preview',
```
chrmod, line 27: indentation here looks weird

ui/src/modules/trackers-preview/trackers-preview.js
```diff
@@ -0,0 +1,170 @@
...
+      <button class="svg-button" onclick="${(host) => dispatch(host, 'close')}">
+        <ui-icon name="close"></ui-icon>
+      </button>
+    </ui-header>
+
+    <main>
+      <h1>Trackers Preview</h1>
+
+      ${stats &&
+      html.resolve(
```
chrmod, line 30: indentation problem? why pretties does not catch it?

chrmod (changes requested):
Looks great! just two nit picks on the indentation

extension-manifest-v2/app/trackers-preview/trackers-preview.js
```diff
@@ -0,0 +1,38 @@
...
+	getStats,
+	close,
+	disable,
+	updateIframeHeight
+} from '@whotracksme/webextension-packages/packages/trackers-preview/src/page_scripts';
+
+import { define, html } from 'hybrids';
+
+const domain = new URLSearchParams(window.location.search).get('domain');
+const stats = getStats(domain);
```
philipp-classen, line 22: Can we skip the call here if domain is "null"? Maybe it works as is, but getStats is not expected to deal with nullable values (and arguably should not need to): https://github.com/whotracksme/webextension-packages/blob/main/packages/trackers-preview/src/page_scripts/index.js#L12

Also, wouldn't `window.location` be always the search page here?

ui/src/modules/trackers-preview/trackers-preview.js
smalluban: Yes, it did (if it would be a problem here or anywhere else, the CI test would break). What do you mean about the "problem"? Prettier splits the conditions in templates into separate lines. It is how it works.

extension-manifest-v2/app/trackers-preview/trackers-preview.js
smalluban: It cannot be null, as the iframe is only loaded by the content script, which runs on google search results, and generates a link to it with a target result domain.

extension-manifest-v2/app/trackers-preview/trackers-preview.js
smalluban: V2 uses tabs, not spaces, but GH shows 1 tab like 4 spaces - but the project defines tab as 2 spaces 🤦 Not much we can do, if we stick with the current styling.

philipp-classen (approved):

chrmod (approved):

PR #858, opened 2022-09-07
[v2] Add autoconsent library

smalluban:
Implements autoconsent feature to v2 extension.

* Moves common UI logic into `ui/` package
* Updates Panel/Android Panel to have autoconsent switch instead of smart browsing
* Adds messages from autoconsent popup and Panel


extension-manifest-v2/src/background.js
```diff
@@ -552,6 +556,30 @@ function onMessageHandler(request, sender, callback) {
...
+			}
+
+			return false;
+		}
+		if (name === 'disable') {
+			if (message.url) {
+				conf.autoconsent_blacklist = conf.autoconsent_blacklist.concat(message.url);
+			} else {
+				conf.enable_autoconsent = false;
+				conf.autoconsent_whitelist = [''];
```
chrmod, line 576: is this correct to have an empty string here?

extension-manifest-v2/src/modules/autoconsent.js
```diff
@@ -0,0 +1,99 @@
+/**
```
chrmod, line 1: how about calling the folder background?

extension-manifest-v2/webpack.config.js
```diff
@@ -153,7 +160,7 @@ module.exports = {
 				],
 			}, {
 				test: /\.(js|jsx)$/,
-				include: [SHARED_COMP_DIR, ONBOARDING_DIR, PANEL_ANDROID_DIR, PANEL_DIR, LICENSES_DIR, CONTENT_SCRIPTS_DIR],
+				include: [APP_DIR],
```
chrmod, line 163: not sure what the consequences of this change may be? wouldn't it be safer to add another dir?

chrmod (approved):
Looks great. 

I have few questions but they may not be important.

extension-manifest-v2/src/classes/ConfData.js
```diff
@@ -107,6 +107,7 @@ class ConfData {
 			_initProperty('current_theme', 'default');
 			_initProperty('enable_ad_block', false);
 			_initProperty('enable_anti_tracking', false);
+			_initProperty('enable_autoconsent', true);
```
philipp-classen, line 110: Would it make sense to have the same logic as adblocking and anti-tracking? Starting with being disabled and then activating after onboarding?

chrmod: Good catch. Makes perfect sense

extension-manifest-v2/src/background.js
```diff
@@ -129,6 +131,8 @@ function tryOpenOnboarding() {
 		return;
 	}
 
+	conf.enable_autoconsent = false;
```
philipp-classen, line 134: See comment in ConfData. If it would start with false and gets enabled at the end of the onboarding, this might not be needed (like for adblocking and anti-tracking).

extension-manifest-v2/src/background.js
smalluban: This is a trick to differ from the state where the user choose his preference for all or selected websites. However, I think it might be confusing so that I will change that to `null` or an array with a list of domains.

extension-manifest-v2/src/modules/autoconsent.js
smalluban: The `src` is already a `background` in your sence. Usually, there are classes, so there is `src/classes` path, but I would avoid that pattern with self initialized class (it should be es module). This is why I named it `modules`.

extension-manifest-v2/webpack.config.js
smalluban: It was listing all of the subpaths of the `APP_DIR`, but not "new ones", which we added recently. It should transpile all or nothing, so I think it is quite safe.

extension-manifest-v2/src/classes/ConfData.js
smalluban: This is not that simple. We have to cover a case, where users update from an earlier version. Then, it has to be set to `true` (there is no other option, which we could migrate). I did not find a better place to initialize if that happens, so if it is `true` by default and disabled if onboarding triggers, every case is covered.

extension-manifest-v2/src/background.js
smalluban: See my comment above. We have to cover a case of user, which updates with onboarding already completed.

extension-manifest-v2/src/classes/ConfData.js
philipp-classen: I see. I think the canonical way in Ghostery to handle migrations is to add code here:

https://github.com/ghostery/ghostery-extension/blob/f7cc668efd375cae83fc9b60741040731dd7d78e/extension-manifest-v2/src/background.js#L1287

```
function initializeVersioning() {
  ...
  if (globals.JUST_UPGRADED) {
    ...
    if (utils.semverCompare(PREVIOUS_EXTENSION_VERSION, '8.8.0') < 0) {
	conf.enable_autoconsent = true;
    }
  }
}
```

That would run once for users before 8.8.0. And initializeVersioning also runs before initializeGhosteryModules, which triggers the onboarding.

chrmod (approved):

extension-manifest-v2/src/classes/ConfData.js
smalluban: Ok, I will change to use your suggestion.

philipp-classen: Thanks, one thing that I just realized. 8.8.0 might need to be changed, since we pushed that already in Ghostery Dawn. Instead of 8.8.0 it should be the next release version. (I thought 8.7.6 is the latest, but it is 8.8.0. Sorry).

smalluban: I used `8.9.0`.

smalluban:
@chrmod @philipp-classen I made a fix for the cases, which Philipp has found. Especially economist.com and spiegel.de uses iframe with a subdomain to show the CMPs. 

At the same time, I fixed the v3, so the last commit is a little bit bigger. To make the code cleaner I took some changes from another PR, which updates the syntax of the options arrays (allowing them to be empty) and simplifies access to store models.

https://github.com/ghostery/ghostery-extension/pull/858/commits/c0598697080d4fc33a5fb33c84f0d4c4d02cac76

philipp-classen (approved):

PR #871, opened 2022-09-16
User's data migration from v2

smalluban:
* Adds migration logic for a case, where Chrome users will update the extension from v2 (8.x) to v3 (9.x)
* Uses hybrids 8.1.3 with a simpler `store.resolve`, and empty arrays support for store models

After a comparison of `chrome.storage.local` and `indexedDB` between v2 and v3, there is nothing that should be shared (kept after the extension update). The V3 uses `indexedDB` only for storing communication public keys. The local storage options are migrated when options are fetched.

The solution does not use `chrome.runtime.onInstalled` event because, it might be too late, and options might be already called. I figure out a different way to make it efficient (not call it every time) and ensure that whatever timing options would be needed, the result will be correct. Generally, the migration is only called if `options` are not set in the storage, and the manifest uses version `3`, which is only valid for Chrome.


chrmod (commented):
This PR looks just fine. 

But it ask for closer analysis. Maybe we could add comments for both mv3 options keys and mv2 conf keys to ensure we have them aligned.

It also ask that we design Settings Sync for MV3 right now, as those also will have to be migrated on MV3 upgrade.

If this help we can merge it now.

smalluban:
Moved to #936

PR #883, opened 2022-09-30
Pause feature to whitelist domains

smalluban:
Adds core logic for the pause feature. The UI has a pause button that whitelists a domain for a minute - blocks cosmetics and adds the 'allow' rule to the dynamic DNR list.

Additionally, I found a bug, which throws a lot of error messages in the console caused by to early closed messaging channel in the cosmetics content script.

We can merge the feature without showing it to the users - we just need to hardcode false condition here - https://github.com/ghostery/ghostery-extension/pull/883/files#diff-ab787eda050f8d097068bea78c9043e942c4480fdb99acf1f1c0213b0bbd859dR159

### DIFF
--- a/extension-manifest-v3/src/background/cosmetics.js
+++ b/extension-manifest-v3/src/background/cosmetics.js
@@ -23,12 +23,17 @@ const adblockerEngines = DNR_RULES_LIST.reduce((map, name) => {
   };
   return map;
 }, {});
+let pausedDomains = [];
 
 let adblockerStartupPromise = (async function () {
   await observe('dnrRules', (dnrRules) => {
     DNR_RULES_LIST.forEach((key) => {
       adblockerEngines[key].isEnabled = dnrRules[key];
     });
+  });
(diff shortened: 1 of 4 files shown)


smalluban:
I tested the feature on Safari (16.0, desktop on macOS). I don't have good news... It looks that dynamic rules are supported (not sure from which version), but they break everything. When the extension tries to update those, in the result it might be even undefined rather than an empty array (also for static rules). It does not work either. 

After using the dynamic rules API updating static list breaks, and throws weird errors. Unfortunately, we may be forced to not use this API until Apple will fix it. 

I think we can skip the dynamic lists, but keep the feature, which will block the cosmetics on the specific domain. This is usually sufficient to unblock "blocked" sites, which detect adblocker. Also, Safari users just don't have right now another option.

smalluban:
This PR was pulled into the https://github.com/ghostery/ghostery-extension/pull/936

PR #884, opened 2022-09-30
Better error recovery if cookies are partially missing

philipp-classen:
It can happen that the extension thinks a user is logged in, even though essential information stored in the cookies, for instance, is missing. The result is that it gets stuck in a state where it suggests in the UI that the user is logged in, but it is not. For instance, the account name in the UI is blank.

The changes contain bug fixes and better error recovery to force a logout in those cases. A successful login should then reset the cookies to a valid state.


extension-manifest-v2/src/utils/api.js
```diff
@@ -33,22 +33,27 @@ class Api {
 			return this._refreshPromise;
 		}
 
-		this._refreshPromise = fetch(`${this.config.AUTH_SERVER}/api/v2/refresh_token`, {
-			method: 'POST',
-			credentials: 'omit',
-		}).then((res) => {
-			if (res.status > 400) {
-				throw res.json();
```
chrmod, line 41: this path is no removed. Is this intentional? afair callers are testing this returned value for errors.

philipp-classen: It is intentional. This is the part where it is being called:
https://github.com/ghostery/ghostery-extension/blob/455a51f087084baa7885557c00513ee2cf1c2568/extension-manifest-v2/src/utils/api.js#L125

```
this.refreshToken()
	.then((res) => {
		const { status } = res;
		if (status >= 400) {
		  ...
		}
		this._sendReq(method, path, body)
		 ...
	..
        .catch(err => reject(err));
```

So, the caller of refreshToken expects that refreshToken does not reject and it will handle the status >= 400 itself.

extension-manifest-v2/src/utils/api.js
```diff
@@ -33,22 +33,27 @@ class Api {
...
+		const pending = (async () => {
+			const response = await fetch(`${this.config.AUTH_SERVER}/api/v2/refresh_token`, {
+				method: 'POST',
+				credentials: 'omit',
```
chrmod, line 39: this request is lacking any credentials, it cannot work. lets try to fix it in this PR

chrmod (changes requested):

extension-manifest-v2/src/utils/api.js
philipp-classen: The experimental authentication server API will try now first the HTTP headers `UserId` and `RefreshToken` and only then fallback to cookies.

chrmod (approved):
Looks great!

PR #891, opened 2022-10-10
Update copy in settings

chrmod:


smalluban (commented):
It's ok, but I am unsure how we set up the translation process. If we generate it from the main, it's okay. 
The second thing is that we used to use upper letters in the headings.

smalluban:
Another question: As we change `options` to `settings`, should we move paths from `pages/options` to `pages/settings` as well?

chrmod:
We decided to call it `Ghostery settings` (small caps) on all platforms. Technically it is in option page, as specified in the manifest, so I think the file name is adequate.

PR #916, opened 2022-10-27
Fix cosmetics background feature

smalluban:
The `injectCosmetics` uses an MV2 hack to promisify the results, which is not required in MV3, and what is more important without the fix, we have thrown an error in each tab:

Zrzut ekranu 2022-10-27 o 12 29 32

Additionally, I have found, that `adblockerInjectStylesWebExtension` returns a promise, which is not consumed at all, so I cleaned it up.

The third thing.. the specific scripts were just omitted in the `sendResponse` callback. I assume this is a bug, so there is a fix.

### DIFF
--- a/extension-manifest-v3/src/background/cosmetics.js
+++ b/extension-manifest-v3/src/background/cosmetics.js
@@ -281,7 +281,9 @@ chrome.webNavigation.onCommitted.addListener((details) => {
 
 chrome.runtime.onMessage.addListener((msg, sender) => {
   if (msg.action === 'getCosmeticsFilters') {
-    adblockerOnMessage(msg, sender);
+    adblockerOnMessage(msg, sender).catch((e) =>
+      console.error(`Error while processing cosmetics filters: ${e}`),
+    );
   }
 
   return false;
--- a/extension-manifest-v3/src/background/cosmetics.js
+++ b/extension-manifest-v3/src/background/cosmetics.js
@@ -45,7 +45,7 @@ const adblockerStartupPromise = (async function () {
   );
 })();
 
-async function adblockerInjectStylesWebExtension(
+function adblockerInjectStylesWebExtension(
   styles,
   { tabId, frameId, allFrames = false },
 ) {
@@ -54,40 +54,34 @@ async function adblockerInjectStylesWebExtension(
     return;
   }
 
-  // Proceed with stylesheet injection.
-  return new Promise((resolve) => {
-    if (chrome.scripting && chrome.scripting.insertCSS) {
-      const target = {
-        tabId,
-      };
+  if (chrome.scripting && chrome.scripting.insertCSS) {
+    const target = {
+      tabId,
+    };
 
-      if (frameId) {
-        target.frameIds = [frameId];
-      } else {
-        target.allFrames = allFrames;
-      }
-      chrome.scripting.insertCSS(
-        {
-          css: styles,
-          origin: 'USER',
-          target,
-        },
-        () => resolve,
-      );
+    if (frameId) {
+      target.frameIds = [frameId];
     } else {
-      const details = {
-        allFrames,
-        code: styles,
-        cssOrigin: 'user',
-        matchAboutBlank: true,
-        runAt: 'document_start',
-      };
-      if (frameId) {
-        details.frameId = frameId;
-      }
-      chrome.tabs.insertCSS(tabId, details, () => resolve);
+      target.allFrames = allFrames;
     }
-  });
+    chrome.scripting.insertCSS({
+      css: styles,
+      origin: 'USER',
+      target,
+    });
+  } else {
+    const details = {
+      allFrames,
+      code: styles,
+      cssOrigin: 'user',
+      matchAboutBlank: true,
+      runAt: 'document_start',
+    };
+    if (frameId) {
+      details.frameId = frameId;
+    }
+    chrome.tabs.insertCSS(tabId, details);
+  }
 }
 
 // copied from https://github.com/cliqz-oss/adblocker/blob/0bdff8559f1c19effe278b8982fb8b6c33c9c0ab/packages/adblocker-webextension/adblocker.ts#L297
@@ -205,13 +199,13 @@ async function adblockerOnMessage(msg, sender, sendResponse) {
     sendResponse({
       active: specificResponses.map((r) => r.active).some((a) => a),
       extended: specificResponses.map((r) => r.extended).flat(),
-      styles: '',
+      scripts: specificResponses.map((r) => r.scripts).flat(),
     });
   }
 }
 
 async function executeScriptlets(tabId, scripts) {
-  // Dynamically injected scripts scripts can be difficult to find later in
+  // Dynamically injected scripts can be difficult to find later in
   // the debugger. Console logs simplifies setting up breakpoints if needed.
   let debugMarker;
   if (DEBUG_SCRIPLETS) {
--- a/extension-manifest-v3/src/background/cosmetics.js
+++ b/extension-manifest-v3/src/background/cosmetics.js
@@ -66,11 +66,13 @@ function adblockerInjectStylesWebExtension(
     } else {
       target.allFrames = allFrames;
     }
-    chrome.scripting.insertCSS({
-      css: styles,
-      origin: 'USER',
-      target,
-    });
+    chrome.scripting
+      .insertCSS({
+        css: styles,
+        origin: 'USER',
+        target,
+      })
+      .catch((e) => console.error('Failed to inject CSS', e));
   } else {
     const details = {
       allFrames,
@@ -82,7 +84,9 @@ function adblockerInjectStylesWebExtension(
     if (frameId) {
       details.frameId = frameId;
     }
-    chrome.tabs.insertCSS(tabId, details);
+    chrome.tabs
+      .insertCSS(tabId, details)
+      .catch((e) => console.error('Failed to inject CSS', e));
   }
 }
 
--- a/extension-manifest-v3/src/background/cosmetics.js
+++ b/extension-manifest-v3/src/background/cosmetics.js
@@ -24,7 +24,7 @@ const adblockerEngines = DNR_RULES_LIST.reduce((map, name) => {
   return map;
 }, {});
 
-const adblockerStartupPromise = (async function () {
+let adblockerStartupPromise = (async function () {
   await observe('dnrRules', (dnrRules) => {
     DNR_RULES_LIST.forEach((key) => {
       adblockerEngines[key].isEnabled = dnrRules[key];
@@ -43,6 +43,8 @@ const adblockerStartupPromise = (async function () {
       adblockerEngines[engineName].engine = engine;
     }),
   );
+
+  adblockerStartupPromise = null;
 })();
 
 function adblockerInjectStylesWebExtension(
@@ -85,13 +87,14 @@ function adblockerInjectStylesWebExtension(
 }
 
 // copied from https://github.com/cliqz-oss/adblocker/blob/0bdff8559f1c19effe278b8982fb8b6c33c9c0ab/packages/adblocker-webextension/adblocker.ts#L297
-async function adblockerOnMessage(msg, sender, sendResponse) {
-  await adblockerStartupPromise;
+async function adblockerOnMessage(msg, sender) {
+  if (adblockerStartupPromise) {
+    await adblockerStartupPromise;
+  }
 
   const genericStyles = [];
   const specificStyles = [];
   let specificFrameId = null;
-  const specificResponses = [];
 
   Object.keys(adblockerEngines).forEach((engineName) => {
     if (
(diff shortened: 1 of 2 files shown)


extension-manifest-v3/src/background/cosmetics.js
```diff
@@ -205,13 +199,13 @@ async function adblockerOnMessage(msg, sender, sendResponse) {
     sendResponse({
       active: specificResponses.map((r) => r.active).some((a) => a),
       extended: specificResponses.map((r) => r.extended).flat(),
-      styles: '',
+      scripts: specificResponses.map((r) => r.scripts).flat(),
```
chrmod, line 202: this is removed by purpose as we inject scripts with webNavigation listener https://github.com/ghostery/ghostery-extension/commit/75fd18ad98676528782fb5dc9a9f6eb1038d1bd8#diff-b70e3b1b9a717375c4c1fedc6bb7c6e2fca37465d4c4b2f0d223579ecb4a4beaL203

extension-manifest-v3/src/content_scripts/cosmetics.js
```diff
@@ -10,4 +10,13 @@
...
 import { injectCosmetics } from '@cliqz/adblocker-webextension-cosmetics';
 
-injectCosmetics(window);
+async function getCosmeticsFilters(args) {
+  const result = await chrome.runtime.sendMessage({
+    action: 'getCosmeticsFilters',
+    ...args,
+  });
+
+  return result || {};
```
chrmod, line 19: so is this a bug in https://github.com/ghostery/adblocker/blob/1c74ba647bb60ebb2627a6d37afc18a1e71e3f2d/packages/adblocker-webextension-cosmetics/adblocker.ts#L67

but does this change anything? a pending promise in this case has same result as resolution with empty object

extension-manifest-v3/src/background/cosmetics.js
```diff
@@ -45,7 +45,7 @@ const adblockerStartupPromise = (async function () {
   );
 })();
 
-async function adblockerInjectStylesWebExtension(
+function adblockerInjectStylesWebExtension(
```
chrmod, line 48: it doesn't change anything, right? (it's fine to simplify, just changing for potential regressions)

chrmod (changes requested):

extension-manifest-v3/src/background/cosmetics.js
philipp-classen, line 202: Unless I'm missing something, scripts are intentionally not returned, since they are executed with "chrome.scripting" (on navigation events)

extension-manifest-v3/src/content_scripts/cosmetics.js
philipp-classen: I think, definitions of bug vary. For me, the practice of using dangling promises is not a bug, since it should not break any functionality in a way that a user can observe. Now, the situation changed somewhat when Chrome started to show warnings in that case. You can now argue that use of the pattern is a bug in Chrome because of the warning.

I'm not against cleaning up the use of the pattern. But personally, I would avoid the term bug here since it misleading: it suggest that the adblocker is broken, even though for 99,9999% of the users it should not be a problem. Only for developers, seeing warnings for Ghostery might be annoying if it happens.

extension-manifest-v3/src/background/cosmetics.js
smalluban: It is not the same - https://github.com/ghostery/ghostery-extension/blob/fix-mv3-cosmetics/extension-manifest-v3/src/background/cosmetics.js#L281

We use other flags in the generation, so I assume the `scripts` list can differ.

extension-manifest-v3/src/content_scripts/cosmetics.js
smalluban: I disagree. If we support Chrome (and we will soon), and there is an error shown in the console, it means that it is a bug. And the behavior is unexpected (the connection should not be closed before the response is received). Whatever other browsers do with that *wrong behavior*, it should be fixed. And what is most important, the fix does not break anything but makes the code work better in `chrome.runtime` API in manifest V3.

extension-manifest-v3/src/background/cosmetics.js
smalluban: See my comment https://github.com/ghostery/ghostery-extension/pull/916#discussion_r1006790485

extension-manifest-v3/src/content_scripts/cosmetics.js
chrmod: sure, lets fix, and then we can issue a fix to cliqz/adblocker

extension-manifest-v3/src/background/cosmetics.js
chrmod: @remusao  maybe we can keep both? it should not harm, right?

chrmod: lets move this discussion to the other thread

remusao: I'd support what Philipp say, we should inject scriptlets only once per frame. If it's already done with webNavigation listener then let's not do it again from content script.

extension-manifest-v3/src/content_scripts/cosmetics.js
philipp-classen: Sure, let's fix the error messages in the console (like this PR does, which is good). I'm all for that. (When backporting to the adblocker engine, you have to be careful though; we learned that it is easy to introduce subtle problem with similar changes).

But if the goal is to improve the code and avoid the use bad patterns, I would strongly recommend to treat unhandled promises on a similar level. Here we remove currently one bad pattern but introduce another (easy to fix though: catching and logging a warning when the scripting API calls are rejected should be enough).

When we can it bugs or whatever, error messages from unhandled rejects are borderline unusable and a pain when debugging: sometimes you get stack traces, not not always.

Just try out to make a call to `chrome.scripting.executeScript` fail by changing the script to `broken`; this is then all you get to work with in the console:

```
Uncaught (in promise) Error: Could not load file: 'broken'. index.js:1
```

Sometimes you get stack traces, but cannot rely on it; I assume it is browser specific. If you have to debug an error as above, you have nothing to work with:

errors


Note that in Node.js, they decided to make unhandled promises terminate the application. If we you Node.js for tests that is also something to keep in mind. When we are touching code to clean up warnings from the JavaScript engine, I would put unhandled promises in the same category.

So, if warning in the console are bugs, I would argue that unhandled rejection fall into a similar class; we should probably not do in the new code. Most of the time, catching and logging a warning is sufficient if you want to neutralize an error from a promise.

smalluban: Added a catch, so I think this completes the discussion.

extension-manifest-v3/src/background/cosmetics.js
smalluban: So... @chrmod @remusao @philipp-classen Can you answer a simple question - is there a difference between those two calls to the API:

```
const { scripts } = engine.getCosmeticsFilters({ ..., getExtendedRules: true, ... });
```

And:
```
const { scripts } = engine.getCosmeticsFilters({ ..., getExtendedRules: false, ... });
```

If it does, then our injected scripts on navigate are not the same as those generated when the content script sends `getCosmeticsFilters` message.

If you answer this question, then we have the following options:
1. Remove condition starting at L174 completely
2. accept the change
3. Make correct change - as maybe those scripts should be passed in another way?

remusao: Extended rules (`getExtendedRules`) should have nothing to do with scripts so these two calls should be equivalent (in fact I think these are disabled in Ghostery's adblocker so it does not matter which value you give to `getExtendedRules`).

smalluban: Cool, thanks for the answer, so we can remove condition completely :)

smalluban: I'll make a change

extension-manifest-v3/src/content_scripts/cosmetics.js
philipp-classen: Actually, it was in two different places what I meant. The calls to:
* chrome.scripting.insertCSS
* chrome.tabs.insertCSS

For getCosmeticsFilters, I'm not even sure if the code itself should neutralize it? Before, errors were already handled there (passed to the caller). I would personally leave it like that: letting getCosmeticsFilters return a rejected promise and let the caller decide how to handle it.

smalluban: Ok, so I had to misunderstand your comment. I am not trying to fix all of the issues (not hitting us especially). This refactor is a result of seeing those logs in the console, which I know are very annoying to the people (of course those who use the console, so mainly developers). However, I have seen stack overflow questions about the Ghostery MV2 console message, and the answer is to turn off Ghostery. 

I think it is quite important to not have those responses in the community, even if in academic terms this is not "a bug". 
Here you have a function that is used if you don't pass yours - https://github.com/ghostery/adblocker/blob/master/packages/adblocker-webextension-cosmetics/adblocker.ts#L59.

From my interpretation, it does not use `reject` at all and only resolves if the result is fine - so it also might leed to a never-ending pending promise.

My solution is only for using native promise-based MV3 API, nothing more - it uses just the same logic inside.

extension-manifest-v3/src/background/cosmetics.js
smalluban: @remusao @chrmod @philipp-classen I have started cleaning up the condition, and if it is really not required, then in that case, this background process should not respond to the content script. What worries me, is that if there are no `scripts`, then the returned `extended` is not also passed there, but I think it should.

I check the adblocker source code, and those `extended` seems to be styling rules:
https://github.com/ghostery/adblocker/blob/master/packages/adblocker/src/engine/bucket/cosmetic.ts#L565

Currently, they are passed, only if there are also scripts, but those scripts are not included, but only `extended`. Is it anyway correct? Are those `extended` related to `scripts`?

remusao: https://github.com/ghostery/adblocker-filters/blob/main/src/configs.ts#L121

We have disabled extended selectors for Ghostery so these will always be empty. Also, there is not relation between extended and scripts, they are totally independent features (one can be enabled without the other).

smalluban:
After a discussion with @remusao, I made another round of cleanup. @chrmod @philipp-classen Please take a look again. 

Hard facts after the discussion:
* Whole condition starting in line 174 can be removed -> no more `sendResponse()` needed -> no need to wait for the response in content script -> everything is injected by the background script.
* Additionally I made a performance fix, to not wait for the initialize every time, even it already has been done

extension-manifest-v3/src/content_scripts/cosmetics.js
philipp-classen: I fully agree that we should fix errors that spam the page. Fixing them is OK.

For the locations that I see, let me quickly add comment there. Glancing over the changes, it introduces unhandled promises at two places (we can catch and log them).

extension-manifest-v3/src/background/cosmetics.js
```diff
@@ -54,50 +56,45 @@ async function adblockerInjectStylesWebExtension(
...
-        runAt: 'document_start',
-      };
-      if (frameId) {
-        details.frameId = frameId;
-      }
-      chrome.tabs.insertCSS(tabId, details, () => resolve);
+      target.allFrames = allFrames;
     }
-  });
+    chrome.scripting.insertCSS({
```
philipp-classen, line 69: this returns a promise (catch and log should be good, we don't need to await it)

extension-manifest-v3/src/content_scripts/cosmetics.js
```diff
@@ -10,4 +10,13 @@
...
+function getCosmeticsFilters(args) {
+  chrome.runtime.sendMessage({
+    action: 'getCosmeticsFilters',
+    ...args,
+  });
+
+  return {};
+}
+
+injectCosmetics(window, true, getCosmeticsFilters);
```
philipp-classen, line 22: I think, getCosmeticFilters needs to return a promise. While testing, I got this error on spiegel.de:

```
cosmetics.js:600 Uncaught TypeError: getCosmeticsFilters(...).then is not a function
    at injectCosmetics (cosmetics.js:600:86)
    at cosmetics.js:647:5
    at cosmetics.js:649:3
```

I think that is the code in the library that throws:
https://github.com/ghostery/adblocker/blob/692b84f6d2de57034e5a99e2a17fbcc3570126c6/packages/adblocker-webextension-cosmetics/adblocker.ts#L259

extension-manifest-v3/src/background/cosmetics.js
```diff
@@ -297,9 +279,9 @@ chrome.webNavigation.onCommitted.addListener((details) => {
   });
 });
 
-chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
+chrome.runtime.onMessage.addListener((msg, sender) => {
   if (msg.action === 'getCosmeticsFilters') {
-    return adblockerOnMessage(msg, sender, sendResponse);
+    adblockerOnMessage(msg, sender);
```
philipp-classen, line 284: Can we also log failed promises here?

philipp-classen (approved):

chrmod (approved):

PR #919, opened 2022-10-28
Add sync user settings after changing never-consent perfs

smalluban:
It looks like settings are not automatically updated when changed (as I supposed that it does, as there is a Proxy, so it could do it easily). 

Added two sync triggers, when the never-consent popup changes its settings.

### DIFF
--- a/extension-manifest-v2/src/background.js
+++ b/extension-manifest-v2/src/background.js
@@ -567,6 +567,8 @@ function onMessageHandler(request, sender, callback) {
 				conf.autoconsent_interactions = 0;
 			}
 
+			account.saveUserSettings().catch(err => log('Background autoconsent', err));
+
 			return false;
 		}
 		if (name === 'disable') {
@@ -579,6 +581,8 @@ function onMessageHandler(request, sender, callback) {
 				conf.autoconsent_blacklist = [];
 				conf.autoconsent_interactions = 0;
 			}
+
+			account.saveUserSettings().catch(err => log('Background autoconsent', err));
 
 			return false;
 		}


philipp-classen:
Did some testing with Opera with the 8.9.4 production Ghostery (not this PR), logging in and then going to spiegel.de, which triggers Never-Consent. Then the extension repeats making requests to the account API forever (about once every second):

```
GET https://accountapi.ghostery.com/api/v2.1.0/settings/<ACCOUNT_ID>....
```
When I turn off Never-Consent, it stops.

Have stopped debugging at this point, but I get the feeling that the proxy responsible to sync conf has more problems with Never-Consent settings (or the way we make changes to the conf object). Normally, explicit calls to force a sync should not be required, so I assume the problem with endless account IDs is related to the observation that Never-Consent settings are not synced properly.

smalluban:
I saw that request too, but it is unrelated to the PR changes. It is triggered even without the changes.

smalluban:
@philipp-classen Fetching user settings every second (when never-consent popup is opened) is related to this line https://github.com/ghostery/ghostery-extension/blob/fix-mv2-autoconsent-sync/extension-manifest-v2/app/autoconsent/index.js#L81

It gets data for displaying the categories wheel every second. It sends two messages - `getPanelData` and `getCommonModuleData` required to display the information, so... it might be a kind of bug (why user settings are fetched then), but the initiator is right.

smalluban:
@chrmod @philipp-classen I looked again if there is anything that we should do differently, and I think the above changes are fine. Fetching user data is a side-effect of getting tracking information (I would rather ask if we should do it there).

philipp-classen (approved):

PR #934, opened 2022-11-15
Integrate WhoTracks.me reporting

philipp-classen:
Integrates WhoTracks.me reporting.

Depends on https://github.com/ghostery/ghostery-extension/pull/934


extension-manifest-v3/src/background/reporting.js
```diff
@@ -0,0 +1,138 @@
...
+const COLLECTOR_DIRECT_URL = 'https://anonymous-communication.ghostery.net';
+const COLLECTOR_PROXY_URL = COLLECTOR_DIRECT_URL; // current we have no proxy configured
+
+const config = {
+  COLLECTOR_DIRECT_URL,
+  COLLECTOR_PROXY_URL,
+  CONFIG_URL: 'https://api.ghostery.net/api/v1/config',
+  ...platformSpecificSettings(),
+};
+console.debug('reporting config:', config);
```
smalluban, line 62: Clean up `console.debug` statement.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -0,0 +1,138 @@
...
+observe('terms', (terms) => {
+  if (terms) {
+    reporting.init().catch((e) => {
+      console.warn(
+        'Failed to initialize reporting. Leaving the module disabled and continue.',
+        e,
+      );
+    });
+  } else {
+    reporting.unload();
```
smalluban, line 100: I think `terms` cannot be unchecked once they are set to true, so this `else` so if `reporting.unload()` should be called only after `reporting.init()` it is not required here.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -0,0 +1,138 @@
...
+        await reporting.processPendingJobs();
+      }
+    } catch (e) {
+      console.warn('Unexpected error in reporting module:', e);
+    }
+  })();
+});
+
+// for debugging service-workers (TODO: provide a way to control logging)
+self.ghostery = self.ghostery || {};
```
smalluban, line 134: Clean up debugging code.

smalluban (changes requested):
Using an external repository makes it harder to checkout the code and test it locally, as I need to add another origin to git. Please, next time, use a branch within the main repository.

extension-manifest-v3/src/background/reporting.js
philipp-classen: I would prefer to log the settings. I didn't see any logger, so I used console.debug.

extension-manifest-v3/src/background/reporting.js
philipp-classen: This is also intentional to have a way to interact with the objects.

extension-manifest-v3/src/background/reporting.js
philipp-classen: With the current UI, it's not reachable that is correct. I wonder if leaving the "unload" path helps much in terms in maintenance? To me, it is less surprising if it stays in sync with the setting and makes it more obvious that it is active iff terms are accepted.

extension-manifest-v3/src/background/reporting.js
smalluban: I would rather make that code inside of the function (with those with `self` commented below), as this is useless for the user, and if used at all, only by us in the development process. Because of that, it can be only "uncomment", or run if required.

Otherwise, it only creates noise in the logs from the service worker.

philipp-classen: OK, I'll expose the config in "self" and remove the logging.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -0,0 +1,138 @@
...
+  const { url, frameId } = details;
+  if (frameId !== 0 || url === 'about:blank' || url.startsWith('chrome://')) {
+    return;
+  }
+
+  // TODO: in private windows, we should also exit (how to get the information?)
+
+  (async () => {
+    try {
+      const jobRegistered = await reporting.analyzeUrl(url);
```
chrmod, line 113: what happens if reporting is not initialized at this point?

extension-manifest-v3/src/background/reporting.js
smalluban: If you really need that for debugging, I would recommend using `globalThis` instead of `self`, as it is normalized access to the global object regardless of the context. 

However, it introduces a side-effect, and I am not still convinced if it should be run without any condition, or behind the function call.

extension-manifest-v3/src/background/reporting.js
philipp-classen: If it is unloaded, it will ignore it:
https://github.com/whotracksme/webextension-packages/blob/322191e126a34e8521d64e72ef645e40c2033d06/packages/reporting/src/reporting.js#L111

extension-manifest-v3/src/background/reporting.js
philipp-classen: I changed it to globalThis.

Regarding side-effects, it extends a global called "ghostery". I would hope this should not impact other APIs. The use case is to inspect the state when you have a production build and cannot change the code. Or to trigger operations like forcing a pattern update. I'm not aware of a better mechanism then using globals, apart from building extra tooling into the extension and exposing them through the UI.

smalluban: Ok, good point with the argument about the production build. You've bought me :)

smalluban (approved):

extension-manifest-v3/src/background/reporting.js
smalluban: It's more than correct that you only turn it on if the terms are accepted. My concern is only about clean up, which for now will never trigger. However, it might in the future, I would keep it as it is.

PR #946, opened 2022-11-22
Fix: uninstall survey not triggering in Firefox

chrmod:
https://github.com/ghostery/ghostery-extension/pull/820 introduced a regression by which `chrome.runtime.setUninstallUrl` was not called as often.

### DIFF
--- a/extension-manifest-v2/src/background.js
+++ b/extension-manifest-v2/src/background.js
@@ -1379,6 +1379,8 @@ function initializeVersioning() {
  * @return {Promise}
  */
 function initializeGhosteryModules() {
+	metrics.setUninstallUrl();
+
 	if (globals.JUST_UPGRADED) {
 		log('JUST UPGRADED');
 		metrics.ping('upgrade');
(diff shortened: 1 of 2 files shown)


extension-manifest-v2/test/src/Dispatcher.test.js
```diff
@@ -23,7 +23,7 @@ describe('dispatcher tests', () => {
 
 	test('dispatcher is triggered once for conf value not in SYNC_ARRAY', () => {
 		conf.paused_blocking = true;
-		return expect(spy).toHaveBeenCalledTimes(1);
+		return expect(spy).toHaveBeenCalledTimes(2);
```
philipp-classen, line 26: Not a problem, but curious why it is now called twice? From the code change, I wouldn't had expected because it appears to move the "metrics.setUninstallUrl" up.

philipp-classen (approved):

extension-manifest-v2/test/src/Dispatcher.test.js
chrmod: comes from here https://github.com/ghostery/ghostery-extension/pull/946/files#diff-086936c62c02dba1767dca312056acb75a897c13f613da59bb2ff68940df79ffR75 
but this change has no sideeffects as no one else is listening to this event

PR #963, opened 2022-12-16
MV2: Hardening init path

chrmod:
* remove obsolete account migration
* preventing race conditions of account init and saving prefs
* critical path error logging
* ensuring default block rules are applied


extension-manifest-v2/src/background.js
```diff
@@ -1620,42 +1628,67 @@ async function recordUTMs() {
  * Application Initializer
  * Called whenever the browser starts or the extension is
  * installed/updated.
+ *
+ * IMPORTANT: nothing on the intialization path should be able to fail.
+ * 	It is ciritical for the application to enter the `globals.INIT_COMPLETE = true` state.
```
philipp-classen, line 1633:
```suggestion
 * 	It is critical for the application to enter the `globals.INIT_COMPLETE = true` state.
```

extension-manifest-v2/src/background.js
```diff
@@ -1569,21 +1569,31 @@ async function initializeAccount() {
 	}, 5000);
 
 	try {
-		await account.migrate();
-		lastStep = 'migrate';
+		try {
+			await account.getUser();
```
philipp-classen, line 1573: This change is not obvious to me: there was a "migrate" before, which did multiple things including modifications to local storage. As I understood, the migrate functionality is no longer needed.

"getUser" is called for its side-effect, I guess. Maybe change the comment in the exception to something like "expected if the user is not logged in".

philipp-classen (changes requested):

philipp-classen (approved):

Issue #971, opened 2022-12-16
User is logged out after logged in


GRadziejewski:
Description
User is logged out after logged in. 

Steps to Reproduce
1. Click on User icon to open login view
![obraz](https://user-images.githubusercontent.com/40433708/208185445-92424de4-ed77-4e06-aa30-45fa8b8925ee.png)
2. Enter login/password
![obraz](https://user-images.githubusercontent.com/40433708/208185497-b97bc2d9-fea1-42f5-b1b7-dc12bbc67f9f.png)
3. Click "Sing in"
![obraz](https://user-images.githubusercontent.com/40433708/208185565-441625f0-58d3-4cf1-8373-164055c2b97b.png)
4. Reload the webpage and check User icon
![obraz](https://user-images.githubusercontent.com/40433708/208185737-9102fb6b-5b5f-45d7-b933-de680d8906e1.png)

Expected Results:
User should not be logged out after logged in. 

GBE version 8.9.8

GRadziejewski:
Checked and fixed.

Issue #970, opened 2022-12-16
User name is missing after clicking three dots


GRadziejewski:
Description
After logging in, user name is missing when User click three dots.

Steps to Reproduce
1. Click on three dots to open settings
![UserName is missing_8 9 8](https://user-images.githubusercontent.com/40433708/208184907-ec92a198-be43-4239-87a6-5760e9503811.png)

Expected Results:
User name should be visible

GBE version 8.9.8

GRadziejewski:
Checked and fixed.

PR #974, opened 2022-12-19
Fix JWT parsing

chrmod:
Breakage introduced by latest WebPack 5 migration. Build system was incorrectly assuming that global.Buffer is present in the browser.

fix https://github.com/ghostery/ghostery-extension/issues/971 
fix https://github.com/ghostery/ghostery-extension/issues/970

### DIFF
--- a/extension-manifest-v2/src/classes/Account.js
+++ b/extension-manifest-v2/src/classes/Account.js
@@ -84,7 +84,6 @@ class Account {
 			body: data,
 			headers: {
 				'Content-Type': 'application/x-www-form-urlencoded',
-				'Content-Length': Buffer.byteLength(data),
 			},
 			credentials: 'omit',
 		}).then((res) => {
@@ -275,7 +274,6 @@ class Account {
 			body: data,
 			headers: {
 				'Content-Type': 'application/x-www-form-urlencoded',
-				'Content-Length': Buffer.byteLength(data),
 			},
 		}).then((res) => {
 			if (res.status >= 400) {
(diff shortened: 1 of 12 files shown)


extension-manifest-v2/src/classes/Account.js
```diff
@@ -84,7 +84,6 @@ class Account {
 			body: data,
 			headers: {
 				'Content-Type': 'application/x-www-form-urlencoded',
-				'Content-Length': Buffer.byteLength(data),
```
philipp-classen, line 87: "Content-Length" will be still automatically filled by "fetch", right?

chrmod: was checking on the backend side and cannot find a proof that this header is in use.

extension-manifest-v2/src/utils/common.js
```diff
@@ -187,58 +187,18 @@ export function hashCode(str) {
...
-	const payload = JSON.parse(_base64urlDecode(payloadSeg));
-
-	return {
-		header,
-		payload,
-		signature: signatureSeg
-	};
+// source https://stackoverflow.com/a/38552302
+export function parseJwt(token) {
+	const base64Url = token.split('.')[1];
```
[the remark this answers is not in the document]
chrmod: have noticed this as well, but as this is a fix PR, have decided not to introduce any changes here

extension-manifest-v2/src/classes/Account.js
philipp-classen: I think, it should be still present in the request because fetch adds it implicitly and takes the size of the body.

PR #989, opened 2023-01-18
Add account session refresh

smalluban:
* Adds check if the session is valid, when account is fetched
* Adds alarm set to every 30 days to reload the cookies
* Adds more specific name for alarms in the options scope

### DIFF
--- a/extension-manifest-v3/src/store/account.js
+++ b/extension-manifest-v3/src/store/account.js
@@ -10,65 +10,12 @@
  */
 
 import { store } from 'hybrids';
-
-const ACCOUNT_URL = 'https://accountapi.ghostery.com/api/v2.1.0';
-const AUTH_URL = 'https://consumerapi.ghostery.com/api/v2';
-const COOKIE_URL = 'https://ghostery.com';
+import * as api from './utils/api.js';
 
 const ALARM_NAME = 'account:refresh';
 const REFRESH_RATE = 60 * 24 * 30; // 30 days in minutes
 
-async function cookie(name) {
-  // TODO: Add firstPartyDomain support
-  // firstPartyDomain: 'ghostery.com',
-  const cookie = await chrome.cookies.get({ url: COOKIE_URL, name });
-  return cookie ? cookie.value : null;
-}
-
-async function get(url) {
-  const accessToken = await cookie('access_token');
-  const csrfToken = await cookie('csrf_token');
-
-  if (!accessToken || !csrfToken) throw Error('Unauthorized');
-
-  const res = await fetch(url, {
-    headers: {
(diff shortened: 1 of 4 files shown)


chrmod (approved):

extension-manifest-v3/src/store/account.js
```diff
@@ -11,36 +11,57 @@
...
+}
+
+async function get(url, headers) {
+  const csrfToken = await cookie('csrf_token');
+
+  const res = await fetch(url, {
     headers: {
       'Content-Type': 'application/vnd.api+json',
-      'X-CSRF-Token': csrfToken,
+      'X-CSRF-Token': csrfToken || undefined,
```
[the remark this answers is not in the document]
smalluban: No, I thought that setting it to `undefined` won't set it. I'll check.

smalluban:
I checked out Safari, and I didn't know, but it does not support getting `httpOnly` cookies, so I have to make a logic change and save `refresh_token` locally.

extension-manifest-v3/src/store/account.js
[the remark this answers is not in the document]
philipp-classen: Yes, please do. I tried that in the extension:
```
await fetch('https://example.com', { headers: { 'X-CSRF-Token': undefined } })
```

For me, it sets it.

extension-manifest-v3/src/store/account.js
```diff
@@ -11,36 +11,61 @@
...
-    // firstPartyDomain: 'ghostery.com',
-  });
+async function session(account) {
+  const userId = await cookie('user_id');
+  if (!userId) return undefined;
+
+  if (account && !(await cookie('access_token'))) {
+    const refreshToken = await cookie('refresh_token');
+    if (refreshToken) {
+      await fetch(`${AUTH_URL}/refresh_token`, {
```
philipp-classen, line 57: As a side-effect, this call updates cookies (access_token & refresh_token), right? Maybe we should still throw if the request fails (non 2xx).

Side-note: On Manifest V2, we changed the API to return also in the response body (for Firefox). As I understand, that workaround could now also be avoided, since we were forced to use httpOnly=false cookies. (But we would need to change the code there and also wait until old clients go away.)

smalluban: Your comment made me think about setting and clearing cookies, and as we omit credentials (we have to for Safari and upcoming Firefox), there must be logic for processing them. I have started work, but unfortunately, I'll not finish it today. Stay tuned on Monday.

smalluban: Should be ok now.

chrmod (approved):

PR #1023, opened 2023-02-06
MV3 Initial support for Firefox

smalluban:
The goals:

* UI / background process works without obvious throws, error logs, etc
* Working cosmetics and script injection

Changes highlights:

* Store models are used in different contexts, which apparently in Firefox sometimes have limited access to chrome API (especially iframe loaded by content script). This behavior forces us to move out side effects to the background process.
* Minor panel CSS fixes (different popup window sizing computation)
* Refactor the build scripts to support multiple targets (`npm start` runs browsers automatically)

The request filtering is out of the scope of this PR.


extension-manifest-v3/README.md
```diff
@@ -8,23 +8,34 @@ This repository contains a version built with the new `chrome.declarativeNetRequ
 
 First, you need to install JavaScript dependencies and download additional resources (e.g. block lists):
 
-```sh
-$ (cd .. && npm ci)
```
chrmod, line 12: will this work correctly? specifically, will it install dependencies in `libs`?

extension-manifest-v3/scripts/build.js
```diff
@@ -197,3 +214,42 @@ for (const [id, path] of Object.entries(mapPaths(content_scripts))) {
...
+        let settings;
+        switch (argv.target) {
+          case 'safari':
+            exec('xed xcode');
+            return;
+          case 'firefox':
+            settings = {
+              target: 'firefox-desktop',
+              firefoxBinary:
+                '/Applications/Firefox Nightly.app/Contents/MacOS/firefox-bin',
```
chrmod, line 233: can we make it platform independent?

extension-manifest-v3/src/pages/onboarding/index.html
```diff
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width" />
     <meta name="theme-color" content="#000000" />
     <title>Welcome to Ghostery</title>
-    <script src="./index.js" type="module" async></script>
+    <script src="./index.js" type="module"></script>
```
chrmod, line 8: haven't you just added it?

extension-manifest-v3/src/store/options.js
```diff
@@ -58,70 +62,15 @@ const Options = {
...
-        } else {
-          chrome.declarativeNetRequest.updateDynamicRules({
-            removeRuleIds: [1],
-          });
-        }
-      }
+      await chrome.storage.local.set({
+        options:
+          __PLATFORM__ === 'firefox'
+            ? JSON.parse(JSON.stringify(options))
```
chrmod, line 71: please add a comment on why this is needed

chrmod (changes requested):

extension-manifest-v3/README.md
smalluban: Yes, we have only one `node_modules` in the root path, so it goes up and install everything, which is needed for the all workspaces.

extension-manifest-v3/scripts/build.js
smalluban: I thought about it, but it will be hard (would have to use some kind of `.env` file with paths), as there is no one place where you put your binaries. I assumed that this is mainly for us, which uses macOS, and we can keep browsers in predefined locations. I updated the README to highlight the problem. 

After all, the user can use `npm run build` and open the browser manually.

extension-manifest-v3/src/store/options.js
```diff
@@ -138,140 +87,91 @@ const Options = {
...
-      DNR_RULES_LIST.forEach((rule) => {
-        const enabled = options.dnrRules[rule];
-        if (enabledRulesetIds.includes(rule) !== enabled) {
-          (enabled ? enableRulesetIds : disableRulesetIds).push(rule);
+    observe: (_, options, prevOptions) => {
+      observers.forEach((fn) => {
+        try {
+          fn(options, prevOptions);
+        } catch (e) {
+          console.error(`Error while observing options: `, e);
```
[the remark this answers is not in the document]
smalluban: No they are not, as only content scripts share the same console with the main frame. Those comments will only show up in our full pages and background/service worker.

extension-manifest-v3/src/utils/shims.js
```diff
@@ -0,0 +1,47 @@
...
+    onAlarm: {
+      addListener(fn) {
+        listeners.add(fn);
+      },
+    },
+  };
+}
+
+if (__PLATFORM__ === 'firefox') {
+  window.chrome = Object.assign(window.browser, {
```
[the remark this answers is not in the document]
smalluban: I think, we can't. Shims work just in opposite direction. `libs` are used by MV2, which targets different platforms, and I am not sure if everywhere `chrome` API is promise-based. There is a `browser` from the polyfill used.

extension-manifest-v3/README.md
smalluban: You are right, I had to have already installed other packages. It must be done from the root.

extension-manifest-v3/src/pages/onboarding/index.html
smalluban: Leftover, I was checking if this was a couse of some Firefox problems, but it looks that it can be `async`. I'll rollback.

extension-manifest-v3/scripts/build.js
chrmod: maybe we can do something like this https://github.com/chrmod/puppeteer-webextension-test/blob/main/index.js#L25

chrmod: I think that web-ext has a way to tell it that you want to run nightly and it will look for it in commonly known path. With command line we can say `--firefox=nightly`

extension-manifest-v3/src/store/options.js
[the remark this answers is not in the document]
chrmod: in Firefox, extension logs are shown in Browser Toolbox and on stdout

extension-manifest-v3/scripts/build.js
smalluban: I read a CLI help messages, even a source code, look for answers on google, stack... As far I can tell, if you want to use another version of Firefox, you must provide a path to the binary.

extension-manifest-v3/src/store/options.js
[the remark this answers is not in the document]
smalluban: This PR does not "touch" logs, so if you think this is something that we should address, please create for it separate issue.

chrmod (approved):

PR #1035, opened 2023-02-19
MV3: adblocker - ensure correct state on startup

chrmod:
During the background page initialization there was a race condition preventing adblocker engines to be marked as enabled. In result, adblocker cosmetic filters never run in the version 10.0.0.


philipp-classen (approved):

smalluban:
@chrmod The bug in else place, L171 in options.js (can't make a comment on not changed line in PR). The async was there on purpose, but there is missing await. 

As this solution fixes adblocker code, observe still should wait for the resolved options, as it is used in other places. Then, I think directly calling resolve in adblocker wan't be necessary.

chrmod:
> can leave it as it is for now, when I'll be back i check it out.

The very same problems happens with the `reporting` https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v3/src/background/reporting.js#L217 - but there the `observe` is called in the top level so we cannot await. I'm thinking about force calling observers at the end of `Store.get` https://github.com/ghostery/ghostery-extension/blob/eeeb38350c82779a7b221e24939a98cd735f4c27/extension-manifest-v3/src/store/options.js#L89 - what you think?

smalluban:
If you would look at the last revision before my change, the observe method is called the fn at the start. As I moved calling observers to `observe` method, I wanted to avoid double execution. The only problem now is with not waiting until that fn is called (is it will be, the store.resolve forces that). 
I think for now the solution might be to take the vale from L171, and call the fn at the end of that method, and call observers in `observe` method in the definition only if lastValue is not thanks.

PR #1038, opened 2023-02-22
MV3: prevent starup races with Options

chrmod:
On physical iOS devices store of Options model was connected before all observers got registered. This broke those modules that relied on the `observe` callback. 

Breakage was introduced here https://github.com/ghostery/ghostery-extension/pull/1023/files#diff-ab787eda050f8d097068bea78c9043e942c4480fdb99acf1f1c0213b0bbd859dL270

Workaround for the adblocker startup https://github.com/ghostery/ghostery-extension/pull/1035/files#diff-df2e21bb6266c3df460f6af7886b15496d496b3bd46182750a55addf0a1a7a89R31 is no longer needed.

### DIFF
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -156,7 +156,7 @@ export async function observe(property, fn) {
       value = options[property];
 
       try {
-        return fn(value, prevValue);
+        return await fn(value, prevValue);
       } catch (e) {
         console.error(`Error while observing options: `, e);
       }
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -87,8 +87,8 @@ const Options = {
       }
       return options;
     },
-    observe: (_, options, prevOptions) => {
-      observers.forEach((fn) => fn(options, prevOptions));
+    observe: (_, options) => {
+      observers.forEach((fn) => fn(options));
     },
   },
 };
@@ -148,15 +148,15 @@ async function migrateFromMV2() {
   }
 }
 
-export function observe(property, fn) {
+export async function observe(property, fn) {
   let value;
-  const wrapper = (options) => {
+  const wrapper = async (options) => {
     if (value === undefined || options[property] !== value) {
       const prevValue = value;
       value = options[property];
 
       try {
-        fn(value, prevValue);
+        return fn(value, prevValue);
       } catch (e) {
         console.error(`Error while observing options: `, e);
       }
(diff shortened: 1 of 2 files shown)


extension-manifest-v3/src/store/options.js
```diff
@@ -168,7 +163,9 @@ export function observe(property, fn) {
 
   observers.add(wrapper);
 
-  store.resolve(Options);
+  // let observer know of the option value
+  // in case when registered after the store.connect
+  store.resolve(Options).then(wrapper);
```
philipp-classen, line 168:
```suggestion
  store.resolve(Options).then(wrapper).catch(console.error);
```

smalluban:
@chrmod the change looks fine, but observe method mist be async, and there must be await in the line with store.resolve. Otherwise, for example setup in adblocker won't wait until the engine values are correctly set.

extension-manifest-v3/src/store/options.js
```diff
@@ -162,13 +155,19 @@ export function observe(property, fn) {
...
+        fn(value, prevValue);
+      } catch (e) {
+        console.error(`Error while observing options: `, e);
+      }
     }
   };
 
   observers.add(wrapper);
 
-  store.resolve(Options);
```
smalluban, line 171: Add await and async in the function definition.

smalluban (changes requested):

extension-manifest-v3/src/store/options.js
chrmod: `observe` is called in top level most of the time where it cannot be awaited.

this is also why the we trigger callback (wrapper) after the `store.resolve` - so that observers are always notified, regardless on when the `observe` was called.

chrmod: but I see the point with the adblocker startup. make sense

extension-manifest-v3/src/store/options.js
```diff
@@ -148,15 +148,15 @@ async function migrateFromMV2() {
...
   let value;
-  const wrapper = (options) => {
+  const wrapper = async (options) => {
     if (value === undefined || options[property] !== value) {
       const prevValue = value;
       value = options[property];
 
       try {
-        fn(value, prevValue);
+        return fn(value, prevValue);
```
philipp-classen, line 159: return await fn(value, prevValue);
  
  otherwise, catch will not have an effect

smalluban (approved):

philipp-classen (approved):

PR #1044, opened 2023-02-27
Use TrackerDB instead of BugsDB

chrmod:


extension-manifest-v3/scripts/download-trackerdb.js
```diff
@@ -0,0 +1,22 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+import { writeFileSync } from 'fs';
+import fetch from 'node-fetch';
+import path from 'path';
+
+const trackerDB = await fetch(
+  'https://github.com/ghostery/trackerdb/releases/latest/download/trackerdb.engine',
+).then((res) => res.arrayBuffer());
```
philipp-classen, line 17: Should throw on non-2xx. I think, the current code would use the bytes from the body of a 4xx, 5xx response.

extension-manifest-v3/src/background/tab-stats.js
```diff
@@ -55,14 +55,16 @@ async function updateTabStats(msg, sender) {
...
-      if (tracker) {
-        stats.trackers.push(tracker);
-      }
-    });
+  const newUrls = msg.urls.filter(
+    (url) => !stats.trackers.some((t) => t.url === url),
+  );
+
+  for (const url of newUrls) {
+    const pattern = await getMetadata(url, stats.sourceUrl);
```
philipp-classen, line 63: Would it make sense to run those requests in parallel?

extension-manifest-v3/scripts/build.js
```diff
@@ -145,6 +145,13 @@ if (manifest.declarative_net_request?.rule_resources) {
   });
 }
 
+// copy TrackerDB
```
smalluban, line 148: This is not required, as whole `src/assets` path is copied in L73. I think it might even support subpaths. It takes a list of paths from L13.

extension-manifest-v3/scripts/build.js
```diff
@@ -178,6 +185,7 @@ const buildPromise = build({
         entryFileNames: '[name].js',
         assetFileNames: 'assets/[name].[ext]',
       },
+      plugins: [],
```
smalluban, line 188: Is this required?

smalluban (changes requested):

extension-manifest-v3/src/background/tab-stats.js
chrmod: this is only for the startup

smalluban (approved):

philipp-classen (approved):

PR #1047, opened 2023-03-02
Fallback when trackerDB misses pattern's organization

smalluban:
The previous bugs.js implementation includes a case when an organization (company) was missing - as it is required in the panel to show the dialog with a list of requests.

Test case: `https://onet.pl` - trackers from 'onet' have pattern, but it does not contain organization

### DIFF
--- a/extension-manifest-v3/src/background/utils/trackerdb.js
+++ b/extension-manifest-v3/src/background/utils/trackerdb.js
@@ -43,14 +43,20 @@ export async function getMetadata(url, sourceUrl) {
   return {
     name: pattern.name,
     category: category.key,
-    company: {
-      id: organization.key,
-      name: organization.name,
-      description: organization.description,
-      website: organization.website_url,
-      contact: organization.privacy_contact,
-      privacyPolicy: organization.privacy_policy_url,
-    },
+    company: organization
+      ? {
+          id: pattern.key,
+          name: organization.name,
+          description: organization.description,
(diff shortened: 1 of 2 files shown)


chrmod (dismissed):

extension-manifest-v3/src/background/utils/trackerdb.js
```diff
@@ -43,14 +43,20 @@ export async function getMetadata(url, sourceUrl) {
...
-      website: organization.website_url,
-      contact: organization.privacy_contact,
-      privacyPolicy: organization.privacy_policy_url,
-    },
+    company: organization
+      ? {
+          id: organization.key,
+          name: organization.name,
+          description: organization.description,
+          website: organization.website_url,
```
chrmod, line 51: maybe we should use `pattern.website_url` here if present

extension-manifest-v3/src/background/utils/trackerdb.js
```diff
@@ -43,14 +43,20 @@ export async function getMetadata(url, sourceUrl) {
...
-      name: organization.name,
-      description: organization.description,
-      website: organization.website_url,
-      contact: organization.privacy_contact,
-      privacyPolicy: organization.privacy_policy_url,
-    },
+    company: organization
+      ? {
+          id: organization.key,
+          name: organization.name,
```
chrmod, line 49: maybe we should use `pattern.name` here if present

chrmod (approved):

PR #1051, opened 2023-03-07
MV2: replace BugDB with TrackerDB

chrmod:
TODO:
* [x] missing icon for category `Extensions`
   Screenshot 2023-03-07 at 17 30 52
* [x] missing translation `Telemetry`
* [x] `Unknown` category handling
  Screenshot 2023-03-07 at 17 30 55
* handle TrackerDB updates (on extension upgrades) 
  * [x] update `selected_app_ids` 
  * [x] how to deal with new trackers (without `ghostery_id`)
* TrackerDB detection and blocking
  * [x] decide if domain matching is needed
  * [x] consider disabling of blocking (as is provided by Ad-blocking already)


extension-manifest-v2/src/classes/BugDb.js
```diff
@@ -10,246 +10,85 @@
...
-			// Bias in favor of blocking if the category will be
-			// blocked by default in a new installtion. This is
-			// mostly relevant if new categories were introduced.
-			majority[category] = 0.5;
-		});
-		Object.entries(old_bugs.apps).forEach(([app_id, { cat: category }]) => {
-			const vote = selected_app_ids[app_id] ? 1 : -1;
-			majority[category] = (majority[category] || 0) + vote;
-		});
+		this.db.bugs = engine.metadata.getPatterns().reduce((all, pattern) => ({
```
philipp-classen, line 82: The number of patterns isn't small, right? Be careful with the reduce-patterns. I think, it results in O(n²) complexity.

chrmod: fixed

smalluban (dismissed):
I did not notice obvious bugs. However, are we fine with hardcoding a trackerdb in MV2, where before the bugs where fetched dynamically?

philipp-classen:
> I did not notice obvious bugs. However, are we fine with hardcoding a trackerdb in MV2, where before the bugs where fetched dynamically?

Valid point (though I don't know what the implications of not updating will be).

As I understand, trackerdb can be seen as a link between community patterns and curated meta data. Meta data doesn't change frequently in my experience; it might be OK to update only on extension releases (even though from a maintainer perspective, it is preferable to be able to push fixes without releases).

In contrast to meta data, community list patterns (and trackers) change frequently. That means, it will naturally become more out-of-sync over time. Whether that will create problems or not, is hard to tell from my side. In any case, it would be desirable if the extension fails gracefully if there are inconsistencies. Maybe that mitigates the problem.

extension-manifest-v2/manifest.json
```diff
@@ -3,8 +3,8 @@
 	"author": "Ghostery",
 	"name": "__MSG_name__",
 	"short_name": "Ghostery",
-	"version": "8.9.14",
-	"version_name": "8.9.14",
+	"version": "8.10.0",
```
smalluban, line 6: I would not bump the version in feature PR, it should be done in a separate PR with its own commit message (and then you have a clean commit, which you can tag).

package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "glossary-converter": "node tools/glossary-converter.js",
-    "postinstall-mv2": "cd extension-manifest-v2 && npm run licenses",
+    "postinstall-mv2": "cd extension-manifest-v2 && npm run licenses && npm run download-trackerdb",
```
smalluban, line 8: Is there a reason why packages postinstall scripts are not in the packages, but here? It was before like `npm postinstall -ws ...`.

ui/src/utils/categories.js
```diff
@@ -16,7 +16,7 @@ const colors = {
   essential: '#FC9734',
   hosting: '#8459A5',
   customer_interaction: '#EF671E',
-  unknown: '#E8E8E8',
+  unidentified: '#E8E8E8',
```
smalluban, line 19: if we add a new category name, we must update locales in MV3 too.

smalluban (changes requested):

package.json
chrmod: Running `postinstall` from packages is not a supported usecase for npm workspaces. it started to fail for me with latest npm upgrade (when we've moved to node 19) - 

This approach is more predictable

extension-manifest-v2/_locales/de/messages.json
```diff
@@ -537,17 +537,14 @@
   },
   "category_telemetry_desc": {
     "description": "User-facing description for the Telemetry tag category.",
-    "message": "Telemetrie"
+    "message": "category_advertising_desc"
```
philipp-classen, line 540: Is this change intentional?

extension-manifest-v2/src/background.js
```diff
@@ -100,24 +100,6 @@ function updateDBs() {
...
-				log('CHECK LIBRARY VERSION CALLED', result);
-				if (result.success) {
-					const nowTime = Date.now();
-					conf.bugs_last_checked = nowTime;
-					if (result.updated) {
-						log('BUGS LAST UPDATED UPDATED', new Date());
-						conf.bugs_last_updated = nowTime;
-					}
-				}
-				resolve({
```
philipp-classen, line 113: Won't removing that "resolve" leave the Promise that "updateDBs" returns dangling?

extension-manifest-v2/src/classes/BugDb.js
```diff
@@ -10,246 +10,128 @@
...
-			categories[category].trackers.push({
-				id: appId,
-				name: db.apps[appId].name,
-				description: '',
-				blocked,
-				shouldShow: true,
-				catId: category,
-				trackerID: db.apps[appId].trackerID,
-			});
+	async init(just_upgraded) {
```
philipp-classen, line 46: Is it possible that this function gets called multiple times in parallel? If so, the engine might get loaded multiple times. (Which would not be a correctness problem, but would be wasteful.)

package.json
smalluban: Ok, understand, but it ties the root with package internals... I get the reason but still don't like the solution.

smalluban (approved):
My requested changes added.

extension-manifest-v2/src/classes/BugDb.js
chrmod: the only place it is being called is during background startup

extension-manifest-v2/_locales/de/messages.json
chrmod: nope

extension-manifest-v2/src/background.js
chrmod: good catch

extension-manifest-v2/src/classes/BugDb.js
```diff
@@ -10,246 +10,135 @@
...
-							selected_app_ids[app_id] = 1;
-						});
-
-						// (triggers a persist)
-						conf.selected_app_ids = selected_app_ids;
-					}
-					db.JUST_UPDATED_WITH_NEW_TRACKERS = true;
+		/* block new trackers if in mostly blocked category */
+		if (just_upgraded) {
+			Object.values(categoriesMeta).forEach((meta) => {
```
philipp-classen, line 112: Shouldn't we preserve the existing logic (_scanForNewTrackersToBlock) to take into account which categories are enabled/disabled by default?

A use case is when a new category is added, or if the extension for some reason started with an empty set of trackers in one category. As I understand the new code, the new trackers would be disabled now.

philipp-classen (approved):

Issue #1046, opened 2023-03-02
100% CPU usage on Raspberry Pi OS


jimis:
I'm running the default (as installed via apt) Firefox-ESR browser on Raspberry Pi OS.
My profile is new, but it has synced many bookmarks and tabs from Firefox Sync. However I have no open tabs on this device, besides the one I'm writing in now.

I have installed only one extension, Ghostery, in order to make browsing snappier on this slow device.

What I notice is that the `WebExtensions` process of Firefox consumes constantly 100% of one core, and the browsing feels sluggish. CPU usage doesn't seem to go down even after 30min or so. It only goes down when I disable Ghostery.

### Versions

+ Ghostery: 8.9.8
+ Browser: Firefox 102.8.0esr (32-bit armhf)
+ OS: Raspberry Pi OS (32-bit) running on Raspberry Pi 400

jimis:
Is there a way to take a stacktrace of Ghostery, maybe using Firefox developer tools or javascript console? It might show in what loop it's stuck.

jimis:
I followed this guide [1] and managed to get some info: stopping the secondary thread `worker.asmjs.bundle.js` releases the CPU. The "copy stacktrace" selection results only in this:
```
onmessage (moz-extension://3cec0a9c-094e-4067-830e-7494d6659c7d/common/hpnv2/worker.asmjs.bundle.js#135)
```

[1] https://extensionworkshop.com/documentation/develop/debugging/#debugging-content-scripts

philipp-classen:
The logs are showing WhoTracks.me messages being sent, but the signing step is slow (the hpnv2 protocol is described in https://0x65.dev/blog/2019-12-04/human-web-proxy-network-hpn.html). It should normally take in the order of 10 ms, but according to the logs is taking around one minute.

The crypto operation (https://github.com/whotracksme/anonymous-credentials) should be using WebAssembly, but that you are seeing "worker.asmjs.bundle.js" may indicate that it is using instead the fallback to asm.js (https://en.wikipedia.org/wiki/Asm.js). And apparently, the JavaScript engine doesn't even apply asm.js optimizations but runs it as standard JavaScript, which is super slow.

Do you know if the `javascript.options.wasm` and `javascript.options.asmjs` options are enabled (in about:config)? It should be enabled by default, but maybe not on Raspberry Pi. We'll also try to reproduce.

jimis:
[...]

PR #1057, opened 2023-03-09
MV2: WTM: use lighter transport when no wasm

chrmod:
fix https://github.com/ghostery/ghostery-extension/issues/1046

### DIFF
--- a/extension-manifest-v2/src/classes/Common.js
+++ b/extension-manifest-v2/src/classes/Common.js
@@ -28,6 +28,7 @@ if (!navigator.userAgent.includes('Firefox')) {
 }
 
 const DEFAULT_ADBLOCKER_MODE = 2; // 2 == Ads + Trackers + Annoyances
+const HAS_WASM = typeof WebAssembly === 'object' && typeof WebAssembly.instantiate === 'function';
 const IS_ANDROID = getBrowserInfo.isAndroid();
 
 COMMON.config.baseURL = '/common/';
@@ -84,10 +85,10 @@ common.load = async () => {
 
 	setPref('modules.anolysis.enabled', false);
 
-	if (!IS_ANDROID) {
+	if (!IS_ANDROID && HAS_WASM) {
 		setPref('modules.human-web.enabled', conf.enable_human_web);
 		setPref('modules.hpnv2.enabled', conf.enable_human_web);
-	} else {
+	} else if (IS_ANDROID) {
 		setPref('modules.human-web-lite.enabled', conf.enable_human_web);
 		setPref('modules.hpn-lite.enabled', conf.enable_human_web);
 	}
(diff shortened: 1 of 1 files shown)


extension-manifest-v2/src/classes/Common.js
```diff
@@ -84,7 +85,7 @@ common.load = async () => {
 
 	setPref('modules.anolysis.enabled', false);
 
-	if (!IS_ANDROID) {
+	if (!IS_ANDROID && HAS_WASM) {
```
philipp-classen, line 88: Found a second place where we check for it:

https://github.com/ghostery/ghostery-extension/blob/33452f0bfba750678fabae0630c870c81aa2b848/extension-manifest-v2/src/classes/Common.js#L128

```
export const setWhotracksmeState = async (enabled) => {
	if (!IS_ANDROID) {
		await setModuleState('hpnv2', enabled);
		await setModuleState('human-web', enabled);
	} else {
		await setModuleState('hpn-lite', enabled);
		await setModuleState('human-web-lite', enabled);
	}
};
```

Perhaps we can unify the logic.

philipp-classen (changes requested):

extension-manifest-v2/src/classes/Common.js
```diff
@@ -84,7 +85,9 @@ common.load = async () => {
 
 	setPref('modules.anolysis.enabled', false);
 
-	if (!IS_ANDROID) {
+	if (!HAS_WASM) {
+		// Do NOT enable WTM reporting as e2e encryption is relatively CPU heavy
```
philipp-classen, line 89: e2e encryption is misleading, since it is about signing (or more precisely direct anonymous attestation).

Thinking about it, I would recommend to move the HAS_WASM prerequisite only around hpnv2+human-web. human-web-lite and hpn-lite don't use wasm.

Conceptually:

```
if (ANDROID) {
  		setPref('modules.human-web-lite.enabled', conf.enable_human_web);
		setPref('modules.hpn-lite.enabled', conf.enable_human_web);
} else if (HAS_WASM) {
	        setPref('modules.human-web.enabled', conf.enable_human_web);
		setPref('modules.hpnv2.enabled', conf.enable_human_web);
}
```

philipp-classen (changes requested):

extension-manifest-v2/src/classes/Common.js
philipp-classen, line 88: Maybe
```
if (IS_ANDROID) {
		await setModuleState('hpn-lite', enabled);
		await setModuleState('human-web-lite', enabled);
} else if (HAS_WASM) {
		await setModuleState('hpnv2', enabled);
		await setModuleState('human-web', enabled);
}
```

Rationale: only hpnv2 depends on wasm. And common/human-web-lite de-facto is a pure Android implementation.

So: if Android -> human-web-lite
otherwise, use human-web+hpnv2 unless the lack of wasm introduce a performance bug in hpnv2

PR #1062, opened 2023-03-10
Add applinks support for contribution link

### LINKED ISSUE #1020: Become a contributor should open an native app
We should outsource this, as it requires specialized iOS development skills

smalluban:
Fixes #1020

* Followed [Apple documentation](https://developer.apple.com/documentation/Xcode/supporting-associated-domains?language=objc) to setup associated domain on ghostery.com
* Configured entitlements in XCode 

It looks like everything is configured correctly, but I can't make it work. The domain can be verified here: https://app-site-association.cdn-apple.com/a/v1/www.ghostery.com

To open URL on iOS simulator you can use command:
```
xcrun simctl openurl booted https://www.ghostery.com/become-a-contributor
```

### DIFF
--- a/extension-manifest-v3/src/pages/panel/views/home.js
+++ b/extension-manifest-v3/src/pages/panel/views/home.js
@@ -33,15 +33,13 @@ const NOTIFICATIONS = [
         : 'https://www.ghostery.com/support?utm_source=gbe',
     action: 'Get help',
   },
-  __PLATFORM__ !== 'safari'
-    ? {
-        icon: 'heart',
-        type: '',
-        text: 'Hey, do you enjoy Ghostery and want to support our work?',
-        url: 'https://www.ghostery.com/become-a-contributor?utm_source=gbe',
-        action: 'Become a Contributor',
-      }
-    : null,
+  {
+    icon: 'heart',
+    type: '',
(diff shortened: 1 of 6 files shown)


chrmod:
@smalluban looks like https://www.ghostery.com/.well-known/apple-app-site-association was incorrect. I've made a fix, lets try again.

smalluban:
@chrmod I've checked the app links, and your change seems to help. But unfortunately, the effect is probably not what we expected. The page opens anyway, but there is a "bar" at the top informing you that you can open it in the application. Maybe linking to the account would change something :man-shrugging:

But on the SwiftUI side, I give up. This is extremely unintuitive, and I have no idea how to refer to a view to fire a method on it. I was looking for a solution, but of course, it's kind of wild. Unless it results from our implementation because there is no normal routing there.

Here you have a place to make a call: https://github.com/ghostery/ghostery-extension/blob/8a390c9686298e340328f6eb0a392b07dba218dd/extension-manifest-v3/xcode/iOS%20(App)/GhosteryApp.swift

smalluban:
This is unlikely to be implemented at all. Closing for now.

chrmod:
I would actually prefer to figure it out. Maybe it will work differently under the new apple developer account.

PR #1067, opened 2023-03-14
Sync: remove deprecated settings

smalluban:


extension-manifest-v2/src/classes/Globals.js
```diff
@@ -108,7 +108,6 @@ class Globals {
 			'enable_click2play',
 			'enable_click2play_social',
 			'enable_wtm_serp_report',
-			'enable_human_web',
 			'enable_metrics',
 			'enable_abtests',
 			'enable_smart_block',
```
chrmod, line 113: lets remove enable_smart_blocking - it is no longer in use

chrmod (approved):

PR #1071, opened 2023-03-16
Clean up session & reporting

smalluban:
* Minor clean up of the reporting listener, to avoid computation when reporting is disabled
* Use JWT data to set user information instead of fetching them from end-point


extension-manifest-v3/src/utils/api.js
```diff
@@ -9,71 +9,71 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
-import * as cookies from './cookies.js';
+import jwtDecode from 'jwt-decode';
```
chrmod, line 12: is this needed? if we won't check the signatures decoding the JWT should be fairly simple

PR #1074, opened 2023-03-16
Show the onboarding reminder once a day, not only on startup.

philipp-classen:
Background: if Ghostery is installed, but the onboarding is not finished, all functionality is disabled. Since it is easy to overlook, show a reminder from time to time, since it is probably unintentional to have it installed but not activated.

(Note: this PR is for testing, but there are drawbacks: Ghostery suddenly makes the browser take focus, which IMHO is unexpected and annoying.)

### DIFF
--- a/extension-manifest-v2/src/background.js
+++ b/extension-manifest-v2/src/background.js
@@ -79,7 +79,8 @@ const IS_EDGE = getBrowserInfo.isEdge();
 const IS_FIREFOX = getBrowserInfo.isFirefox();
 const IS_ANDROID = getBrowserInfo.isAndroid();
 const VERSION_CHECK_URL = `${CDN_BASE_URL}/update/v4.1/versions.json`;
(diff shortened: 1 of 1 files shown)


philipp-classen:
Leaving it as a draft. To see the unintended effect of grabbing focus, you can reduce the interval from one day to one minute:
```
--		await tryOpenOnboarding();
++		await tryOpenOnboarding({ period: ONE_MINUTE_MSEC });
  		setInterval(() => {
  			log('run onboarding reminder (only if the extension is installed but inactive)...');
--			tryOpenOnboarding().catch(alwaysLog);
--		}, ONE_DAY_MSEC + ONE_MINUTE_MSEC);
++			tryOpenOnboarding({ period: ONE_MINUTE_MSEC }).catch(alwaysLog);
++		}, ONE_MINUTE_MSEC + ONE_MINUTE_MSEC);
```

PR #1075, opened 2023-03-16
Onboarding telemetry

chrmod:


extension-manifest-v3/src/pages/onboarding/index.js
```diff
@@ -37,5 +41,12 @@ define({
     ></ui-onboarding>`,
 });
 
-// Update options showAt timestamp
-store.set(Options, { onboarding: { shownAt: Date.now() } });
+store.ready(Options).then(() => {
```
smalluban, line 44: It is wrong, you are not using `store.ready()` with the definition, but the model instance. Use `store.resolve(Options)` if you need to read some options before updating.

extension-manifest-v3/src/background/telemetry.js
```diff
@@ -59,6 +59,10 @@ const getConf = async () => {
     enable_ad_block: Object.values(options.engines).some((enabled) => enabled),
     enable_human_web: options.terms,
     install_date,
+    setup_complete: options.onboarding.done && options.terms,
+    setup_skip: options.onboarding.done && options.terms,
```
smalluban, line 63: It should be `!options.terms`

smalluban (changes requested):

extension-manifest-v2/src/background.js
```diff
@@ -133,6 +133,7 @@ async function tryOpenOnboarding({ force = false, period = ONE_DAY_MSEC } = {})
 	const now = Date.now();
 	if (!conf.setup_timestamp || ((now - conf.setup_timestamp) > period)) {
 		conf.setup_timestamp = now;
+		conf.setup_shown += 1;
```
smalluban, line 136: I would update that value after the tab is created/focused, not before that action.

smalluban (changes requested):

smalluban (changes requested):

PR #1078, opened 2023-03-20
Only run Never Consent on HTML pages (not XML).

### LINKED ISSUE #983: XMLs are not displayed properly in Firefox
When Ghostery is enabled, XML files are not displayed correctly. It affects Firefox 108, but not Chrome.

To reproduce, open https://whotracks.me/blog/feed.xml

xml-display-bug

* Left: Firefox without Ghostery
* Middle: Firefox with Ghostery 8.9.8 (Ghostery breaks the styling)
* Right: Chromium with Ghostery

philipp-classen:
Never Consent mistakenly tries to inject its popup in non-HTML pages.

This patch adds a guard to the injection in the content-script:
```
if (document.contentType === 'text/html') {
  ...
}
 ```

fixes #983

### DIFF
--- a/extension-manifest-v2/app/content-scripts/autoconsent.js
+++ b/extension-manifest-v2/app/content-scripts/autoconsent.js
@@ -12,26 +12,28 @@
 import AutoConsent from '@duckduckgo/autoconsent';
 import { showIframe } from '@ghostery/ui/autoconsent/iframe';
 
-const consent = new AutoConsent(msg => chrome.runtime.sendMessage(
-	{ ...msg, action: 'autoconsent' },
-));
+if (document.contentType === 'text/html') {
+	const consent = new AutoConsent(msg => chrome.runtime.sendMessage(
+		{ ...msg, action: 'autoconsent' },
+	));
 
-let shownIframe = false;
-chrome.runtime.onMessage.addListener((msg) => {
-	if (msg.action === 'autoconsent') {
-		if (msg.type === 'openIframe') {
-			if (shownIframe) return false;
+	let shownIframe = false;
+	chrome.runtime.onMessage.addListener((msg) => {
+		if (msg.action === 'autoconsent') {
+			if (msg.type === 'openIframe') {
+				if (shownIframe) return false;
 
-			showIframe(chrome.runtime.getURL(
-				`app/templates/autoconsent.html?host=${encodeURIComponent(msg.domain)}&default=${msg.defaultForAll ? 'all' : ''}`
-			));
-			shownIframe = true;
+				showIframe(chrome.runtime.getURL(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: extension-manifest-v2/app/content-scripts/autoconsent.js [hunk 1]
```diff
@@ -12,26 +12,28 @@
 import AutoConsent from '@duckduckgo/autoconsent';
 import { showIframe } from '@ghostery/ui/autoconsent/iframe';
 
-const consent = new AutoConsent(msg => chrome.runtime.sendMessage(
-	{ ...msg, action: 'autoconsent' },
-));
+if (document.contentType === 'text/html') {
+	const consent = new AutoConsent(msg => chrome.runtime.sendMessage(
+		{ ...msg, action: 'autoconsent' },
```
chrmod, line 17: how about adding the condition here? diff will be much smaller and the effect the same

chrmod (changes requested):
please introduce the same change in MV3

philipp-classen:
> please introduce the same change in MV3

The PR has already changes for both MV2 and MV3.

extension-manifest-v2/app/content-scripts/autoconsent.js
philipp-classen: The diff is effectively two lines (adding one if statement). I don't know how that could be made smaller (or if there is any benefit in moving the if only before the sending?).

ghostwords:
FYI, [here is the content script check](https://github.com/EFForg/privacybadger/blob/04901bc1364de776c0a795eefb003a4290da12a6/src/js/contentscripts/dnt.js#L20-L27) that Privacy Badger uses.

smalluban (approved):

philipp-classen:
> FYI, [here is the content script check](https://github.com/EFForg/privacybadger/blob/04901bc1364de776c0a795eefb003a4290da12a6/src/js/contentscripts/dnt.js#L20-L27) that Privacy Badger uses.

Interesting. The difference is better support for XHTML pages (that may use "application/xhtml+xml"). I think, in the context of Never Consent limiting to HTML is probably correct; but we should keep it in mind if Never-Consent stops working on pages that we want to support.

extension-manifest-v2/app/content-scripts/autoconsent.js
chrmod: indentation makes the diff big
Screenshot 2023-03-21 at 14 23 33

philipp-classen: Github has an option to show diffs without white spaces if that is the concern. It's a small code change.

PR #1079, opened 2023-03-21
POC: Contextual Onboarding

chrmod:


extension-manifest-v3/src/background/index.js
```diff
@@ -12,6 +12,7 @@
 import '../utils/shims.js';
 
 import './onboarding.js';
+import './contextual-conboarding';
```
smalluban, line 15: no extension

extension-manifest-v3/src/content_scripts/contextual-onboarding.js
```diff
@@ -0,0 +1,19 @@
+import { showIframe } from '@ghostery/ui/contextual-onboarding/iframe';
```
smalluban, line 1: No copyrights

extension-manifest-v3/src/background/contextual-conboarding.js
```diff
@@ -0,0 +1,26 @@
...
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+function openIframe(tabId) {
```
smalluban, line 12: I would use `src/background/onboarding.js`, and add this there. It is the same feature, but only another representation - no need to a new file here.

extension-manifest-v3/src/manifest.chromium.json
```diff
@@ -105,6 +105,14 @@
       ],
       "all_frames": true
     },
+    {
+      "matches": ["*://*/*"],
+      "run_at": "document_start",
```
smalluban, line 110: It might be `document_idle` as it does not have to be "super fast".

extension-manifest-v3/src/store/stats.js
```diff
@@ -59,15 +59,9 @@ const Stats = {
...
-      const currentTab =
-        chrome.tabs &&
-        (await chrome.tabs.query({ active: true, currentWindow: true }))[0];
-
-      const tabId =
-        currentTab?.id ||
-        (await chrome.runtime.sendMessage({
-          action: 'getCurrentTabId',
-        }));
+      const tabId = await chrome.runtime.sendMessage({
```
smalluban, line 62: The `getCurrentTabId` was here only for the fallback for the Firefox iframe limitation. Is using the `chrome.tabs` API not preferred where possible?

smalluban (changes requested):
Generally, I think we could use `ui/onboarding` module, and extend it with a new component - so you would avoid a lot of copy & paste.

ui/src/modules/onboarding/iframe.js
```diff
@@ -0,0 +1,88 @@
...
+export function showIframe(url) {
+  const wrapper = document.createElement('div');
+  const shadowRoot = wrapper.attachShadow({ mode: 'closed' });
+  const template = document.createElement('template');
+
+  template.innerHTML = `
+    <iframe src="${url}" frameborder="0"></iframe>
+    <style>
+      :host {
+        all: initial !important;
```
chrmod, line 21: @smalluban please take a look on this. some pages have styles that are as broad as `div`. Eg. example.com

smalluban:
Alredy done in #1104

PR #1082, opened 2023-03-22
MV3: Fix adblocker request and minor bugfixes

smalluban:
[Use rebase when merging]

See commits for a list of changes.


extension-manifest-v3/src/background/utils/request.js
```diff
@@ -0,0 +1,49 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { Request } from '@cliqz/adblocker';
+import { parse } from 'tldts-experimental';
+
+export default class ExtendedRequest extends Request {
+  static fromRequestDetails(details) {
```
chrmod, line 16: doesn't adblocker/webextension provide similar functionality?

chrmod (changes requested):

extension-manifest-v3/src/background/utils/request.js
smalluban: The source domain parsed data is lost, so it had to be parsed again. The adblocker-webextension calls `Request` directly, so it can not be overwritten with a version, which keeps the `sourceDomain`.

chrmod (approved):

PR #1084, opened 2023-03-23
Version bump + dev setup cleanup

chrmod:


philipp-classen (approved):

chrmod:
had to re-organise postinstall scripts as something changed in their order of execution and vendor-copy was failing due to missing ghostery-common. The problem may come from recent bump of nodejs to 19.x, but it's not confirmed.

PR #1089, opened 2023-03-24
Historical stats for new-tab extension

### LINKED ISSUE #1011: MV3: Historical stats
https://github.com/ghostery/common/blob/main/modules/insights/sources/background.es

smalluban:
Fixes #1011 and #1009.

* Adds daily statistics of found and blocked trackers
* Adds support for external messages from the new tab extension
* Fixes minor issue with missing `chrome.alarms.get` method on Firefox (used by session)
* Adds conditional cleaning indexedDbs, as Firefox does not support getting a full list of dbs

The new tab extension is not prepared for hiding values, which are constant zeros, so I'll make another PR there for it.


extension-manifest-v3/src/pages/settings/components/devtools.js
```diff
@@ -46,8 +46,11 @@ function clearStorage(host, event) {
       chrome.storage.local.clear();
 
       // Remove all indexedDBs
-      const dbs = await indexedDB.databases();
-      await Promise.all(dbs.map((db) => indexedDB.deleteDatabase(db.name)));
+      // Note: Firefox doesn't support indexedDB.databases()
```
chrmod, line 49: for Firefox we should still delete known databases

extension-manifest-v3/src/background/stats.js
```diff
@@ -79,6 +87,58 @@ const setIcon = throttle(
   __PLATFORM__ === 'firefox' ? 1000 : 250,
 );
 
+export async function getStats(since = '') {
+  let entries = await dailyStats.getAll();
```
chrmod, line 91: after few months, will this still be fast?

extension-manifest-v3/src/background/stats.js
```diff
@@ -16,13 +17,20 @@ import { getOffscreenImageData } from '@ghostery/ui/wheel';
...
 import Request from './utils/request.js';
-import { getMetadata } from './utils/trackerdb.js';
+import * as trackerDb from './utils/trackerdb.js';
 import Options, { observe } from '/store/options.js';
 import AutoSyncingMap from './utils/map.js';
 
-// If you bump this number, the extension will start with a
-// clean state. Normally, this should not be needed.
 const tabStats = new AutoSyncingMap({ storageKey: 'tabStats:v1' });
+const dailyStats = new AutoSyncingMap({
```
chrmod, line 25: not sure if this kind of storage is suitable for the data we put here. AutoSyncingMap pushes all its storage to `chrome.storage.local` on every update. That means it gets slower with the amount of data.

chrmod (changes requested):

extension-manifest-v3/src/background/stats.js
smalluban: At maximum, we will have there 1827 keys (if a user would use the browser every single day for 5 years). We are keeping in the local storage in MV2 a much bigger set of data, so this is not the case. 

Also, in MV3 we are "getting" specific keys (opposite to whole in MV2 - options are in the root object), so daily stats will be serialized for now, only if a new tab extension needs it.

extension-manifest-v3/src/background/stats.js
smalluban: As I wrote above, the data is limited to less than 2k items with a rather simple structure. As the local storage keeps an object instance to keep data, there is no other way to update the whole object.

However, I doubt that saving 2k items can take a long time (as it is done in "idle" - the action does not block anything) - it can be longer. If we really want that, we can measure it. 

Also, we can extend hard and soft times of sync, so it will happen less often.

extension-manifest-v3/src/pages/settings/components/devtools.js
smalluban: I'll look at this tomorrow

extension-manifest-v3/src/background/stats.js
```diff
@@ -16,13 +17,20 @@ import { getOffscreenImageData } from '@ghostery/ui/wheel';
...
+  storageKey: 'dailyStats:v1',
+  softFlushIntervalInMs: 1000, // 1 second
+  hardFlushIntervalInMs: 5000, // 5 seconds
+  ttlInMs: 1000 * 60 * 60 * 24 * 365 * 5, // 5 years
+  maxEntries: 365 * 5 + 2, // ~ 5 years of data + 2 days for 29th of February (worst case scenario)
```
philipp-classen, line 30: If entries are added each day, you might want to add a bit of buffer here and for the cleanup rely on the TTL to expire. If maxEntries hit, it will purge all entries, which means all history would be lost. (maxEntries is kind of a panic mode to detect cases that should never happen and then forces a reset).

extension-manifest-v3/src/background/stats.js
philipp-classen: I assume you could try to fill it to capacity to simulate the worst case. And then see if it would create issues. If the operation is not a critical path (or during startup), it might not be a problem.

chrmod: it's not only about the data size but also write frequency

extension-manifest-v3/src/background/stats.js
smalluban: I made a test of setting object with 2500 entries with the structure of the daily stats, and the average from 10 tries is about **21ms** on my machine (unfortunately slowing down the CPU in dev tools does not help - it means, that the core of the work is done in native part). 

* 2500 ~ 21ms
* 1500 ~ 13ms
* 500 ~ 8ms

I assume that on slower machines it might be longer, but still I don't expect anything more than 50ms - so it is quite fast ;)

extension-manifest-v3/src/pages/settings/components/devtools.js
smalluban: @chrmod @philipp-classen This change is not related to the main PR goal - I have noticed an error message when testing in Firefox. The arguments are correct, but I made a separate issue for that - https://github.com/ghostery/ghostery-extension/issues/1090

extension-manifest-v3/src/background/stats.js
chrmod: think of 2018 iPhone, or a mid-range Android device

extension-manifest-v3/src/background/stats.js
smalluban: The `chrome.storage.local.get()` is an async operation, so it does not block the execution of the service-worker sync operations.

extension-manifest-v3/src/background/stats.js
chrmod: Simulated data of one day with 280 different trackers:

Screenshot 2023-03-28 at 13 40 23

Fairly pessimistic estimation:
5 years = 1825 * 2586 bytes = 4,7 MB

smalluban:
As there so many doubts about keeping the data in local storage and put it into the memory, i think I would move it to the custom indexeddb instance.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -122,7 +124,7 @@ class Storage {
  */
 class IndexedDBKeyValueStore {
   constructor(dbName, { version = 1, objectStore = 'default' } = {}) {
-    this._dbName = dbName;
+    this._dbName = registerDatabase(dbName);
```
smalluban, line 127: Reporting generates db names in module code, so I came up with the idea of storing those names, so that when we want to clean up databases, we keep them.

smalluban:
@chrmod @philipp-classen With the last commit I have switched to the indexedDB solution (to avoid further doubts on the local storage, as some of the arguments against it were valid). Please review it again.

extension-manifest-v3/src/store/daily-stats.js
```diff
@@ -0,0 +1,103 @@
+import { store } from 'hybrids';
+import * as IDB from 'idb';
+
+import { registerDatabase } from '/utils/indexeddb.js';
+
+// Synchronously register name of the database
+// so if a user don't open any page, it is still possible
+// to delete the database from settings page
+const DB_NAME = registerDatabase('stats');
```
chrmod, line 9: can we use same db name and schema as on MV2 so we wont have to migrate the data?

extension-manifest-v3/src/background/utils/map.js
```diff
@@ -96,6 +96,13 @@ export default class AutoSyncingMap {
     return this.inMemoryMap.get(key);
   }
 
+  async getAll() {
+    this._warnIfOutOfSync();
```
philipp-classen, line 100: This function is unused now, right? Maybe we should remove it?

If keeping, we should remove _warnIfOutOfSync since it is async, waits for _pending and thus is always consistent from the perspective of the caller.

extension-manifest-v3/src/store/daily-stats.js
philipp-classen: To avoid introducing dexie, we could either drop all old stats in the MV2 -> MV3 move. Or do a one-time migration (in the worst case using the native indexedb API, even it is ugly).

smalluban:
I made a change, so now it uses `insights` DB name, and upgrades it (clean up unused stores). The data structure is messed up in MV2 (no consistency with counters, and extra fields, which are not needed anymore), so I made a one-time data migration (by design the upgrade method cannot be used for async operations, so it must be done outside).

I have only one problem with the solution - the `keyPath` of the already created store cannot be updated, so there must be a `day` property, even though it is redundant information. You're welcome if you know how to do it.

extension-manifest-v3/src/store/daily-stats.js
smalluban: Dixie or idb does not limit operations on the db instance - they are "wrapper", which makes handling it easier. The idb is much smaller and less abstract, but you can do everything with it like you would with a plain connection.

More here: https://github.com/ghostery/ghostery-extension/pull/1089#issuecomment-1490139236

extension-manifest-v3/src/background/utils/map.js
smalluban: You might be right, but after switching to indexedDB, changes in `map` are rather cosmetic, allowing initialize more than one map instance (it was a singleton before). If you think there are things to clean up, please make a separate issue for this.

philipp-classen: From what I see, the PR here only adds the extra function "getAll" though it no longer uses it.

Thinking about it again, I would rather remove it. Mostly since the class was originally designed to provide synchronous access (to be only used in situations where data loss is acceptable). Unless there is a reason, I would keep all functions synchronous (or only expose an explicit wait function to guarantee that loading is finished).

smalluban: Good point 👍

extension-manifest-v3/src/background/stats.js
```diff
@@ -173,4 +243,16 @@ if (__PLATFORM__ !== 'safari' && __PLATFORM__ !== 'firefox') {
       urls: ['<all_urls>'],
     },
   );
+
+  chrome.webRequest.onErrorOccurred.addListener(
+    (details) => {
+      if (details.error !== 'net::ERR_BLOCKED_BY_CLIENT') return;
```
smalluban, line 249: It looks like there is a way to check if the request was blocked on the client side - so it must be our extension... or another one :P This is the only drawback, which I could imagine - if a user has more adblockers installed, it might be false positive - but still, the request is blocked, so the information is correct.

extension-manifest-v3/src/background/devtools.js
```diff
@@ -0,0 +1,33 @@
+import { store } from 'hybrids';
```
chrmod, line 1: please add copywrite header

extension-manifest-v3/src/background/reporting.js
chrmod: wonder how we can make this work if external dependencies like whotracksme/packages would like to use indexeddb. For now reporting uses dependency injection, but upcoming antitracking has few databases - it may be quite complicated

extension-manifest-v3/src/utils/indexeddb.js
```diff
@@ -0,0 +1,26 @@
+import { deleteDB } from 'idb';
```
chrmod, line 1: please add copywrite header

chrmod (approved):

extension-manifest-v3/src/background/reporting.js
smalluban: As the last resort, we can initialize: 

```js
const dbs = new Set(['known-database-name-which-cant-be-registered']);
```

extension-manifest-v3/src/store/daily-stats.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  if (upgradeFromMV2) {
+    const db = getDb.current;
+    const oldStats = await db.getAll('daily');
+    const tx = db.transaction('daily', 'readwrite');
+    const daily = tx.objectStore('daily');
+
+    for (const stats of oldStats) {
+      daily.delete(stats.day);
```
philipp-classen, line 62: Since it returns a promise, arguably we should await it:
```
await daily.delete(stats.day);
```

extension-manifest-v3/src/store/daily-stats.js
```diff
@@ -0,0 +1,164 @@
...
+      },
+      blocking() {
+        getDb.current.close();
+        getDb.current = null;
+        upgradeFromMV2 = false;
+      },
+    });
+  }
+
+  if (upgradeFromMV2) {
```
philipp-classen, line 55: Since it is a rare and fairly critical operation, logging when a migration was started and then again at the end (if it was successful), would be a good idea.

philipp-classen (approved):

chrmod (approved):

PR #1104, opened 2023-04-05
Onboarding v3

smalluban:


extension-manifest-v3/src/background/onboarding.js
```diff
@@ -9,21 +9,34 @@
...
+import { sendShowIframeMessage } from './utils/iframe.js';
+import { tabStats } from './stats.js';
 
-const DAY_IN_MS = 24 * 60 * 60 * 1000;
+const NOTIFICATION_DELAY = 60 * 60 * 1000; // an hour in milliseconds
 
-(async function onboarding() {
-  const options = await store.resolve(Options);
-  const now = Date.now();
+let shownAt = undefined;
```
philipp-classen, line 18: From what I see, shownAt can have the three values "undefined | true | false", but is used later from computation as a timestamp (Date.now() - shownAt > NOTIFICATION_DELAY).

smalluban: It is first check if it is truthy: `shownAt &&`

philipp-classen: Yes, but isn't `Date.now() - shownAt` in the next line effectively `Date.now() - true`? In other words, `Date.now() - 1`?

philipp-classen: I see. Didn't notice that when reading over it. Still note that shownAt will become "false" if "onboard.done" is true. If you take all paths, it can have the states undefined, false, or be a Unix epoche. So, in TypeScript it would be "undefined | boolean | number".

From a reader perspective, splitting the state into a flag and a timestamp is an alternative to consider. But from what I can tell, the current logic is correct though.

chrmod (approved):

PR #1113, opened 2023-04-17
Use generated lists for DNR rules

smalluban:
Fixes https://github.com/ghostery/ghostery-dnr-lists/issues/17

This is not complete, as I still keep getting errors like "Failed to apply rules...", but as I get false positives, I can't get what rules are breaking it.


chrmod:
@smalluban we probably should bump a minimum version of MacOS/iOS as urlFilters were added fairly recently AFAICT

smalluban:
According to [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/declarativeNetRequest#browser_compatibility) we can set the minimum version to 15.4 for iOS.

However, for the macOS, it is not that trivial. The Safari version is not tight with the OS version, so it is very likely that someone who has macOS 12 can have the latest version of Safari. Moreover, there are laptops that ended support for newer versions of macOS, but they still might receive Safari updates (this is my guess).

extension-manifest-v3/scripts/download-engines/utils.js
```diff
@@ -0,0 +1,96 @@
...
+    rule.condition.regexFilter?.match(/(\{\d*,\d*\}|\{\d\}|\|)/) ||
+    !supportedActions.includes(rule.action.type) ||
+    (resourceTypes && resourceTypes.length === 0)
+  ) {
+    return null;
+  }
+
+  return {
+    priority: rule.priority,
+    ...(rule.action.type === 'allowAllRequests'
```
chrmod, line 46: should we add some unit tests for all that?

chrmod (approved):

extension-manifest-v3/scripts/download-engines/utils.js
smalluban: Finally, this should be taken out from the repository to https://github.com/ghostery/adblocker-filters. We can keep it here to test out with upcoming release and then move it out.

PR #1115, opened 2023-04-18
Fixes a potential race during reporting startup

philipp-classen:
Fixes a potential race in the reporting initialization. When the service worker wakes up and immediately runs the listener callback, the first event may get lost.

Also, include a fix for a race that is documented here: https://github.com/ghostery/ghostery-extension/pull/1129
But fixed by postponing the observer registration (not as in #1129, which is more elaborate).

### DIFF
--- a/extension-manifest-v3/src/background/reporting.js
+++ b/extension-manifest-v3/src/background/reporting.js
@@ -216,14 +216,8 @@ const reporting = new Reporting({
   communication,
 });
 
-const pendingStartup = new Promise((done) => {
-  let timeout = setTimeout(() => {
-    console.warn('Startup takes too long. Waking up pending listeners...');
-    timeout = null;
-    done();
-  }, 3000);
-
-  observe('terms', async (terms) => {
+let pendingStartup = (async function () {
+  await observe('terms', async (terms) => {
     try {
       if (terms) {
         try {
@@ -238,11 +232,10 @@ const pendingStartup = new Promise((done) => {
         reporting.unload();
       }
     } finally {
-      clearTimeout(timeout);
-      done();
+      pendingStartup = null;
     }
   });
-});
+})();
 
 function delay(timeInMs) {
   return new Promise((resolve) => setTimeout(resolve, timeInMs));
@@ -280,7 +273,16 @@ function onLocationChange(details) {
     }
 
     try {
-      await pendingStartup;
+      if (pendingStartup) {
+        await Promise.race([
+          pendingStartup,
+          delay(2000).then(() => {
+            throw new Error(
+              'Init of reporting is hanging. Skipping event to avoid queuing.',
+            );
+          }),
+        ]);
+      }
       if (!reporting.isActive) {
         return;
       }
--- a/extension-manifest-v3/src/background/reporting.js
+++ b/extension-manifest-v3/src/background/reporting.js
@@ -216,17 +216,32 @@ const reporting = new Reporting({
   communication,
 });
 
-observe('terms', (terms) => {
-  if (terms) {
-    reporting.init().catch((e) => {
-      console.warn(
-        'Failed to initialize reporting. Leaving the module disabled and continue.',
-        e,
-      );
-    });
-  } else {
-    reporting.unload();
-  }
+const pendingStartup = new Promise((done) => {
+  let timeout = setTimeout(() => {
+    console.warn('Startup takes too long. Waking up pending listeners...');
+    timeout = null;
+    done();
+  }, 3000);
+
+  observe('terms', async (terms) => {
+    try {
+      if (terms) {
+        try {
+          await reporting.init();
+        } catch (e) {
+          console.warn(
+            'Failed to initialize reporting. Leaving the module disabled and continue.',
+            e,
+          );
+        }
+      } else {
+        reporting.unload();
+      }
+    } finally {
+      clearTimeout(timeout);
+      done();
+    }
+  });
 });
 
 function delay(timeInMs) {
@@ -234,8 +249,6 @@ function delay(timeInMs) {
 }
 
 function onLocationChange(details) {
-  if (!reporting.isActive) return;
-
   const { url, frameId, tabId } = details;
   if (frameId !== 0 || url === 'about:blank' || url.startsWith('chrome://')) {
     return;
@@ -267,6 +280,11 @@ function onLocationChange(details) {
     }
 
     try {
+      await pendingStartup;
+      if (!reporting.isActive) {
+        return;
+      }
+
       const jobRegistered = await reporting.analyzeUrl(url);
       if (jobRegistered) {
         // TODO: This part here is not robust:
--- a/extension-manifest-v3/src/background/reporting.js
+++ b/extension-manifest-v3/src/background/reporting.js
@@ -218,23 +218,20 @@ const reporting = new Reporting({
 
 let pendingStartup = (async function () {
   await observe('terms', async (terms) => {
-    try {
-      if (terms) {
-        try {
-          await reporting.init();
-        } catch (e) {
-          console.warn(
-            'Failed to initialize reporting. Leaving the module disabled and continue.',
-            e,
-          );
-        }
-      } else {
-        reporting.unload();
+    if (terms) {
+      try {
+        await reporting.init();
+      } catch (e) {
+        console.warn(
+          'Failed to initialize reporting. Leaving the module disabled and continue.',
+          e,
+        );
       }
-    } finally {
-      pendingStartup = null;
+    } else {
+      reporting.unload();
     }
   });
+  pendingStartup = null;
 })();
 
 function delay(timeInMs) {
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -161,14 +161,12 @@ export async function observe(property, fn) {
       value = options[property];
 
       try {
-        return await fn(value, prevValue);
+        await fn(value, prevValue);
       } catch (e) {
         console.error(`Error while observing options: `, e);
       }
     }
   };
-
-  observers.add(wrapper);
 
   try {
     const options = await store.resolve(Options);
@@ -179,6 +177,7 @@ export async function observe(property, fn) {
   } catch (e) {
     console.error(e);
   }
+  observers.add(wrapper);
 
   // Return unobserve function
   return () => {


extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,26 +216,39 @@ const reporting = new Reporting({
...
+          console.warn(
+            'Failed to initialize reporting. Leaving the module disabled and continue.',
+            e,
+          );
+        }
+      } else {
+        reporting.unload();
+      }
+    } finally {
+      clearTimeout(timeout);
```
smalluban, line 241: The observer is called multiple times, if the background process starts with `terms: false` and then `terms: true` - for example a usual case when a user installs an extension. I think `done()` nor `clearTimeout` should not be called multiple times.

philipp-classen: Since the timer is set to "null", calling clearTimeout should be safe. Calling resolve (the "done" function") multiple times is also safe: the first call wins and following calls will be ignored by Promises that are already resolved.

From pure correctness, I should the current version is safe. Unless changing has other benefits that I'm missing?

extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,26 +216,32 @@ const reporting = new Reporting({
...
+          );
+        }
+      } else {
+        reporting.unload();
+      }
+    } finally {
+      pendingStartup = null;
+    }
+  });
+})();
```
smalluban, line 238: You can simply pass here `pendingStartup = null`, as you `await observe()..`, and remove `try {} finally {}` wrapper.

See here: https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v3/src/background/adblocker.js#L51

smalluban (changes requested):

extension-manifest-v3/src/background/reporting.js
```diff
@@ -267,6 +273,20 @@ function onLocationChange(details) {
     }
 
     try {
+      if (pendingStartup) {
+        await Promise.race([
```
smalluban, line 277: This might be a good pattern to add in `adblocker.js` as if indeed promise hangs, the events will be queuing...

extension-manifest-v3/src/background/reporting.js
smalluban: Now I think, it might be a good idea to create a global util function, which takes an async function, and returns a promise, which resolves, or rejects in some threshold. Then in 277 it could just safely await the promise.

Something like:

```js
const setup = asyncSetup(async () => { ... });

...
if (setup.promise) {
  await setup.promise;
}
```

Obviously, the util should return an object with promise, which is cleared.

philipp-classen: I think that would break: "pendingStartup" resolves immediately while the callback is still pending. For instance:

```
let pendingStartup = (async function () {
  await observe('terms', async (terms) => {
    await delay(10000000);  // <--- simulate that reporting hangs 
    if (terms) {
      try {
        await reporting.init();
      } catch (e) {
        console.warn(
          'Failed to initialize reporting. Leaving the module disabled and continue.',
          e,
        );
      }
    } else {
      reporting.unload();
    }
  });
  pendingStartup = null;
})();
````

That was why I moved the "pendingStartup = null" at the end of the callback execution. Or did I misunderstood what you meant?

Comment (philipp-classen) [reply extension-manifest-v3/src/background/reporting.js:238]: I debugged into (with changed version above). It hits another bug is related to the change-detection in line https://github.com/ghostery/ghostery-extension/blob/b05fa7de45e02ad52f86b45f025f2328348f53fc/extension-manifest-v3/src/store/options.js#L159

In a nutshell, adding it to the observer and later calling the wrapper races.

The wrapper function is called twice:
1. As part of the observers set:https://github.com/ghostery/ghostery-extension/blob/b05fa7de45e02ad52f86b45f025f2328348f53fc/extension-manifest-v3/src/store/options.js#L92
2. When it is now called from reporting: https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v3/src/store/options.js#L178, the wrapper has been already execute (part of the observer set)

Only the first call blocks (but in the observer trigger, it's ignored: https://github.com/ghostery/ghostery-extension/blob/b05fa7de45e02ad52f86b45f025f2328348f53fc/extension-manifest-v3/src/store/options.js#L92). When it now runs for the second time, the value matches the old value and the call is skipped (value=false and options[property]=false): https://github.com/ghostery/ghostery-extension/blob/b05fa7de45e02ad52f86b45f025f2328348f53fc/extension-manifest-v3/src/store/options.js#L159

The effect is that the promise now immediately resolves ("observe" resolved), even though the function is still executing.

Comment (philipp-classen) [reply extension-manifest-v3/src/background/reporting.js:238]: It's the question how to proceed. If we fix the race in "observe", I also think it could be changed to the alternative version. Since from the caller perspective, the semantic that the first execution is guaranteed to have finished is more intuitive, I would prefer that.

But maybe it should be fixed in its own PR, since it has global effect. And once it's fixed, the try-finally construction with de-facto serves as a workaround could be removed.

Addressing the race is a bit subtle. For instance, postponing the registration of the observer could fix it in this case, but may break other code that currently relies on that behavior.

Comment (smalluban) [reply extension-manifest-v3/src/background/reporting.js:238]: Calling the `observe()` function, and then `fn` by internal options implementation is not related. It works correctly in that it won't call the `fn` again if the value hasn't changed.

It's a simple thing, `observe()` returns a promise, which guarantees that it is resolved after calling `fn` (at least once). This is the only thing that matters here. If the observed value changes later/in the middle/whatever time, it will call `fn` again with a defined side effect. 

The core "problem" of the implementation is that some features, which can be required at any time require configuration, which is getting async (`chrome.storage.local`). 

The observe method rejects if options cannot be fetched, or when `fn` rejects. I am not even sure if there is a need for `Promise.race` with a timeout, as there is no option to hang that promise.

In my opinion the pattern in `adblocker.js` is fine, but as we have more places, which need that structure, it might be a good idea to create a helper util, which covers setting promise, and then clear it out (this is also an optimization, that events don't have to wait every time for already fulfilled promise).

Comment (philipp-classen) [reply extension-manifest-v3/src/background/reporting.js:238]: My point is that with the current implementation, observe is not guaranteed to block (see above). Only on the first call, but there are multiple callers out of your control, so you can't rely on it.

The PR in the current state is not affected, since it waits in the call. But if I would make the change that you requested (replacing try-finally), it would hit the bug.

Comment (philipp-classen) [reply extension-manifest-v3/src/background/reporting.js:238]: To make it clear what I mean, this other PR fixes the problem that I tried to describe: https://github.com/ghostery/ghostery-extension/pull/1129

The other PR would guarantee a caller that the wrapper ran once after "await observe" resolves. I tested with that fix, and then I can remove the try-finally constructor and move the nulling out of the function.

Comment (smalluban) [reply extension-manifest-v3/src/background/reporting.js:238]: Yes, you have a guarantee that the passed function to the `observe()` has finished - https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v3/src/store/options.js#L178, when you initialize observer by calling `await observe(fn)`.

Comment (smalluban) [reply extension-manifest-v3/src/background/reporting.js:238]: The only thing that might be changed is to move that line after the initial call - https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v3/src/store/options.js#L171, just before the `return` statement. It would avoid calling fn while there is an initial call (which is pointless).

Comment (philipp-classen) [reply extension-manifest-v3/src/background/reporting.js:238]: > Yes, you have a guarantee that the passed function to the `observe()` has finished - https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v3/src/store/options.js#L178, when you initialize observer by calling `await observe(fn)`.

That is not correct. It needs a fix that like one here: https://github.com/ghostery/ghostery-extension/pull/1129

At the moment, we don't have the guarantee. The reason is the race though the observer set. It can be reproduce with the change above (https://github.com/ghostery/ghostery-extension/pull/1115#discussion_r1172593303) that simulates a delay.

**Comment by philipp-classen** [comment]
Closing this one to preserve it to understand https://github.com/ghostery/ghostery-extension/pull/1115.

The more elegant fix is to register the observer after the first call was completely. That eliminates all races between the first call and the execution of observers.

I'll include the fix (moving the registration) to https://github.com/ghostery/ghostery-extension/pull/1115 and use it to simplify it.

File: extension-manifest-v3/src/background/reporting.js (hunk 2 above)
Comment (philipp-classen) [reply extension-manifest-v3/src/background/reporting.js:235]: Included the fix now for the observer race (by moving the registration), and applied the simplification now (getting rid of try-finally).

File: extension-manifest-v3/src/background/reporting.js [hunk 4]
```diff
@@ -267,6 +270,20 @@ function onLocationChange(details) {
...
+          pendingStartup,
+          delay(2000).then(() => {
```
smalluban, line 276: The last thing, which I am not sure, is if we need that timeout. In this case (and adblocker too), we mainly waits for the local storage to return data, no another "real" async work. It should not take more than few ms. 
Another thing is that is anything goes wrong, the promise rejects and does not clear itself, so that case is also covered.

philipp-classen: If it rejects (or times out), it will be handled by the catch.

Reporting does more things during initialization. It's true that a slow startup (exceeding the timeout) would be a bug, but I would still leave the timeout. It's easy to accidentally introduce delay, for instance, if a network request hangs and blocks startup, etc.

A difference to the adblocker is also that queuing up request in reporting is arguably more harmful.

extension-manifest-v3/src/store/options.js
```diff
@@ -161,15 +161,13 @@ export async function observe(property, fn) {
       value = options[property];
 
       try {
-        return await fn(value, prevValue);
+        await fn(value, prevValue);
```
chrmod, line 164: why we drop the return value from here? It may be useful is some new use-cases.

smalluban (approved):

extension-manifest-v3/src/store/options.js
philipp-classen: I changed it because my linter complained. We don't use the value and the function isn't guaranteed to return a value (because of the catch).

smalluban:
@chrmod @philipp-classen Is there anything, which holds us from merging this PR?

chrmod:
@smalluban give me a bit time. I may have a bit simpler implementation.

philipp-classen:
Closing in favor of https://github.com/ghostery/ghostery-extension/pull/1152

PR #1129, opened 2023-04-21
Guarantee that after "await observe" the callback was executed at least once

philipp-classen:
Background: observers (in the Set) and the normal wrapper call (during calling "observe") race for being the first to see the property change. The second call will no longer block since the value matches the last seen value. This comment describes it in more detail: https://github.com/ghostery/ghostery-extension/pull/1115#discussion_r1172923900

This change changes the semantic to provide the guarantee to the caller of observe that the wrapper function finished its first call.

refs https://github.com/ghostery/ghostery-extension/pull/1115

### DIFF
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -154,6 +154,11 @@ async function migrateFromMV2() {
 }
 
 export async function observe(property, fn) {
+  let called;
+  const firstCall = new Promise((resolve) => {
+    called = resolve;
+  });
+
   let value;
   const wrapper = async (options) => {
     if (value === undefined || options[property] !== value) {
@@ -161,13 +166,21 @@ export async function observe(property, fn) {
       value = options[property];
 
       try {
-        return await fn(value, prevValue);
+        await fn(value, prevValue);
       } catch (e) {
         console.error(`Error while observing options: `, e);
+      } finally {
+        called();
       }
     }
   };
 
+  // Note: this is why the "firstCall" detection is added. When observers
+  // are called, they may be first in seeing the initial value change.
+  // When we call the wrapper (in this function), it may be already the
+  // second time that the wrapper gets called. In that case, it would
+  // resolve immediately (while the first call that the observer triggered
+  // might still be pending).
   observers.add(wrapper);
 
   try {
@@ -179,6 +192,7 @@ export async function observe(property, fn) {
   } catch (e) {
     console.error(e);
   }
+  await firstCall;
 
   // Return unobserve function
   return () => {


extension-manifest-v3/src/store/options.js
```diff
@@ -154,20 +154,33 @@ async function migrateFromMV2() {
...
 
+  // Note: this is why the "firstCall" detection is added. When observers
+  // are called, they may be first in seeing the initial value change.
+  // When we call the wrapper (in this function), it may be already the
+  // second time that the wrapper gets called. In that case, it would
+  // resolve immediately (while the first call that the observer triggered
+  // might still be pending).
   observers.add(wrapper);
 
   try {
```
smalluban, line 186: I just don't agree -

We have scope of the async functions:

```
const promise = async observe() -> await wrapper() -> await fn()
```

The above means that `promise` is only fulfills if a **full** chain of await are fulfilled. You don't need `called()` or other stuff.

smalluban (changes requested):

extension-manifest-v3/src/store/options.js
smalluban: The only thing that can help is that - https://github.com/ghostery/ghostery-extension/pull/1115#discussion_r1173592533

smalluban: If you move `observers.add(wrapper)` after the `try {} catch {}`, the wrapper cannot be called until the first execution ends. I think this is a proper solution, no need for any other change.

philipp-classen: The race can happen as following (see https://github.com/ghostery/ghostery-extension/pull/1115#discussion_r1172923900):

* observer.add(wrapper) adds the wrapper to the global set
* it may get executed here (before await wrapper runs!): https://github.com/ghostery/ghostery-extension/blob/b05fa7de45e02ad52f86b45f025f2328348f53fc/extension-manifest-v3/src/store/options.js#L92
* as a side-effect, "value" gets updated
* now, "await wrapper" gets called. Since the value hasn't changed, it immediately resolves.

I also provided a minimal reproducible example. You should be able to reproduce for yourself if you don't trust me.

smalluban: Ok, I trust you ;) But the only thing that should be changed is to move L184 after try/catch.

philipp-classen: Let me test that.

philipp-classen: Moving the observer below also fixes it. I agree that it is clearer. So far, I also don't notice side-effects (that we have code that depends on the current behavior).

I'll probably close this PR, since I mostly opened it to document the race (which would be lost). And include the fix in the other PR, which I can then simplify with the other changes that you recommend. In combination with both changes, it should be possible.

philipp-classen:
Closing this one to preserve it to understand https://github.com/ghostery/ghostery-extension/pull/1115.

The more elegant fix is to register the observer after the first call was completely. That eliminates all races between the first call and the execution of observers.

I'll include the fix (moving the registration) to https://github.com/ghostery/ghostery-extension/pull/1115 and use it to simplify it.

PR #1131, opened 2023-04-21
MV3: unit tests in Deno

chrmod:
This PR is a POC that Deno can be used to create webextension like environment to perform unit tests for webextenions outside of the browser.


extension-manifest-v3/test/store/setup.js
```diff
@@ -0,0 +1,4 @@
+import './pre-setup.js';
```
chrmod, line 1: pre-setup is needed so chrome is present as a global before `webextension-polyfill` is loaded.

extension-manifest-v3/package.json
```diff
@@ -30,14 +30,16 @@
...
-    "web-ext": "^7.3.1"
+    "web-ext": "^7.3.1",
+    "webextension-polyfill": "^0.10.0"
   },
   "dependencies": {
     "@cliqz/adblocker": "^1.26.5",
     "@cliqz/adblocker-webextension-cosmetics": "^1.22.7",
     "@duckduckgo/autoconsent": "^4.0.0",
-    "@ghostery/libs": "^1.0.0",
+    "@ghostery/libs": "file:../libs",
```
chrmod, line 42: was trying to use Deno source maps to deal with this but without luck

extension-manifest-v3/test/store/setup.js
```diff
@@ -0,0 +1,4 @@
...
+import browser from 'webextension-polyfill';
```
chrmod, line 2: `webextension-polyfill` is needed as `sinon-chrome` does not support promise based APIs

extension-manifest-v3/test/store/options.test.js
```diff
@@ -0,0 +1,16 @@
+import './setup.js';
```
chrmod, line 1: this should be called before all tests

extension-manifest-v3/test/store/pre-setup.js
```diff
@@ -0,0 +1,6 @@
+import chrome from 'sinon-chrome';
+
+window.document = {};
```
chrmod, line 3: required for `hybrids` to prevent:
```
./test/store/options.test.js (uncaught error)
error: TypeError: Cannot read properties of undefined (reading 'adoptedStyleSheets')
    at file:///Users/chrmod/Projects/github.com/ghostery/ghostery-extension/extension-manifest-v3/node_modules/.deno/hybrids@8.1.15/node_modules/hybrids/src/template/layout.js:3:49
This error was not caught from a test and caused the test runner to fail on the referenced module.
It most likely originated from a dangling promise, event/timeout handler or top-level code.
```

chrmod:
addit

PR #1137, opened 2023-04-25
MV3: fix reporing startup race

chrmod:
Alternative implementation of https://github.com/ghostery/ghostery-extension/pull/1115

### DIFF
--- a/extension-manifest-v3/src/background/reporting.js
+++ b/extension-manifest-v3/src/background/reporting.js
@@ -220,25 +220,38 @@ function delay(timeInMs) {
   return new Promise((resolve) => setTimeout(resolve, timeInMs));
 }
 
-function defer() {
-  const deferred = {};
-  deferred.promise = new Promise((resolver) => {
-    deferred.resolve = resolver;
+function guard() {
+  const _guard = {};
+
+  let wasUpdatedOnce = false;
+  let promsie = new Promise((resolver) => {
+    _guard.update = (value) => {
+      wasUpdatedOnce = true;
+      resolver(value);
+    };
   });
-  deferred.waitUntil = async ({
-    timeout,
-    errorMessage = 'Timeout exceeded',
-  } = {}) =>
+
+  _guard.reset = () => {
+    if (wasUpdatedOnce) {
+      promsie = new Promise((resolver) => {
+        _guard.update = resolver;
+      });
+    }
+  };
+
+  _guard.waitUntil = ({ timeout, errorMessage = 'Timeout exceeded' } = {}) =>
     Promise.race([
-      deferred.promise,
+      promsie,
       delay(timeout).then(() => Promise.reject(errorMessage)),
     ]);
-  return deferred;
+
+  return _guard;
 }
 
-const deferredStartup = defer();
+const startupGuard = guard();
 
 observe('terms', async (terms) => {
+  startupGuard.reset();
   if (terms) {
     try {
       await reporting.init();
@@ -251,7 +264,7 @@ observe('terms', async (terms) => {
   } else {
     reporting.unload();
   }
-  deferredStartup.resolve(reporting.isActive);
+  startupGuard.update(reporting.isActive);
 });
 
 function onLocationChange(details) {
@@ -287,11 +300,11 @@ function onLocationChange(details) {
 
     try {
       if (
-        await defer.waitUntil({
+        !(await startupGuard.waitUntil({
           timeout: 2000,
           errorMessage:
             'Init of reporting is hanging. Skipping event to avoid queuing.',
-        })
+        }))
       ) {
         return;
       }
--- a/extension-manifest-v3/src/background/reporting.js
+++ b/extension-manifest-v3/src/background/reporting.js
@@ -220,38 +220,21 @@ function delay(timeInMs) {
   return new Promise((resolve) => setTimeout(resolve, timeInMs));
 }
 
-function guard() {
-  const _guard = {};
-
-  let wasUpdatedOnce = false;
-  let promsie = new Promise((resolver) => {
-    _guard.update = (value) => {
-      wasUpdatedOnce = true;
-      resolver(value);
-    };
+function defer() {
+  const deferred = {};
+  deferred.promise = new Promise((resolve) => {
+    deferred.resolve = resolve;
   });
-
-  _guard.reset = () => {
-    if (wasUpdatedOnce) {
-      promsie = new Promise((resolver) => {
-        _guard.update = resolver;
-      });
-    }
-  };
-
-  _guard.waitUntil = ({ timeout, errorMessage = 'Timeout exceeded' } = {}) =>
-    Promise.race([
-      promsie,
-      delay(timeout).then(() => Promise.reject(errorMessage)),
-    ]);
-
-  return _guard;
+  return deferred;
 }
 
-const startupGuard = guard();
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,23 +216,41 @@ const reporting = new Reporting({
...
   }
+  deferredStartup.resolve(reporting.isActive);
 });
 
-function delay(timeInMs) {
-  return new Promise((resolve) => setTimeout(resolve, timeInMs));
-}
-
 function onLocationChange(details) {
   if (!reporting.isActive) return;
```
philipp-classen, line 255: This checked needs to be removed to resolve the race. Or at least delayed before the call to reporting.analyzeUrl.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,23 +216,41 @@ const reporting = new Reporting({
...
+function delay(timeInMs) {
+  return new Promise((resolve) => setTimeout(resolve, timeInMs));
+}
+
+function defer() {
+  const deferred = {};
+  deferred.promise = new Promise((resolver) => {
+    deferred.resolve = resolver;
+  });
+  deferred.waitUntil = async ({ timeout, errorMessage = '' } = {}) =>
```
philipp-classen, line 228: Empty string is not a good default value (even if it is not used here), since you may end up with `Uncaught (in promise) <empty string>`. Maybe something like `Timeout exceeded` or leave out the default and the caller always pass it.

smalluban:
As I mentioned in another PR, I think we should create a global helper function. It should take an async callback, and support a timeout race (and then rejects the promise). 

We are trying to do the same over and over again. I think the https://github.com/ghostery/ghostery-extension/pull/1115 after many discussions and changes is fine to be merged. 

I can prepare the global function, so you will see what I mean.

extension-manifest-v3/src/background/reporting.js
chrmod: I will remove it from here as this way all events prior to initialisation of reporting will be dropped.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,26 +216,45 @@ const reporting = new Reporting({
...
       console.warn(
         'Failed to initialize reporting. Leaving the module disabled and continue.',
         e,
       );
-    });
+    }
   } else {
     reporting.unload();
   }
+  deferredStartup.resolve(reporting.isActive);
```
philipp-classen, line 254: Combining the state and a defer is problematic, since it won't detect changes.

For instance:
* Start with a new installation (the deferred will soon resolve to "false")
* Enable terms (the module now initializes, but the deferred is still "false")

In the end, the module is technically enabled, but because of the "if" below it remains practically disabled.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -267,6 +286,16 @@ function onLocationChange(details) {
     }
 
     try {
+      if (
```
philipp-classen, line 289: Like the comment above, this will fail if the state changes.

Dropping the "if" and only awaiting the defer (as if waiting if the initial setup is complete) should work. And let the defer resolve without any value.

Or let it always resolve to "true" to indicate the the initial startup is done (but IMHO that is less clear).

extension-manifest-v3/src/background/reporting.js
chrmod: I see.

Isn't the same happening in the other implementation. In the and pendingPromise was being used only one time.

philipp-classen: In https://github.com/ghostery/ghostery-extension/pull/1115/commits/1a5a373f6e91b820d95613f5c89ab5f9d3836280 or https://github.com/ghostery/ghostery-extension/pull/1115/files, the goal of pendingStartup is to delay the processing of the event. It doesn't have the semantic that the module is afterwards fully loaded. Also, it relies on the initialization in the module itself.

The main challenge is that to initialize, you need to read persisted state (to learn if it starts enabled or disabled). That forces at least one async operation. That unfortunately clashes with the other requirement in service workers to be able to process events on the first tick.

(That is where the idea to delay comes from. It's not the only implementation strategy; buffering events during startup would be another. Though buffering may conflict with in low-latency requirements, e.g. the adblocker.)

chrmod:
when thinking about a perfect interface I see pushing more responsibilities into the reporting module itself. That is - its methods should await own initialization and respect own state. That way to code could be simplified into this:

```js
const report = new Reporting();

observer('terms', (terms) => {
  terms ? reporting.init() : reporting.unload();
});

chrome.webNavigation.onCommitted.addListener(async ({ url }) => {
  await Promise.race([
    reporting.isReady(), 
    delay(2000).then(() => Promise.reject('timeout')),
  ]);
  await reporting.analyzeUrl(url);
});
```
@philipp-classen what do you think?

philipp-classen:
> when thinking about a perfect interface I see pushing more responsibilities into the reporting module itself. That is - its methods should await own initialization and respect own state. That way to code could be simplified into this:
> 
> ```js
> const report = new Reporting();
> 
> observer('terms', (terms) => {
>   terms ? reporting.init() : reporting.unload();
> });
> 
> chrome.webNavigation.onCommitted.addListener(async ({ url }) => {
>   await Promise.race([
>     reporting.isReady(), 
>     delay(2000).then(() => Promise.reject('timeout')),
>   ]);
>   await reporting.analyzeUrl(url);
> });
> ```
> 
> @philipp-classen what do you think?

Taking a step back. As I understand, we are currently trying at the same time solving two independent problems at the same time:
1. Dealing with the concrete case: the short time period after the service worker comes up but where the extension was not yet able to read something from storage and (depending on the state) finish the initialization (or keep it disabled)
2. Finding a generic way that abstracts to other modules (which as I understand drives the discussion)

For 1) I think it is likely that reporting will eventually handle all listener registrations internally, mostly because upcoming code seems to go in that direction. That would simplify the startup (making the glue code here almost trivial); but the complexity doesn't go away but only shifts into the module. (The details are not fully clear to me yet, so I would postpone internal changes in the module until it becomes more concrete.)

For 2) The code above also removed all error handling. I assume in a full implementation, you would want to handle them (mostly to get reliable stack traces). When it gets more verbose due to error handling, I wonder if it is overall much simpler. The core simplification, as I understand it, is to move the logic into the module. As said, I think that will eventually happen but I feel it's easier to reason about it once we get there naturally.

extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,26 +216,43 @@ const reporting = new Reporting({
...
   } else {
     reporting.unload();
   }
-});
 
-function delay(timeInMs) {
-  return new Promise((resolve) => setTimeout(resolve, timeInMs));
-}
+  startupDefer.resolve();
+  startupDefer = null;
```
philipp-classen, line 252: I wouldn't "null" it out. (And thus you can also simplify a bit by removing
```
 if (!startupDefer) {
    startupDefer = defer();
  }
  ```
  and later you don't have to check for being null.
  

The reason is that the semantic change between nulling and not nulling is that when terms are toggled (but it has been already initialized), it will again wait. Without nulling, it will not wait.

Not waiting is the preferred semantic here IMHO. And has also the benefit that its simplifies the implementation.

PR #1147, opened 2023-04-28
MV3: fix reporing startup race

chrmod:
Alternative for https://github.com/ghostery/ghostery-extension/pull/1137 and #1115

### DIFF
--- a/extension-manifest-v3/src/background/utils/startup-guard.js
+++ b/extension-manifest-v3/src/background/utils/startup-guard.js
@@ -1,0 +1,56 @@
+export default (starter, stopper) => {
+  let resolver, rejecter, init;
+  let isPending = true;
+  const reset = () => {
+    isPending = true;
+    init = new Promise((resolve, reject) => {
+      resolver = () => {
+        resolve();
+        isPending = false;
+      };
+      rejecter = () => {
+        reject();
+        isPending = false;
+      };
+    });
+  };
+  reset();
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/src/background/utils/startup-guard.js
```diff
@@ -0,0 +1,56 @@
+export default (starter, stopper) => {
```
smalluban, line 1: If you want to have the option to abort a "pending" promise, we can use the `AbortController` API, as it is already widely supported: https://caniuse.com/abortcontroller (including Safari, which is our main target currently).

smalluban, line 1: Please use the `function` statement in modules where possible (if you don't set `const something = () => {}`. Also, the copyright missing.

smalluban (changes requested):
Generally, this solution only adds a feature to switch the result of the setup process, so it is mutable (not like in #1152, where once finished, the promise is cleared).

However, it does not help with the reporting much, as we could simply check `reporting.isActive` as before. When `terms` go to `false` (which is only possible by dev tools, which clears out the options), observe method can just call `reporting.unload()` - which is sync operation by the way.

I can't see here a case, which cannot be covered with #1152

chrmod:
fixed by https://github.com/ghostery/ghostery-extension/pull/1152

PR #1152, opened 2023-05-09
Add asyncSetup utility

smalluban:
I know that it is a third attempt at the same subject, but I think it is worth considering this solution. It is very short, covers all of the edge cases, supports race conditions, and takes only 20 lines of code. 

To use it, simply:

```js
const setup = asynsSetup([promise1, promise2]);

// consume in async function
```

### DIFF
--- a/extension-manifest-v3/src/background/reporting.js
+++ b/extension-manifest-v3/src/background/reporting.js
@@ -16,6 +16,8 @@ import * as IDB from 'idb';
 import { observe } from '/store/options.js';
 import { registerDatabase } from '/utils/indexeddb.js';
 
+import asyncSetup from './utils/setup.js';
+
 function platformSpecificSettings() {
   if (
     /iPad|iPhone|iPod/.test(navigator.platform) ||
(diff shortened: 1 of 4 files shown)


smalluban:
@philipp-classen Thanks for the thumbs up, but the approved review would be helpful too :)

extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,24 +218,28 @@ const reporting = new Reporting({
...
-        'Failed to initialize reporting. Leaving the module disabled and continue.',
-        e,
-      );
-    });
-  } else {
-    reporting.unload();
-  }
-});
+const setup = asyncSetup([
+  observe('terms', (terms) => {
```
philipp-classen, line 222: Here, we need to wait for the initialization of reporting (please add async + await):

```
async (terms) => {
  if (terms) {
    await reporting.init().catch(...)
```

extension-manifest-v3/src/background/reporting.js
```diff
@@ -216,24 +218,28 @@ const reporting = new Reporting({
...
+  }),
+]);
 
 function delay(timeInMs) {
   return new Promise((resolve) => setTimeout(resolve, timeInMs));
 }
 
-function onLocationChange(details) {
+async function onLocationChange(details) {
+  setup.pending && (await setup.pending);
```
philipp-classen, line 241: For clearer error messages (avoiding unhandled rejections if startup timed out or failed):

```
try {
} catch (e) {
  console.warn('reporting is unavailable:', e);
  return;
}
```

extension-manifest-v3/src/background/utils/setup.js
```diff
@@ -0,0 +1,32 @@
...
+        }, threshold);
+      }),
+    ]).then((...args) => {
+      result.pending = false;
+      clearTimeout(timeoutId);
+
+      return args;
+    }),
+  };
+
```
philipp-classen, line 30: I would suggest to have a catcher statement if the promise fails (timeout or error) and warns.

```
  result.pending.catch((e) => {
    console.warn('Error during setup', e);
  });
  ```
  
  Then you get a proper stacktrace that points to the original caller:
  
  
better-error

It also avoids a rejected promise warning, since for instance, in reporting only the handler would eventually handle the message. But if there are no event, it will not run quickly enough and Chrome thinks it is unrejected and gives you a potentially misleading error.

smalluban:
@philipp-classen I discovered, that you have passed the async function directly to the listener, which is anty-pattern, and also we have async IIFE inside of an already async function. I cleared it up.

philipp-classen (approved):

philipp-classen:
> @philipp-classen I discovered, that you have passed the async function directly to the listener, which is anty-pattern, and also we have async IIFE inside of an already async function. I cleared it up.

"function onLocationChange" wasn't originally declared async (before this PR). But the final version (after the cleanup) makes sense.

PR #1161, opened 2023-05-17
Add sync options for logged in users

### LINKED ISSUE #1012: Settings sync

### LINKED ISSUE #1166: Ghostery settings: account tab
In the new account tab we want to show:
* Link to sign in or Account details if already logged in + log out link
* Link to become a contributor or info about Contributor status
* Toggle of Settings sync
* Settings sync status (last update)

smalluban:
Fixes #1012. Fixes #1166 

TBA


smalluban:
I discovered that we might merge not sync server options with currently set local options, so only actually changed keys are sent to the server, the rest is taken from the server.

The last commit moves sync logic into the `get` and `set` store methods, as we need a list of changed keys.

With this change if on one browser option A is changed, and then (before the sync) on another browser option B is changed, the second browser after the change will not overwrite the changed A option, but it will merge them.

The only problem now is that the main features are nested in `options.engine` object, but the store provides a list of root keys of the model, so those options are overwritten. We use that nested list of options as a list of adblocker engines, but we could hardcode that instead and move options to the root (it will require a data migration).

extension-manifest-v3/src/store/options.js
```diff
@@ -37,35 +49,66 @@ const Options = {
...
       return options;
     },
-    async set(_, options) {
-      if (options === null) options = {};
+    async set(_, options = {}, keys) {
+      if (!keys.includes('revision')) {
+        options = Object.assign({}, options, {
+          revision:
+            options.terms && options.sync && (await store.resolve(Session)).user
+              ? Date.now()
```
philipp-classen, line 83: Real time is notoriously difficult in distributed systems. An approach through counters is resilient to the situation that the local clock time is wrong.

The idea is roughly to keep a counter and increment it locally before sending. And when receiving (from the server), update it to the maximum of the local counter and the counter from the server. https://en.wikipedia.org/wiki/Lamport_timestamp

The algorithm needs also something to break ties. Not 100% sure if we need it (since the server ultimately keeps the state and can break ties by preferring the first received message). But if needed, the extension could create a random ID on startup before sending the first value. It doesn't need to be persisted; if the service worker dies, it can create a new random one.

smalluban: Thanks for your propositions. It looks like the counter thing might be tempting, but I think it is rather easy to find an edge case. 

Browser A (offline) updates multiple times, for example from 1 to 5.
Then Browser B (online) has an initial state of 1, and it updates once to 2.

The server after that events has state 2, with the newest value (correct one). Then browser A becomes online, so by the internal check (chrome.alarms), in the next hour fetches data from server, and.. it seems that local state 5 is newer than state 2 on the server, so it overwrites the server with values and state 5, which is wrong...

philipp-classen: Note that the mechanism would not change. That means collision detection and handle would be identical. In both cases, you need a timestamp to decide if an action took place before another. Using local system clock is a timestamp, but has a chance to introduce errors (if two devices are online at the same time, for instance). Idea like Lamport's timestamp are only another form of getting a timestamp.

But I think, in the scenario that you described, it is not avoidable to lose update (if they update the same fields): if two devices make changes without syncing with the server and (if merging is not possible) the only solution that I see is to decide which update came first and than reject the conflicting update later.

Let's look at the scenario and compare how clocks make a difference:

9:13, t=0: Browser A and Browser B started with a consistent state that they got from the server. Let's say assume it is 9:13 o'clock and the Lamport timestamp is 0.  

Browser A:
9:17, t=0: updates one field
9:17, t=1: (clock moved forward because of the change)
9:18, t=1: (the device is offline and unable to sync with the server)
9:23, t=1: updates another field
9:23, t=1: (clock doesn't move forward, only the internal dirty state changes)
... (more updates, but it always stays at t=1)

(in the meantime on Browser B)
10:17, t=0: updates one field
10:17, t=1: (clock moves forward)
10:18, t=1: able to sync with the server (the server now updates its database and the time progresses to t=2)
10:18, t=1: sees the timestamp from the server in the response (t=2).
10:18, t=2: (clock moves forward since the server clock is higher than the local clock)

(back to Browser A):
10:30, t=1: finally able to sync all changes with the server: the server sees that the changes are from the past. Now it is difficult because it is a collision: either it can be merged or not. If it can't be merged, then Browser A loses all updates and should get the latest state from the server. Or it can be merged.

But note that the consistent timestamp makes it at least possible to detect it reliably independent if the local clock is wrong. Also, note that in the last update at 10:30, the relevant (local) timestamp would have been rather 9:17 (not 10:30) because that was the last time went the client sync successfully. It's like Git commits: it matters from where you forked. And counters are a safer way to pinpoint it than using local clock time (like it is safer to say, I forked from commit XYZ than saying I forked at 10 PM yesterday).

smalluban: In my implementation, the revision timestamp is updated locally (regardless of the connection with the server). 

In your example, Browser A could update after 10:18 (offline), so it still has t=1, but the update is "the latest". Then when online, it will be overwritten by the server at 10:30 (from your example).

As I think there is no sense (in that flat few options changing rarely) to creating an event system with a log of all changes (more or less a git system) - **I strongly opt-in to make the newest (in time) change as a source of truth** - without a time, you cannot do it. A counter mechanism is not time aware.

smalluban: > With that a counter cannot be used to decide the order of events 

And that's the opposite of my idea - the "newest" in time change should have priority, as it is the most expected from the user.

> With that a counter cannot be used to decide the order of events (which happened before the other), I disagree.

Please explain how it would work with the example, I showed above (that the last change is lost with the counter).

The only drawback of a timestamp is that it might be wrong in a super edge case, which in my opinion is negligible. This is not an API, which syncs every second or minute.. as users rather set the extension and forget about it. Even if we take into account passed domains (in the future).

So.. if we agree that timestamp is fine, we have an advantage, in that it can be used for what I empathized:

>  I strongly opt-in to make the newest (in time) change as a source of truth

If you look closely at the implementation, it is even smarter, as if a server has a newer version than a local one, which is tried to push to the server, we only overwrite changed keys from the local version to the server (as currently sent local version is always "the newest"), so indeed we have some merge mechanism, if there is a conflict.

philipp-classen: I think, we are mixed here three things that play together but are separate concerns:
* Timestamp (logical stamps vs local clock)
* Conflict detection: how to detect a conflict
* Merging: an attempt to automatically resolve a conflict

I was originally only commenting that for conflict detection, a logical timestamp would be sufficient. Since it is only used to decide what "happened-before". But I think, the crucial part of the approaches is rather who is the source of truth and how merging is handled.

> > With that a counter cannot be used to decide the order of events
> 
> And that's the opposite of my idea - the "newest" in time change should have priority, as it is the most expected from the user.

That is now conflict resolution. The idea is to merge on the client, I understand. It is possible to do that, though my intuition would have that the server would be the only source of truth.

I haven't though a lot about merging to be honest. I was rather assuming to sync changes like most REST Apis do (fetch from the server, update locally, and send the updated state back). On conflicts, you are out of luck and have to throw them away.

Typically you have internet connection and can sync quickly after each change of a sync property, which hopefully makes conflicts unlikely enough that you can get away with it. For instance, before flipping a synced pref, first fetch the server state to be in sync. Then modify the pref and send the whole state back it to the server. (Some REST APIs try to improve by providing PATCH operations to require only one request. But they are harder to implement on the server.)

Merging on client side is possible. You end up with an almost p2p-like system, where clients manage the state and only use the server to share their state with the other clients. There is no well-defined leader and the last client that sent defines what a new extension installations would see.

If it works, it is nice. I find it only a bit hard to reason about; but maybe it works well for our use case.

(For instance, for me it is not trivial for me to see that if  multiple extensions are being logged in that the system is eventually consistent. That means, after waiting long enough - not changing anything and all are able to reach the server - if they would always end up with the same state of the sync parameters. If the database of the server is the only source of truth, it's easier to see that updates will eventually reach all clients.

Also, error recovery. If clients make changes in their local databases but fail to sync, will they eventually resolve to the same state once they can reach the server again? Maybe yes, but for me, it's not trivial to reason about that either.)

smalluban: We need to face it, that the only problem occurs with updates to the options, which both try to sync, but fail. Timestamp gives information about, which is "latest", so we can assume that this is what the user wants. 

Think of that... we are talking about a single user with multiple devices (is rather rare that it would be browsers on the same device). How often that it might be an imagined situation that we try to figure out? If a user tried to update options on mobile, but it failed and looks at the computer and sees no change, it will update it there, so then he thinks that this is the source of truth. Nothing complicated. 

About the source of truth, it is indeed the server. If it has a newer revision it overwrites the local version. And whenever options are get first check if the server has a newer version.

philipp-classen: Then why not keep it like the old extension did it? It fetches state from the server, overwrites its local state, and on each change, updates the server state (and retry a couple of time if it fails).

To me that would be the straightforward solution under the assumptions.

For instance, flipping a pref that is synced:
* fetch the state from the server
* update it locally (flip the pref)
* send the new state to the server

From time to time poll for changes.

(BTW, I find multiple extension running on the same device not unrealistic. For instance, people may use Firefox and Chrome with Ghostery on the same machine. I'm not aware of bug reports for the old extension regarding the syncing.)

smalluban: The state is empty on the first hit on the server, and you may not make any changes to options after logging in. The idea to fetch and save data from the server is technically fine, but I wanted to avoid setting local options when it is not required - then we have to be careful to avoid a loop of setting the local version and fetching it from the server. That's why I introduced revision. It is just like you wrote but with a check of the time of the last update.

We have a limited list of options, which only a part could be migrated, so when the user will update extension MV2 to MV3 options will be migrated locally, so when the user will be logged in **For the first time** it will send the current state to the new API end-point. It means, that migration is covered.

smalluban:
@chrmod @philipp-classen Regardless of the sync algorithm, I think the more important question is if we are fine with the nested `engines` option, which is overwritten when set. 

The only case when it set it "wrong", is when one browser sets engines and another does not yet sync the change, so the user **must open the settings page before the other change**.  I think it is quite rare, as every time a user opens a panel or settings page, options are synced.

For me it's fine, but you may prefer to flatten engine options. The only drawback is to set explicitly names of engines for different usage, as for now it is taken from `options.engines` keys.

extension-manifest-v3/src/store/options.js
```diff
@@ -76,21 +119,20 @@ const Options = {
...
+        .catch(() => {});
+
       return options;
     },
-    observe: (_, options) => {
-      observers.forEach((fn) => fn(options));
+    observe: (_, options, prevOptions) => {
+      observers.forEach((fn) => {
+        try {
+          fn(options, prevOptions);
```
chrmod, line 131: should this be awaited?

extension-manifest-v3/src/store/options.js
```diff
@@ -76,21 +119,20 @@ const Options = {
...
-        });
-      } catch (e) {
-        console.error(
-          `Error while sending update options to other contexts: `,
-          e,
-        );
-      }
+      chrome.runtime
+        .sendMessage({ action: UPDATE_OPTIONS_ACTION_NAME })
+        .catch(() => {});
```
chrmod, line 124: please add a log

chrmod:
regarding the nested `engines` and all sync options in general, it may be beneficent to keep all of them flat for better readability and querying on the server side. 

Additionally, in upcoming month we will likely introduce support for optional block lists, like regional filters or annoyances. One natural place to configure those could be the `engines` property, but then we again have problems with clients from different versions overwriting each other. 

So it looks like that we should never change the meaning of once introduced synced property and by that it may be better to avoid **complex** values.

chrmod (changes requested):

extension-manifest-v3/src/store/options.js
smalluban: It should not. Observers are not related to each other and should be called as soon as possible. 

Moreover, this PR does not change a logic here, just moves `try {} catch {}` from the wrapper into this method.

extension-manifest-v3/src/store/options.js
smalluban: The function only `throws` if the other end does not exist. Mainly, when the background process starts, but there is another content script or extension page, which could receive a message. This error is annoying (as the behavior is expected). The log message would introduce again something that I wanted to avoid.

extension-manifest-v3/src/store/options.js
smalluban: I'll add it.

extension-manifest-v3/src/store/options.js
chrmod: Got it. Please leave a comment with an explanation

smalluban:
@philipp-classen I resolved our conversation to limit noise in review - the subject is still valid, and if we want to address it somehow, I suggest a meeting with the three of us.

@chrmod I made a migration of flattened main options, moving nested `options.engines` to root options. 

The last commit also includes new UI in settings.

extension-manifest-v3/src/background/session.js
```diff
@@ -0,0 +1,71 @@
...
+// Trigger options sync every hour
+const ALARM_SYNC_OPTIONS = 'session:sync:options';
+const ALARM_SYNC_OPTIONS_RATE = 1 * 60; // 1 hour in minutes
+
+// Try to check cookies every 30 days if the user is still logged in
+const ALARM_UPDATE_SESSION = 'session:update';
+const ALARM_UPDATE_SESSION_DELAY = 1000 * 60 * 24 * 30; // 30 days in milliseconds
+
+// Observe cookie changes (login/logout actions)
+chrome.cookies.onChanged.addListener(async ({ cookie, removed }) => {
```
chrmod, line 27: I've heard that this listener is not very reliable - need to be tested in all browser. Most importantly Firefox and Safari.

extension-manifest-v3/src/pages/onboarding/index.js
```diff
@@ -19,10 +19,9 @@ async function updateOptions(host, event) {
   const success = event.type === 'success';
 
   await store.set(Options, {
-    engines: Object.keys(Options.engines).reduce(
-      (all, rule) => ({ ...all, [rule]: success }),
-      {},
-    ),
+    ads: success,
```
chrmod, line 22: how about we make those names more description, eg. `protectionAds` or `featureAdBlocking`, or maybe `Ad-Blocking`, `Anti-Tracking` `Never-Consent`?

extension-manifest-v3/src/store/options.js
```diff
@@ -167,19 +234,19 @@ async function migrateFromMV2() {
...
 export async function observe(property, fn) {
-  let value;
-  const wrapper = async (options) => {
-    if (value === undefined || options[property] !== value) {
-      const prevValue = value;
-      value = options[property];
-
-      try {
+  let wrapper;
+  if (property) {
```
chrmod, line 238: we really need unit tests for this one - it is impossible to tell how it works or what we change. This is not a critic of the change, I simply incapable to judge its effect without use cases being listed

chrmod (commented):
Looks very well. Lets sync on the comments.

extension-manifest-v3/src/background/session.js
smalluban: I'll check that.

extension-manifest-v3/src/pages/onboarding/index.js
smalluban: `feature` prefix - all options are a kind of feature. I would avoid using dashes in names, as you are forced to use `options["Ad-Blocking"] in the code.

The `engines` scope was fine. The only reason to move it to the top is this sync feature. If there is a need for the indicator, might be something, which is related to the logic, like `block` ? `blockAds`, `blockTrackers`, `blockAnnoyances` ?

extension-manifest-v3/src/store/options.js
smalluban: I am thinking of making API like this inside of the `hybrids`, so it might not be needed here anymore. I would postpone it to a separate issue.

smalluban:
@chrmod Good catch with testing in Firefox/Safari. In Firefox it worked well, but in Safari, even though there is the `chrome.cookies.onChanged`, it does not work... (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged#browser_compatibility)

After a while, I figured out the idea to watch the navigation to account / sign on, as this is the only why to login/logout. Changes in the last commit.

smalluban:
@philipp-classen @chrmod Let's make a final round of reviews, and decide what we do with the feature. I made all of the required changes. If there is still doubt in sync implementation (a local time, etc..), please consider a meeting to work it through.

chrmod (dismissed):

smalluban:
Let's wait to merge for @philipp-classen

extension-manifest-v3/src/utils/shims.js
```diff
@@ -22,6 +22,8 @@ function shimAlarms() {
     },
     async clear(name) {
       clearTimeout(alarms.get(name));
+      clearInterval(alarms.get(name));
```
philipp-classen, line 25: clearTimeout and clearInterval for the same value. I assume, there should be only one statement.

philipp-classen (changes requested):

extension-manifest-v3/src/utils/shims.js
smalluban: It's intentional, it might be a timeout or interval. See it here and there:

timeout - https://github.com/ghostery/ghostery-extension/pull/1161/files#diff-4484ec0f55d82abb59d1c3a6bb7539a7290569072aa1a99a8c1bd66a0be70ceaR44

interval - https://github.com/ghostery/ghostery-extension/pull/1161/files#diff-4484ec0f55d82abb59d1c3a6bb7539a7290569072aa1a99a8c1bd66a0be70ceaR50

philipp-classen: https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout

> It's worth noting that the pool of IDs used by [setTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) and [setInterval()](https://developer.mozilla.org/en-US/docs/Web/API/setInterval) are shared, which means you can technically use clearTimeout() and [clearInterval()](https://developer.mozilla.org/en-US/docs/Web/API/clearInterval) interchangeably. However, for clarity, you should avoid doing so.

As long as browsers implement it that way, it will work (the same as if you call clearTimeout twice). But it relies on that property that the pool is shared. (If an implementation would different pools of counters if would not be correct anymore.)

If possible I would avoid it, or if absolutely needed, rather only leave one call and add a comment that it assume that the JavaScript engine uses a shared pool (which all current ones, I do I think).

smalluban: The current implementation uses both for periodical alarms. Of course, it is possible to set `timeoutId` and `intervalId` separately, but it will overcomplicate the code. I choose a comment in that situation. 

BTW there was missing clear out of the interval in `clearAll()`, so I made a fix.

smalluban:
@chrmod @philipp-classen 

I made another round of refactoring, including:
* non-blocking sync in `observe` and `set` methods
* changed a revision impl. to use a counter instead of a local timestamp

Everything works as expected. However, for the `counter` there is one downside (over the timestamp), that is easier to make a non-consistent state between browsers, which will have the same revision number (and cannot be easily fixed):

As the counter bumps in every change, if you have opened two browsers with synced state, and the user will update both before they sync periodically, they will send and set two different states as the same counter value. It's not a big deal, but they will never update the local state until the server bumps )(another change), as they think that their state is fine. 

This is related to **non-blocking** sync on set, as it only syncs while the settings are set (but does not take into account the value from the server, and it can't as then it would have to await the result from the server).

The above problem can be simplified to the question of what is more important: the time of saving the settings, or bumping properly the counter. If we don't want to wait for the server, then the implementation is fine and can be merged.

philipp-classen:
> As the counter bumps in every change, if you have opened two browsers with synced state, and the user will update both before they sync periodically, they will send and set two different states as the same counter value. It's not a big deal, but they will never update the local state until the server bumps )(another change), as they think that their state is fine.

I also think that client behavior is sound. The server should indeed increase the timer on each update that it accepts. Then it is guaranteed that if clients keep polling, and they will eventually end up with the same state (and same as the server). (If we haven't implemented it on the server, we should, so we can test the client under realistic circumstances.)

extension-manifest-v3/src/utils/api.js
```diff
@@ -11,14 +11,20 @@
 
 import jwtDecode from 'jwt-decode';
 
-const DOMAIN = 'ghostery.com';
+const DOMAIN = 'ghosterystage.com';
```
philipp-classen, line 14: Reminder: before we merge, we should change it (and the production server should be updated).

extension-manifest-v3/src/store/options.js
```diff
@@ -75,40 +118,80 @@ const Options = {
...
+
+        options = await store.set(Options, {
+          ...values,
+          revision: serverOptions.revision + (keys ? 0 : 1),
+        });
+      }
+
+      // TIP: Keys are only passed when the options are set
+      if (serverOptions.revision < revision || keys) {
+        await setUserOptions(
```
philipp-classen, line 174: With discussed protocol changes, the server will increase the version if it accepts the change. Will depend how the server implements it, but the client will see from the result whether it was accepted or not. If it was accepted, the client should also update its local version, since it would otherwise be behind the server and it would not be able to update till the next sync.

(One idea discussed is that we change the server to return 202 (ACCEPTED) if it updated the state. In that case, the version will be one higher than the client. And if it rejected, 409 (CONFLICT). In that case, the client should not update its local version. But it may be a good time to poll for the new changes in that situation.)

extension-manifest-v3/src/utils/api.js
smalluban: This is not intentional, I was undoing that change before committing. I'll fix it.

extension-manifest-v3/src/store/options.js
philipp-classen: Let's consider the situation when a client has local changes, but knows that its version is outdated. In other words, the user changed some other settings in the meantime. These other changes have also have propagated to other clients. 

What to do with the local changes now? In the end, it is a UI question, but these are the options that I can see:

1. Update to the latest version and discard the local changes (with or without visual feedback)
2. Stay on the local version (with or without visual feedback, something like "out of sync, syncing is now disabled). Optionally ask the user to resolved it (e.g. turn off sync forever, force a sync to the current version and discard local changes, or force-push the local changes and lose the updates on the other devices)
3. By default, force-push changes (in other words, ignore the conflict, with or without visual feedback). The drawback is that the other client will now lose the changes (in some sense, we would be again but to the same problem)
4. Resolve the conflict without user interaction and push the merged version as the new state.

For me personally, option 1 (without any visual feedback) would be OK. Optionally, having a warning that the client is not able to sync is nice. For instance, warn if the client has local changes, but fails to sync them to the server.

Staying on the old version is possible (option 2), but then you are stuck in that state and syncing will be disabled. The use can still change settings locally; but unless the client gets in sync again, these changes will not reach other clients. We could ask the user how to proceed (discard or force-push) to enable it again. On the plus side, it would give the user more control.

Always automatically force-pushing (option 3) creates more problems that it solves IMHO. It turns off optimistic locking and switches to the model as we have in MV2 Ghostery.

Option 4 might be the best user experience if it works. But it is not easy to implemented correctly. Though the extension could try to merge and if it fails, fall back to one of the other strategies.

(Personally, I'm leaning towards option 1 and would hope that it doesn't happen too often.)

smalluban:
I'll make another PR for the clear spot on the review - as this become too long.

PR #1162, opened 2023-05-17
Use local implementation for delaying icon update

### LINKED ISSUE #977: Remove dependency `lodash-es`
In MV3 we only use `throttle` once in here https://github.com/ghostery/ghostery-extension/blob/455a51f087084baa7885557c00513ee2cf1c2568/extension-manifest-v3/src/background/tab-stats.js#L20

If we replace it with something simple like:
```js
const throttle = (fn, delay) => {
  let lastCalled = 0;
  return (...args) => {
    let now = Date.now();
    if(now - lastCalled < delay) {
      return;
    }
    lastCalled = now;
    return fn(...args);
  }
}
```
we could remove `loadsh` from dependency list.

smalluban:
Fixes #977 

Hits update on both edges (to be fast but also to update changes during the delay).

It looks like it is even better in another context, before the change, the background process cleared updates of the icon in not the "newest" tab - as we used only one throttle for all tabs.

### DIFF
--- a/extension-manifest-v3/package.json
+++ b/extension-manifest-v3/package.json
@@ -43,7 +43,6 @@
     "@whotracksme/webextension-packages": "^2.1.5",
     "idb": "^7.1.1",
     "jwt-decode": "^3.1.2",
-    "lodash-es": "^4.17.21",
     "tldts-experimental": "^5.7.66"
   },
   "webExt": {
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/package.json
```diff
@@ -43,7 +43,6 @@
     "@whotracksme/webextension-packages": "^2.1.5",
     "idb": "^7.1.1",
     "jwt-decode": "^3.1.2",
-    "lodash-es": "^4.17.21",
```
chrmod, line 46: :heart:

chrmod (approved):

PR #1173, opened 2023-05-25
Minor fixes for Firefox/Ghostery Browser 

smalluban:
Minor fixes for UI/navigation for Firefox Andoird / Ghostery Browser for Android.

Highlights:
* Where possible I added a click listener for external links, and used `chrome.tabs.create` API
* Settings page is broken on Firefox if opened in a new tab (there is no way to fix it using data from the browser) - it's a bug, reported here: https://bugzilla.mozilla.org/show_bug.cgi?id=1813882. They should add `env(safe-area-inset-bottom)`, but it returns `0`
* Closing the window from navigated page (not the first in history) in the "popup" window does not work - the tab opens in the background, so in the settings page click on account / log in must stay in the popup

To test it on Firefox Android, build the extension with `npm run build firefox` and use `web-ext`:

```
npx web-ext run --target=firefox-android --android-device=XXXX
```

### DIFF
--- a/extension-manifest-v2/webpack.config.js
+++ b/extension-manifest-v2/webpack.config.js
@@ -110,8 +110,6 @@ module.exports = {
 			onBuildEnd: {
 				scripts: [
 					`${RM} ./dist/foundation.js`,
-					`${RM} ./dist/panel.js`,
-					`${RM} ./dist/panel_android.js`,
 					`${RM} ./dist/purplebox_styles.js`,
 					'npm run licenses',
 				],
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -351,7 +351,7 @@ if (__PLATFORM__ === 'firefox') {
         );
       }
 
-      if (details.frameAncestors.length > 0) {
+      if (details.frameAncestors?.length > 0) {
         for (const { url } of details.frameAncestors) {
           const { domain, hostname } = parse(url);
           if (pausedDomains.includes(domain || hostname)) {
@@ -402,7 +402,7 @@ if (__PLATFORM__ === 'firefox') {
     (details) => {
       const request = Request.fromRequestDetails(details);
 
-      if (details.frameAncestors.length > 0) {
+      if (details.frameAncestors?.length > 0) {
         for (const { url } of details.frameAncestors) {
           const { domain, hostname } = parse(url);
           if (pausedDomains.includes(domain || hostname)) {
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -351,6 +351,15 @@ if (__PLATFORM__ === 'firefox') {
         );
       }
 
+      if (details.frameAncestors.length > 0) {
+        for (const { url } of details.frameAncestors) {
+          const { domain, hostname } = parse(url);
+          if (pausedDomains.includes(domain || hostname)) {
+            return;
+          }
+        }
+      }
+
       if (
         pausedDomains.includes(request.sourceDomain || request.sourceHostname)
       ) {
@@ -392,6 +401,15 @@ if (__PLATFORM__ === 'firefox') {
   chrome.webRequest.onHeadersReceived.addListener(
     (details) => {
       const request = Request.fromRequestDetails(details);
+
+      if (details.frameAncestors.length > 0) {
+        for (const { url } of details.frameAncestors) {
+          const { domain, hostname } = parse(url);
+          if (pausedDomains.includes(domain || hostname)) {
+            return;
+          }
+        }
+      }
 
       if (pausedDomains.includes(request.sourceDomain)) {
         return {};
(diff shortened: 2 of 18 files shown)


extension-manifest-v3/src/utils/tabs.js
```diff
@@ -0,0 +1,19 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+export function openTabWithUrl(host, event) {
+  const { href } = event.currentTarget;
+  event.preventDefault();
+
+  Promise.resolve(chrome.tabs.create({ url: href })).then(() => {
```
chrmod, line 16: why `Promise.resolve`?

ui/src/modules/panel/components/stats.js
```diff
@@ -11,6 +11,17 @@
 
 import { html, msg, router, dispatch } from 'hybrids';
 
+function openTabWithUrl(host, event) {
```
chrmod, line 14: isn't this redundant?

chrmod (changes requested):

extension-manifest-v3/src/utils/tabs.js
smalluban: In MV2 `chrome.tabs.create` does not return a promise (in our case Firefox target).

ui/src/modules/panel/components/stats.js
smalluban: Yes, this is intentional. This code is repeated in separate packages. It would be placed somewhere in `ui` utils and exported if shared. I thought that it is not worth it.

extension-manifest-v3/src/utils/tabs.js
smalluban: It might not be required, as we shim `chrome` API in Firefox, using `browser.`. I'll check that.

extension-manifest-v2/webpack.config.js
```diff
@@ -110,8 +110,6 @@ module.exports = {
 			onBuildEnd: {
 				scripts: [
 					`${RM} ./dist/foundation.js`,
-					`${RM} ./dist/panel.js`,
```
chrmod, line 113: why this is removed?

extension-manifest-v3/src/utils/tabs.js
chrmod: and how is it? :)

smalluban: It looks, like we can omit `Promise.resolve()`

smalluban:
@chrmod Rebased with current `main` and fixed `Promise.resolve()`. You can have a final look.

extension-manifest-v3/src/pages/panel/index.js
```diff
@@ -66,6 +66,6 @@ if (__PLATFORM__ === 'firefox') {
     let el = event.target;
     while (el && !el.href) el = el.parentElement;
 
-    if (el?.target) requestIdleCallback(window.close);
+    if (el?.target) setTimeout(window.close, 100);
```
chrmod, line 69: can you please put a comment here on why such timeout is needed

extension-manifest-v3/src/pages/panel/views/tracker-details.js
```diff
@@ -105,7 +107,11 @@ export default {
                 underline
                 layout="padding margin:-1"
               >
-                <a href="${tracker.website}" target="_blank">
+                <a
+                  href="${tracker.website}"
+                  target="_blank"
```
chrmod, line 112: with `onclick` the `target` is likely not needed.

ui/src/modules/settings/components/toggle.js
```diff
@@ -89,6 +89,6 @@ export default {
         }
       }
     `,
-    { deleteFocus: true },
+    { delegateFocus: true },
```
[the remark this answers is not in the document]
smalluban: Yes, I accidentally found it.

extension-manifest-v3/src/background/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
+  switch (msg.action) {
+    case 'getCurrentTab':
```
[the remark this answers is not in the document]
smalluban: Firstly, I also moved here a message receiver for opening a new tab, so there were more of them. 

I think a switch is clearer here (as we can introduce more of the helper messages). For now, it could be if/else, but the switch does not introduce any perf issue, so I would keep it.

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -393,6 +402,15 @@ if (__PLATFORM__ === 'firefox') {
     (details) => {
       const request = Request.fromRequestDetails(details);
 
+      if (details.frameAncestors?.length > 0) {
```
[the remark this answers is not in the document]
smalluban: Fixed

chrmod (approved):

PR #1174, opened 2023-05-26
Fix errors that can happen if tabs are closed

philipp-classen:
Fixes some situations where closed tabs are not handled. Or convert error to warnings, since it is expected that tabs are being closed and attempts like injecting CSS will fail.

### DIFF
--- a/extension-manifest-v3/src/background/stats.js
+++ b/extension-manifest-v3/src/background/stats.js
@@ -40,8 +40,9 @@ observe('terms', async (terms) => {
 
 async function refreshIcon(tabId) {
   const stats = tabStats.get(tabId);
-  const options = await store.resolve(Options);
+  if (!stats) return;
 
+  const options = await store.resolve(Options);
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/src/background/stats.js
```diff
@@ -57,11 +58,18 @@ async function refreshIcon(tabId) {
...
       );
     }
-    try {
-      await chrome.action.setIcon({ tabId, ...data });
-    } catch (e) {
-      console.error('Error while trying update the icon:', e);
-    }
+
+    // Note: Even in MV3, this is not (yet) returning a promise.
+    chrome.action.setIcon({ tabId, ...data }, () => {
```
chrmod, line 63: should we still await it? how about using `webextension-polyfill`? 

Below there is a call to `setBadgeText` - doesn't it have the same problem?

chrmod (approved):

PR #1183, opened 2023-06-02
Remove Github templates

chrmod:
Those were not really commonly used. 

Alternatively we could consider using issue forms or links. Important one would be to link to Broken Page Reports.


philipp-classen (approved):

philipp-classen:
I also tend to ignore the templates. I'm OK with removing, but I see the value when it comes to external contributors.

chrmod:
@yokoffing is experimenting with issue forms in the Broken Page Reports repo, I think we can use that instead of the template. Lets try. Merging for now.

PR #1186, opened 2023-06-06
MV2: ensure panel and browser badge consistency

chrmod:


extension-manifest-v2/src/classes/FoundBugs.js
```diff
@@ -309,20 +310,27 @@ class FoundBugs {
...
-				allowed: 0
+				allowed: 0,
+				found: 0,
 			};
 		}
 
 		const { blocked } = this._foundApps[tab_id].issueCounts;
 		const allowed = this._foundApps[tab_id].apps.length - blocked;
 
+		const adBlockingTrackers = getCommonData(tab_id, tabUrl).unidentifiedTrackers;
```
philipp-classen, line 325: Would feel more natural to explicitly list the third argument: 

const adBlockingTrackers = getCommonData(tab_id, tabUrl, false).unidentifiedTrackers;
const antiTrackingTrackers = getCommonData(tab_id, tabUrl, true).unidentifiedTrackers;

Took me first a while to understand what was the difference been both getCommonData calls.

chrmod: make sense. Fixed!

philipp-classen (approved):

PR #1197, opened 2023-06-12
MV2: fix unidentified category

chrmod:
This PR aim to fix problem with unidentified category trackers, most notably:
* once whitelisted trackers did not appear in the panel
* whitelisting did not cover eTLD+ 
* whitelisting was exact page hostname matched
* unidentified trackers had random order in the panel

### DIFF
--- a/extension-manifest-v2/app/panel/reducers/blocking.js
+++ b/extension-manifest-v2/app/panel/reducers/blocking.js
@@ -70,9 +70,15 @@ function mergeTrackers(adblockerTrackers, antiTrackingTrackers, whitelist, pageU
 			all.set(tracker.name, existing);
 		}
 	});
-	const findTrackerByWhitelistedDomain = domain => Array.from(all.values()).find(
-		tracker => !!tracker.domains.some(whitelistedDomain => whitelistedDomain.endsWith(domain))
-	);
+
+	const findTrackerByWhitelistedDomain = (domain) => {
+		for (const tracker of all.values()) {
+			if (tracker.domains.some(whitelistedDomain => whitelistedDomain.endsWith(domain))) {
+				return tracker;
+			}
+		}
+		return null;
+	};
 
 	const pageTld = parse(pageUrl).domain;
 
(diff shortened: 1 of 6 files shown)


GRadziejewski:
Changes were checked. All tests are finished. It's working according to the acceptance criterias.

extension-manifest-v2/app/panel/reducers/blocking.js
```diff
@@ -69,6 +70,31 @@ function mergeTrackers(adblockerTrackers, antiTrackingTrackers) {
 			all.set(tracker.name, existing);
 		}
 	});
+	const findTrackerByWhitelistedDomain = domain => Array.from(all.values()).find(
```
philipp-classen, line 73: I wonder if this may become a problem if the whitelist grows. There are three nested loops here:
* For each whitelist entry, it gets called
* Then it iterates over all trackers
* Then it iterates over all tracker domains

I don't see an obvious way to optimize it though.

Extracting `Array.from(all.values())` (also used as the return value) would make sense (I don't know if the JIT can work that out).

extension-manifest-v2/app/panel/reducers/blocking.js
```diff
@@ -69,6 +70,31 @@ function mergeTrackers(adblockerTrackers, antiTrackingTrackers) {
...
+	);
+
+	const pageTld = parse(pageUrl).domain;
+
+	Object.keys(whitelist).forEach((domain) => {
+		if (!whitelist[domain].hosts.some(host => host === pageTld)) {
+			return;
+		}
+		const tld = parse(domain).domain;
+		const existing = findTrackerByWhitelistedDomain(tld) || all.get(tld);
```
philipp-classen, line 84:
```suggestion
		const existing = all.get(tld) || findTrackerByWhitelistedDomain(tld);
```

should be equivalent, right? Unless it is extremely unlikely that `all.get(tld)` will find something, then trying the single map look first (before going into the loops) would be good (people could configure a huge whitelist).

extension-manifest-v2/app/panel/reducers/blocking.js
chrmod: `Array.from(all.values())` cannot be reused as we modify it with every iteration. I've moved to For-Of loop instead.

philipp-classen (approved):

PR #1205, opened 2023-06-20
Renew onboarding

smalluban:
Fixes https://github.com/ghostery/journal/issues/116

### The feature highlights

* Added content script to the main bundle, so the popup is triggered if conditions are met
* Added background module, which initializes local storage data if conditions are met
* Added `renew` component in onboarding UI module
* Main onboarding screen updated with `renew` flag, which changes copy and `abort` action (without confirm dialog)

### Testing

1. Turn on `debug` mode in `manifest.json`, by adding `"debug": true`
2. Ensure the staging AB testing server returns `true` for `the "terms"` test for your browser
3. Install the extension and make a normal setup
4. Open DevTools on the background process (or any extension page), and run:
    `chrome.storage.local.set({ enable_human_web: false })`
5. Restart the browser (the background process must re-run)
6. Open any web page
7. You should see (only once per hour) a popup with a message about renewing the setup


extension-manifest-v2/app/content-scripts/renew.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+import { showIframe } from '@ghostery/ui/iframe';
+
+const HOUR_IN_MS = 1000 * 60 * 60;
+
+chrome.storage.local.get(['renew_setup'], ({ renew_setup }) => {
```
chrmod, line 15: this will fire for every page load. Would prefer to have it background triggered instead.

extension-manifest-v2/src/background.js
```diff
@@ -938,9 +939,11 @@ function onMessageHandler(request, sender, callback) {
  * @memberOf Background
  */
 function setupABTests() {
-	if (abtest.has('terms')) {
+	if (abtest.hasTest('terms')) {
 		tryOpenOnboarding({ force: true, period: ONE_WEEK_MSEC });
```
chrmod, line 943: this was just a stub, can be removed

extension-manifest-v2/src/modules/renew.js
```diff
@@ -0,0 +1,64 @@
+/**
```
chrmod, line 1: please move it to already existing folder. Otherwise packaging script will require an update.

chrmod (changes requested):
was it tested in Firefox?

extension-manifest-v2/app/content-scripts/renew.js
smalluban: Trigger the whole popup from the background might be tricky, as it would take an active tab, which might not be a HTTP(s) page - like a new tab page, settings, etc.. so then it won't be shown after the delay.

However, I agree that calling storage on each page load might not be the best idea. I will move it to the background with a cache, so the content script will trigger a message instead.

extension-manifest-v2/src/background.js
smalluban: Ok.

extension-manifest-v2/src/modules/renew.js
smalluban: I am not sure what you mean here. The `modules/` path has been already used before: https://github.com/ghostery/ghostery-extension/blob/main/extension-manifest-v2/src/modules/autoconsent.js

The file is bundled, so the `modules/` is not in the result code.

smalluban:
I retested in Firefox, made some fixes, so @chrmod please review again.

extension-manifest-v2/app/content-scripts/renew.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+import { showIframe } from '@ghostery/ui/iframe';
+
+const HOUR_IN_MS = 1000 * 60 * 60;
+
+chrome.runtime.sendMessage({ action: 'renew:setup' }).then((renew_setup) => {
```
chrmod, line 15: can we make it a push instead of a pull? additional message on every page load for sake of one time event is not great.

extension-manifest-v2/src/background.js
```diff
@@ -933,16 +934,6 @@ function onMessageHandler(request, sender, callback) {
 	return false;
 }
 
-/**
- * Configure A/B tests based on data fetched from the A/B server
- * @memberOf Background
- */
-function setupABTests() {
-	if (abtest.has('terms')) {
-		tryOpenOnboarding({ force: true, period: ONE_WEEK_MSEC });
```
chrmod, line 942: please remove unused `ONE_WEEK_MSEC`

chrmod (changes requested):

extension-manifest-v2/app/content-scripts/renew.js
smalluban: Unfortunately, I think we cannot. This works the same as never-consent, so...

If we would "push" a message from the background, who should be the receiver? Active tab? It can be a settings page or another from the browser (so there is no required content script). If we filter them out, then it can be a tab that is not active anymore...

If you now some trick, let me know ;)

I think I already explained that here: https://github.com/ghostery/ghostery-extension/pull/1205#discussion_r1237122111

chrmod: I can see would be to make a combination of mechanisms like follow:
* move triggering to the background - it will by time bound
* on trigger, create a new webNavigation listener that await a capable website (with URL that looks like a real page)
* on page found, send a message to the content script after X (2?) seconds 

The supposition is to show messages quickly after a page load.

smalluban: As this is MV2, we can add listener later, so i think it might work. I'll try to do it, and get it out.

smalluban:
@chrmod @philipp-classen Here you have the pull implementation. The only drawback, which I can see is if a user never riches an hour of browsing and closes the browser ;) so then the popup code never runs.

chrmod (approved):

PR #1208, opened 2023-06-21
MV3: Anti-tracking

chrmod:


smalluban:
Please fix MV3 issue first (from the CI test): 
```
"WebRequestPipeline" is not exported by "../node_modules/@whotracksme/webextension-packages/packages/reporting/src/index.js", imported by "src/background/reporting/index.js".
```

extension-manifest-v3/src/pages/panel/views/tracker-details.js
```diff
@@ -79,22 +112,9 @@ export default {
...
-                    html`
-                      <gh-panel-copy oncopy="${showCopyNotification}">
-                        ${blocked ? html`<s>${url}</s>` : url}
-                      </gh-panel-copy>
-                    `,
-                )}
-              </div>
-            </div>
-          </div>
+          ${requestList('URLs observed', 'shield', requestsObserved)}
```
smalluban, line 115: Use
```
msg`URLs observed`
```

Otherwise, it won't be translated. The same for the next lines.

extension-manifest-v3/src/pages/panel/views/tracker-details.js
```diff
@@ -42,7 +63,19 @@ export default {
     stats.trackers.find((t) => t.id === trackerId),
   wtmUrl: ({ tracker }) =>
     `https://www.whotracks.me/trackers/${tracker.id}.html`,
-  content: ({ tracker, wtmUrl }) => html`
+  requestsObserved: ({ tracker }) =>
```
smalluban, line 66: This should be moved to the store model definition - as the `Tracker` have `requestsBlockedCount` and `requestsModifiedCount`. If we need whole objects (not only counts) I would change it in the definition to be `requestsModified` etc...

extension-manifest-v3/src/store/tab-stats.js
```diff
@@ -26,6 +30,16 @@ const Tracker = {
 const Stats = {
   domain: '',
   trackers: [Tracker],
+  trackersBlockedCount: ({ trackers }) =>
```
smalluban, line 33: It is better to return `Tracker` instances here, as then you can use them somehow, and still easily take a count from it - `stats.trackersBlocked.length`

ui/src/modules/panel/components/stats.js
```diff
@@ -199,5 +204,9 @@ export default {
         `}
       </ui-panel-switch>
     </template>
+  `.css`
+    .tracker ui-icon {
```
smalluban, line 208: `ui-icon` supports `color` attribute.

extension-manifest-v3/src/pages/panel/views/home.js
```diff
@@ -242,5 +290,22 @@ export default {
         </section>
       `}
     </template>
+  `.css`
+    .stats {
+      background-color: #F0F2F7;
```
smalluban, line 295: Don't use direct colors.

extension-manifest-v3/src/pages/panel/views/home.js
```diff
@@ -242,5 +290,22 @@ export default {
...
+    }
+    .stats-number {
+      font-size: 24px;
```
smalluban, line 298: We are using only pre-defined fonts from our token database. Please don't make direct CSS.

smalluban (changes requested):

smalluban:
If you need "new" colors or fonts - they have to be putted here for the panel UI - https://github.com/ghostery/ghostery-extension/blob/mv3-firefox-first-party-isolation/ui/src/modules/panel/styles.css

extension-manifest-v3/src/pages/panel/views/home.js
```diff
@@ -242,5 +290,22 @@ export default {
         </section>
       `}
     </template>
+  `.css`
```
smalluban, line 293: Don't create CSS in views. It should be only in components. If you need to change visuals, change it in the corresponding component, or make a new one.

extension-manifest-v3/src/pages/panel/views/home.js
```diff
@@ -93,6 +93,31 @@ function setStatsType(host, event) {
...
+        </span>
+        <ui-icon
+          name="info"
+          color="gray-400"
+          layout="size:2"
+        ></ui-icon>
+      </ui-tooltip>
+    </div>
+  </section>
+  </section>
```
smalluban, line 118: Double closing tag.

extension-manifest-v3/src/pages/panel/views/home.js
```diff
@@ -93,6 +93,31 @@ function setStatsType(host, event) {
   store.set(host.options, { panel: { statsType: type } });
 }
 
+const trackerStatus = ({ count, title, description, icon, iconColor }) => html`
```
smalluban, line 96: This looks to me like a visual component, so please make a new one in `/ui/src/modules/panel/componens` (We keep them there for the possible sharing of the UI with MV2, but I think it might never happen, so eventually we can move it to he MV3 source)

smalluban (changes requested):

extension-manifest-v3/src/pages/panel/views/tracker-details.js
chrmod: I kept it here as it is the only client and wasn't sure if having it in a model would not cause unnecessary computation.

extension-manifest-v3/src/pages/panel/views/home.js
chrmod: why encapsulate something that is not shared?

extension-manifest-v3/src/pages/panel/views/tracker-details.js
smalluban: Getters in models are lazy - it means, that they are calculated and cached when they are first time called (for the current instance). It is safe to move it there, and I would keep data and logic separated as much as possible.

extension-manifest-v3/src/pages/panel/views/home.js
smalluban: There are a few separate things:

1. It should be a component, as it is then easier to debug (the function just disappears, when called). Try to avoid creating a nested render function. It then is obvious that it should be a component (with possible slots, so you don't have to pass all of those arguments).
2. Moving it to the `ui` - this is what I explained in the brackets - it could be in the `src/pages/panel/components` as well, but for now we keep panel components in the `ui` module. And there is another rule - one file for one component, flat structure, with auto-definition, so you can just create a file there with a new component.

extension-manifest-v3/src/store/tab-stats.js
```diff
@@ -14,7 +14,10 @@ import { store } from 'hybrids';
 const Tracker = {
   id: true,
   name: '',
-  requests: [{ url: '', blocked: false }],
+  requests: [{ url: '', blocked: false, modified: false }],
+  requestsBlocked: ({ requests }) => requests.filter((r) => Boolean(r.blocked)),
+  requestsModified: ({ requests }) =>
+    requests.filter((r) => Boolean(r.modified)),
```
smalluban, line 20: The `Boolean` function wrapper is not required here, and below, the filter function maps to boolean internally (by checking truthy/falsy value.

extension-manifest-v3/src/pages/panel/views/tracker-details.js
```diff
@@ -34,6 +34,27 @@ function showCopyNotification(host) {
   host.querySelector('#gh-panel-company-alerts').appendChild(wrapper);
 }
 
+const requestList = (name, icon, requests) =>
```
smalluban, line 37: For me this is again a visual component, so you can avoid nested function in the template.

smalluban (changes requested):

extension-manifest-v3/src/pages/panel/views/tracker-details.js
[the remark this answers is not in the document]
smalluban: Still computed property in the view. Please create `requestsObserved` in the model.

extension-manifest-v3/src/background/env.js
```diff
@@ -0,0 +1,22 @@
...
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+if (!chrome.webRequest) {
```
smalluban, line 12: As we are not in a rush, can we fix the issue in a final way, without monkey-patching the chrome API?

extension-manifest-v3/src/store/tab-stats.js
```diff
@@ -14,7 +14,9 @@ import { store } from 'hybrids';
 const Tracker = {
   id: true,
   name: '',
-  requests: [{ url: '', blocked: false }],
+  requests: [{ url: '', blocked: false, modified: false }],
+  requestsBlocked: ({ requests }) => requests.filter((r) => r.blocked),
```
smalluban, line 18: Move computed properties to the end of the map. It splits visually what is a part of the source data, and what is a computed local data. It is done in that way for `Stats`.

extension-manifest-v3/src/store/tab-stats.js
```diff
@@ -26,6 +28,10 @@ const Tracker = {
 const Stats = {
   domain: '',
   trackers: [Tracker],
+  trackersBlocked: ({ trackers }) =>
```
smalluban, line 31: Move `trackers...` to the end of the computer properties list (after `categories`).

extension-manifest-v3/src/background/env.js
philipp-classen: Oh, actually good point. Monkey patching makes also feature detection of APIs difficult (or impossible?). For page messages, I'm assuming that it is possible to detect missing APIs. Monkey patching can break assumptions in other places.

chrmod: it is not easy to fix it the right way. having this merged and tested on a platform that does not resolve to monkey-patching is IMAO way more important.

philipp-classen: I'm just realizing that it is in the wrapper, not the library itself. I would be fine with leaving it for now as is. Ideally, we should eventually change the library code to expect missing APIs. I think, then we can safely remove the monkey patching.

(I also checked: it will not conflict with current assumptions in page messages.)

extension-manifest-v3/src/pages/panel/views/tracker-details.js
chrmod: I've inlined it with some code repetition

smalluban (approved):

PR #1219, opened 2023-07-04
8.11.0 release

chrmod:

PR #1223, opened 2023-07-06
Enable blocking of service worker requests in Chrome

philipp-classen:
To be used in combination https://github.com/ghostery/common/pull/83

Background: requests from a website's service worker will not have a tab. That breaks breaks assumptions in the Ghostery adblocker. The result is that these requests will be skipped.


chrmod:
covered by https://github.com/ghostery/ghostery-extension/pull/1219

PR #1225, opened 2023-07-17
Registerable trackers preview and serp tracking content scripts

### LINKED ISSUE #869: Simplify manifest host permissions
To clarify that Ad-Blocking requires so called "wide host permissions" we have to load all content scripts on all pages. 

If we continue to load Trackers Preview just on google.*, Safari will list all those domains in the extension preference pages, which makes it look like we need permissions to google only. This is misleading and incorrect. We want users to understand that Ghostery needs access to all pages to block ads.

### LINKED ISSUE #1213: ghostery.com cookie handling in Firefox
Handle:
* storage partitioning
* first party isolation

smalluban:
Fixes #869.

I checked the docs about registering content scripts and tested it out in Chrome, Firefox, and Safari. It looks (and as you think about it, it should) that registered content scripts are cleared when a user updates (re-install). It works the same as for the content scripts set in the `manifest.json` file - they are reloaded when the extension is installed/updated.

It means that we are safe to just register the content script if it should be there, but if we have a change/bugfix, we can just simply update the code and publish a new version - already registered content scripts will reload.

Before:
Zrzut ekranu 2023-07-17 o 12 00 38

After:
Zrzut ekranu 2023-07-17 o 12 06 29


extension-manifest-v3/src/store/options.js
```diff
@@ -167,19 +167,19 @@ async function migrateFromMV2() {
 }
 
 export async function observe(property, fn) {
```
smalluban, line 169: This is a copy-paste from the https://github.com/ghostery/ghostery-extension/pull/1161/files#diff-ab787eda050f8d097068bea78c9043e942c4480fdb99acf1f1c0213b0bbd859dL169

extension-manifest-v3/src/background/serp-tracking.js
```diff
@@ -0,0 +1,26 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { observe } from '/store/options.js';
+
+import { domainList } from './utils/google.js';
+import { toggleDynamicContentScript } from './utils/scripts.js';
+
+observe('terms', async (terms) => {
```
smalluban, line 17: I know we had an option in the design for the Settings page, which would allow a user to switch on/off this feature. However, the current design does not have it. Do we want to add an option, or do we want to check only the `terms`?

smalluban:
There are two issues that I have discovered:
1. After deeper testing in Safari I found a bug, where using this API unregisters all of the content scripts when you try to register new ones. It clears them out for the current browsing context. After reloading the browser, it goes back to the correct state (with properly registered dynamic content scripts)
2. Apple added support for that API in Safari 16.4, which is quite fresh, so we still have a number of users with older versions. I am not sure if this is a good time to bump the minimal version in the manifest.

PR #1228, opened 2023-07-18
MV2: externalise trackers preview

chrmod:


chrmod:
based on this I'm not sure if we should go this way https://www.npmjs.com/package/webpack-target-webextension

chrmod:
closing in favour of #1270

PR #1234, opened 2023-07-20
Sync options

smalluban:
Replaces https://github.com/ghostery/ghostery-extension/pull/1161

Fixes https://github.com/ghostery/ghostery-extension/issues/1012, Fixes https://github.com/ghostery/ghostery-extension/issues/1166

I hope this is the final approach. In short, the following logic is implemented:

1. We listen when a user visits `account.ghostery.com` to detect if the user actively log in / log out
2. Each time options are fetched for the first time in the current context, we try to sync (check session, get options from server, and eventually send to server), like opening panel, opening settings page
3. In the background process we set the one-day delayed alarm for periodic checks if the server state has changed
4. The `revision` is always updated from the last state of the server

Additionally:
* Marking dirty state (if sending to the server fails) is done by adding a negative sign to the local revision
* If the session is not found, the sync process clears out the revision
* If a user updates options with an out-of-sync state, the server options are merged with local ones, so the currently updated options are not overwritten
* `options.engines` are replaced with flat options, like `blockAds`, etc...


chrmod (dismissed):
Looks good - still please wait for a review from @philipp-classen

extension-manifest-v3/src/utils/api.js
```diff
@@ -77,3 +83,45 @@ export async function session() {
...
+    body: JSON.stringify({
+      data: {
+        type: 'options',
+        id: userId,
+        attributes: { options },
+      },
+    }),
+  });
+
+  return (await data.json()).data.attributes.options || {};
```
philipp-classen, line 126: Looks like this can fail if the session is expired:

```
Error while syncing options:  TypeError: (intermediate value).data is undefined
```

If the session is expired, the fetch above will see a 401 with a JSON body like that:

```
{
  "errors": [
    {
      "title": "token is not valid",
      "status": "401",
      "code": "10020",
      "links": {
        "about": "https://github.com/ghostery/auth/wiki/Errors#10020--token is not valid"
      }
    }
  ]
}
```

Same in getUserOptions, I suppose. If there is a refresh token, the extension could try to renew the Bearer token and if successful, retry the get/set operation that failed.

smalluban: Hm.. but we do a refresh of the token, each time we call for `session()`, and it is used in sync process:

https://github.com/ghostery/ghostery-extension/pull/1234/files#diff-ab787eda050f8d097068bea78c9043e942c4480fdb99acf1f1c0213b0bbd859dR147

smalluban: We can add an additional check, or throw an error if the status is not 2xx, but the effect is the same, as the sync process is prepared to catch an error.

The sync requires a check of the user session, so I though that is redundant to make it again in those functions, as they are only used there.

philipp-classen (dismissed):

philipp-classen (approved):

PR #1237, opened 2023-07-21
Clean up DNR rules generator for Safari

### LINKED ISSUE #1227: DNR lists hit the limit
I can't find more reliable information about it than this site: https://webkit.org/blog/11727/release-notes-for-safari-technology-preview-126-with-safari-15-features/, which claims about the limit of 150k rules. However, the error in the background process shows that we cannot turn on all three lists at once.

smalluban:
It does not fix #1227, but, while testing DNR in Safari, I found that our helper function for list generation can be cleaned up, and it can better support some cases.

### DIFF
--- a/extension-manifest-v3/scripts/download-engines/utils.js
+++ b/extension-manifest-v3/scripts/download-engines/utils.js
@@ -43,32 +43,23 @@ export function getCompatRule(rule) {
 
   return {
     priority: rule.priority,
-    ...(rule.action.type === 'allowAllRequests'
-      ? {
-          action: { type: 'allow' },
-          condition: {
-            domains: rule.condition.requestDomains?.map((d) => `*${d}`),
-            resourceTypes,
-            regexFilter: '.*',
-          },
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/scripts/download-engines/utils.js
```diff
@@ -82,6 +73,8 @@ export function setupStream(path) {
 
   return {
     write: (rule) => {
+      if (!rule) return;
```
chrmod, line 76: when does this occur? shouldn't we at least log when this happen?

chrmod (approved):

extension-manifest-v3/scripts/download-engines/utils.js
smalluban: It moves the condition from the generator to the stream. If you look at the change in the PR, now we call `write` always, but the Safari util sometimes returns null, for example when the rule is not supported.

PR #1238, opened 2023-07-21
Add support for first-party cookie isolation

### LINKED ISSUE #1213: ghostery.com cookie handling in Firefox

smalluban:
Fixes #1213

### DIFF
--- a/extension-manifest-v3/src/utils/api.js
+++ b/extension-manifest-v3/src/utils/api.js
@@ -25,8 +25,26 @@ export const SIGNON_PAGE_URL = `https://signon.${DOMAIN}/`;
 export const CREATE_ACCOUNT_PAGE_URL = `https://signon.${DOMAIN}/register`;
 export const ACCOUNT_PAGE_URL = `https://account.${DOMAIN}/`;
 
+async function isFirstPartyIsolation() {
+  if (isFirstPartyIsolation.value === undefined) {
+    try {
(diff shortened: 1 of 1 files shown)


extension-manifest-v3/src/utils/api.js
```diff
@@ -38,6 +56,7 @@ export async function setCookie(name, value, durationInSec = COOKIE_DURATION) {
     name,
     value,
     expirationDate: Date.now() / 1000 + durationInSec,
+    firstPartyDomain: (await isFirstPartyIsolation()) ? DOMAIN : '',
```
chrmod, line 59: are you sure that this param can be there on all platforms and with first party isolation disabled? I remember the API throwing exceptions when first party isolation was not supported, this is way MV2 adds this param conditionally firstPartyDomain

chrmod (changes requested):
please confirm it was tested on all supported platforms

extension-manifest-v3/src/utils/api.js
smalluban: I can check again to be 100% sure, but from the MDN it looks like it should be fine to set `''` - this is the state when it is not supported.

smalluban: In Safari is fine, but Chrome does not like the parameter, so you were right ;)

chrmod (approved):

PR #1253, opened 2023-08-10
Release 10.1.2

smalluban:
TestFlight: V10.1.2 (115) 

89028d4b945dd0111565a7342986f218dc3e6f33 Add options migration (#1252)

## Testing

Update case:
1. Install build 114 and onboard it. Disable one/two main options.
2. Update to build 115
3. All main options should be switched to on.
4. Change one of the main options and restart the browser
5. Switched off option should still be switched off

Install case:
1. Clear extension data if needed (Dev Tools)
2. Install build 115
4. Onboard the user - main options should be switched on


GRadziejewski (approved):
Tests are finished.

PR #1254, opened 2023-08-11
Safari: fix DNR rules

chrmod:


philipp-classen (dismissed):

extension-manifest-v3/scripts/download-engines/utils.js
```diff
@@ -55,12 +64,31 @@ export function getCompatRule(rule) {
...
-      urlFilter:
-        rule.condition.urlFilter ||
-        (rule.condition.regexFilter ? undefined : '*'),
+      urlFilter: rule.condition.urlFilter,
       regexFilter: rule.condition.regexFilter,
+      isUrlFilterCaseSensitive: rule.condition.isUrlFilterCaseSensitive,
     },
   };
+
+  if (!newRule.condition.urlFilter && !newRule.condition.regexFilter) {
```
smalluban, line 73: The logic didn't change from the previous revision, only you added a log message. Is this gives us anything?

smalluban (changes requested):
I am sure that logging was quite helpful when debugging, but I am not sure if we should leave it there, as the second argument of the `getCompatRule()` is never used. At least, I would add a way to call the command like this `npm run download-engines --debug` or something similar.

smalluban (approved):
I have added `--debug` mode, so it can stay as it is for now.

PR #1259, opened 2023-08-21
Fix wrong Safari cookies API implementation

### LINKED ISSUE #1257: Buggy implementation of `cookies.set()` in Safari
When a user `access_token` and `csfr_token` is refreshed by the extension, the `cookies.set()` method is used. Safari has a buggy implementation of the API:

* When setting, `expirationDate` is counted from 2001-01-01 00:00, instead of the beginning of the epoch
* When getting, `expirationDate` returns a timestamp in milliseconds, instead of seconds

Reference: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/set

Because of the bug, in Safari, cookies are currently set with a date in the far future (2054..). The `access_token` expires in a day, so when a user tries to access `account.ghostery.com`, the application assumes that `access_token` is valid, making a page an endless loop of redirects.

We should fix both, the extension and account page.

smalluban:
Fixes #1257

### DIFF
--- a/extension-manifest-v3/src/utils/api.js
+++ b/extension-manifest-v3/src/utils/api.js
@@ -19,11 +19,46 @@ const ACCOUNT_URL = `https://accountapi.${DOMAIN}/api/v2.1.0`;
 export const COOKIE_DOMAIN = `.${DOMAIN}`;
 const COOKIE_URL = `https://${DOMAIN}`;
 const COOKIE_DURATION = 60 * 60 * 24 * 90; // 90 days in seconds
-const COOKIE_SHORT_DURATION = 60 * 60 * 24; // 1 day in seconds
+const COOKIE_SHORT_DURATION = 60 * 60; // 1 hour in seconds
+let COOKIE_EXPIRATION_DATE_OFFSET = 0;
 
 export const SIGNON_PAGE_URL = `https://signon.${DOMAIN}/`;
 export const CREATE_ACCOUNT_PAGE_URL = `https://signon.${DOMAIN}/register`;
 export const ACCOUNT_PAGE_URL = `https://account.${DOMAIN}/`;
+
+if (__PLATFORM__ === 'safari') {
+  // Safari has two major inconsistency with the specification:
+  // * for cookies.set() the `expirationDate` is in seconds since 2001-01-01T00:00:00Z (instead of beginning of epoch)
+  // * cookies.get() returns the cookie with `expirationDate` in in milliseconds (instead of seconds)
+  //
+  // Below code tries to detect the offset between the two
+  // and use it to convert the expirationDate to seconds
+  try {
+    const cookie = await chrome.cookies.set({
+      url: COOKIE_URL,
+      domain: COOKIE_DOMAIN,
+      path: '/',
+      name: 'test',
+      value: '',
+      expirationDate: 1,
+    });
+
+    // Other browsers don't return a cookie with expirationDate is in the past
+    if (cookie) {
+      const date = new Date(cookie.expirationDate);
+
+      // If the year is not 1970, it means there is a bug
+      if (date.getFullYear() !== 1970) {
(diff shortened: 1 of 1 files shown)


extension-manifest-v3/src/utils/api.js
```diff
@@ -66,6 +100,20 @@ export async function session() {
 
   let accessToken = await getCookie('access_token');
```
chrmod, line 101: shouldn't you extract the cookie value here? as it was in https://github.com/ghostery/ghostery-extension/pull/1259/files#diff-c917c022c9d3fb48f76186de6bbbed139cc1e135fcbfd191fa76dd7408483371L48

extension-manifest-v3/src/utils/api.js
```diff
@@ -100,19 +148,19 @@ export async function session() {
...
+  return jwtDecode(accessToken.value);
 }
 
 export async function getUserOptions() {
   const userId = await getCookie('user_id');
   const accessToken = await getCookie('access_token');
   const csrfToken = await getCookie('csrf_token');
 
-  const res = await fetch(`${ACCOUNT_URL}/options/${userId}`, {
+  const res = await fetch(`${ACCOUNT_URL}/options/${userId.value}`, {
```
chrmod, line 159: in case the userId cookie is falsy this will result in not very readable error. Shouldn't you check if the cookie exist before you use it value?

extension-manifest-v3/src/utils/api.js
```diff
@@ -19,12 +19,47 @@ const ACCOUNT_URL = `https://accountapi.${DOMAIN}/api/v2.1.0`;
...
+      if (date.getFullYear() !== 1970) {
+        COOKIE_EXPIRATION_DATE_OFFSET = -Math.round(
+          // Clear "1" from setting the cookie and transform to seconds
+          (cookie.expirationDate - 1000) / 1000,
+        );
+      }
+    }
+  } catch (e) {
+    // Protect against throwing an error when trying to detect the offset
+  }
```
chrmod, line 60: how about removing the cookie in the finally block?

extension-manifest-v3/src/utils/api.js
```diff
@@ -19,12 +19,47 @@ const ACCOUNT_URL = `https://accountapi.${DOMAIN}/api/v2.1.0`;
...
 
+if (__PLATFORM__ === 'safari') {
+  // Safari has two major inconsistency with the specification:
+  // * for cookies.set() the `expirationDate` is in seconds since 2001-01-01T00:00:00Z (instead of beginning of epoch)
+  // * cookies.get() returns the cookie with `expirationDate` in in milliseconds (instead of seconds)
+  //
+  // Below code tries to detect the offset between the two
+  // and use it to convert the expirationDate to seconds
+  try {
+    const cookie = await chrome.cookies.set({
```
chrmod, line 37: wonder what's the cost of this test and if there is a way we cache the result between iOS extension restarts. 
Alternatively we could check Safari version which should be cheaper.

extension-manifest-v3/src/utils/api.js
smalluban: Yes, and we do. To avoid multiple calls to `session()` it is done out side of this function. This one assumes that user is logged in correctly.

extension-manifest-v3/src/utils/api.js
smalluban: The cookie with past `expirationDate` is never saved, even in buggy Safari.

extension-manifest-v3/src/utils/api.js
smalluban: Ad 1. I don't think calling one API (cookie.set) is faster than another API, as the `chrome.storage.local` must eventually get data from the disk space.

Ad 2. If we set the version, which would it be? What do you think about the future versions? We have no idea when, or if any, they will fix the issue.

chrmod (approved):

PR #1262, opened 2023-08-25
Disable main features before terms are accepted

smalluban:
It's rather an edge case, but if a user clears out the storage by the dev tools, and still has open the settings page, it is possible to switch on the features without accepting the terms.

The PR adds a hard condition to main features, to be enabled only if `options.terms` is set. Because of that, we can simplify the default values for main options to be `true` (then onboarding does not have to switch them).

### DIFF
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -45,9 +45,9 @@ export const ENGINES = [
 
 const Options = {
   // Main features
-  blockAds: false,
-  blockTrackers: false,
-  blockAnnoyances: false,
+  blockAds: true,
+  blockTrackers: true,
+  blockAnnoyances: true,
 
   // Never-consent popup
   autoconsent: {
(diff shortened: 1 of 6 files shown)


extension-manifest-v3/src/store/options.js
```diff
@@ -45,9 +45,9 @@ export const ENGINES = [
 
 const Options = {
   // Main features
-  blockAds: false,
```
chrmod, line 48: couldn't this be achieved fully on model level? eg. `blockAds` getter could take `terms` into consideration

chrmod (commented):
wonder if this wont backfire in cases when we forget to add `terms` condition around any of the main features.

extension-manifest-v3/src/store/options.js
smalluban: It could, but the main features are already implemented. Even if there will be something, that needs that condition, I think it is worth it.

philipp-classen (approved):

PR #1270, opened 2023-09-04
Move trackers-preview to the main repository

smalluban:
Moves `trackers-preview` and `prevent-serp-tracking` from `webextension-packages` into this repository.

As the code is shared between both extensions (mv2, mv3), we have to add another workspace. As the number of them grows, I thought that shared packages should be placed in one subpath: `/packages/*`. We can keep the main apps in the root path, as they are using those "packages" (`ui`, `libs`, and now also `trackers-preview`).

The `prevent-serp-tracking` is only used in MV3, so I've just placed the code directly in the source code.

In addition, I made some cleanup/refactoring:
* Eslint & family packages, which are used by all of the packages/apps, are now dev-deps in the root, so it will be easier to update them
* All packages have bumped version to `1.0.0` (it should be not required to bump them anymore - they are linked)
* `webextension-packages` is updated to `v4.0.0` (where moved features are removed)


chrmod (approved):

smalluban:
@chrmod I have hacked GitHub. It requires passing a non-existing check - I moved `test-libs`... so I can't merge the PR. Can you do it? Do you have rights?

philipp-classen (approved):

PR #1272, opened 2023-09-05
Ghostery 10 packaging

chrmod:
TODO:
* [x] continuous deployment (on github releases) 
* [x] decision on how to work with `npm run release` (cc @smalluban)

### DIFF
--- a/.github/workflows/artifact-upload.yml
+++ b/.github/workflows/artifact-upload.yml
@@ -1,25 +1,23 @@
-name: Upload builds
+name: Artifact Upload
 
 on:
-  push
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
 
 jobs:
-  upload-builds-mv3:
+  artifact-upload:
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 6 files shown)


smalluban:
In my understanding, the flow should look like this:
1. When we want to release a new version, run `npm run release` by hand - it creates a new branch, bumps the version, checks localizations, etc.
2. Push the branch and create PR for the release
3. Use `npm run package` on that branch to prepare packages for testing (this could be automatic - for example, a GitHub action, which runs only for a specific branch name - `mv3-release` - and this is hardcoded in the `release` script)

As we publish (to iOS/macOS) version from the branch, I am skeptical about making packages after merging to the main (or for the release tag) - as we have no guarantee that anything else was merged there in the meantime (This is a downside of the current process). 

If we would agree that we can skip some version numbers, we can bump the version (and make packages) every time something is merged to the main (and touches anything except mv2 code). It would open the possibility of making it more automatic. However, there will be still unsolved problem with translations - then they should be updated by hand in the separate PR (which might be ok).

chrmod:
merging for sake of testing the CI

PR #1275, opened 2023-09-07
Fix trackers preview for endless scroll on SERP

smalluban:
Add support for the endless scroll in SERP. Technically, it adds the `data-wtm` attribute to already generated result items, so it only fetches new results.


packages/trackers-preview/src/content_scripts/index.js
```diff
@@ -118,15 +118,6 @@ export default function setupTrackersPreview(popupUrl) {
           return;
         }
 
-        document.addEventListener('click', (event) => {
```
smalluban, line 121: This event listener was only to hide the popup when clicking on the result item link. I think it's not important (I am not sure why we added it here), and it can't be in that form with reloading the setup for the endless scroll case (it would add multiple events on the document).

philipp-classen: Unfortunately, I cannot try it out, since I'm currently unable to get any Tracker Preview popup at all. (Not tested with this PR, but in general. Don't know why.)

Maybe the feature is to have a convenient way to close the popup by clicking on anything outside the popup itself to close it (not only on the X). I could see that this adds value.

smalluban: It works only for links, so when you navigate out from the page.

packages/trackers-preview/src/content_scripts/index.js
```diff
@@ -181,5 +174,18 @@ export default function setupTrackersPreview(popupUrl) {
         resizePopup(height);
       }
     });
+
+    const observer = new MutationObserver((mutations) => {
+      if (mutations.some((m) => m.addedNodes.length)) {
+        observer.disconnect();
+        console.log('delaying setupTrackersPreview');
```
chrmod, line 181: remove

chrmod (changes requested):

philipp-classen (approved):

smalluban:
@chrmod I need your approval as you requested changes.

philipp-classen:
> @chrmod I need your approval as you have requested changes.

I just merged. (The console log is gone, so that should be resolved)

PR #1276, opened 2023-09-07
Release 10.1.4

chrmod:
d3889952412b47ad2861dbd14f388633f78bbd48 Fix trackers preview for endless scroll on SERP (#1275)
23f745042720db0abe1ab591d240f52bf5e2ba50 Tracker Preview: detect more Google layouts (#1282)
35d467d09ef934f6e22604cd2d8356d6b9783666 Dynamic content scripts for trackers preview and serp tracking (#1278)
f66c6ee2397a68eb9bdd3013cc20ae40e9e58835 Prevent from syncing not existing property (#1280)
3f7b401951582674aaf943328820f906884cf09b Follow the `true` default value for main options in the MV2 migration (#1281)
0926a2aeb76ce2873c52544470c145803b0de117 CI: upload named artefacts only on labeled PRs (#1277)
068efccb26652cbe83676cfbe7b6ee4de348ee96 Ghostery 10 packaging (#1272)
159b7d58efa064217457708b49246428a0f8ecc8 Move trackers-preview to the main repository (#1270)
ea2beeec83f4be619951256c798baaeb74eaf32b Fix options assignment for MV2 migration (#1271)
b832d188f7540617d4617ae6956bae8303964b41 Disable main features before terms are accepted (#1262)
0a9f35311aae1107e8ef845fa2d9f49991c0bc7b Catch for errors when trying to refresh user's session
dd8d02b47084f6320058a9c8363557be03748e38 Rename general `alarms.js` to `paused.js`
b257e6ccfc5dcb1f983135b0de1a4a95022873f0 Update Autoconsent
41833316e0577176d106c02fe2bdbdc6341d86aa Update WTM packages

## Testing highlights

* Fixes and change to dynamic loading on `trackers-preview`
* Switchable SERP tracking - new option in settings
* Minor fixes on session fetching (better catching errors) and options sync

PR #1278, opened 2023-09-08
Dynamic content scripts for trackers preview and serp tracking

### LINKED ISSUE #869: Simplify manifest host permissions

smalluban:
Fixes #869.

Instead of registering content scripts (as Safari has a bug, which clears out content scripts from manifest.json), this code change introduces script and CSS injection dynamically. 

Highlights:
* Adds `serpTrackingPrevention ` option, and copy from existing design - it might have to be reviewed
* No changes in `packages/trackers-preview`, so in MV2 everything should work like before 🎉


extension-manifest-v3/src/pages/settings/views/privacy.js
```diff
@@ -129,6 +134,35 @@ export default {
...
+            <div
+              layout="column gap:2"
+              layout@768px="row items:center gap:5 grow"
+            >
+              <div layout="column grow gap:0.5">
+                <ui-text type="headline-s">
+                  Tracking Prevention on Search Engine Result Pages
+                </ui-text>
+                <ui-text type="body-l" mobile-type="body-m" color="gray-600">
+                  While using a non-private search engine like Google or Bing,
```
smalluban, line 155: We do it only for Google SERP, so the `Bing` here might should be removed.

extension-manifest-v3/src/store/options.js
```diff
@@ -63,6 +63,8 @@ const Options = {
 
   // Tracker wheel on SERP
   wtmSerpReport: true,
+  // Prevent SERP tracking
+  serpTracking: true,
```
chrmod, line 67: lets call it `serpTrackingPrevention`

extension-manifest-v3/src/background/serp.js
```diff
@@ -0,0 +1,81 @@
...
+import { store } from 'hybrids';
+import {
+  tryWTMReportOnMessageHandler,
+  isDisableWTMReportMessage,
+} from '@ghostery/trackers-preview/background';
+import css from '@ghostery/trackers-preview/content_scripts/styles.css?raw';
+
+import Options from '/store/options.js';
+
+const SERP_URL_REGEXP = /google\.[a-z]+(\.[a-z]+)?\/search/;
```
philipp-classen, line 21: At other places, we are using patterns like that to avoid false-positives:

`^https:[/][/][^/]*[.]google[.]`

So, this more conservative regexp should still match:
```
/^https:[/][/][^/]*[.]google[.][a-z]+([.][a-z]+)?[/]search/
```

chrmod (changes requested):

chrmod (approved):

chrmod:
lets test before merging @GRadziejewski

philipp-classen (approved):

philipp-classen:
> lets test before merging @GRadziejewski

I would prefer to directly merge, since it helps to review https://github.com/ghostery/ghostery-extension/pull/1275

For me, without this fix here, the injection runs too early, finds nothing and as a consequence the other PR also has no effect. Only with these fixes, I would see the preview on MV3 in Chrome.

smalluban:
Let's move testing to [`v10.1.14` ](https://github.com/ghostery/ghostery-extension/pull/1276) PR.

PR #1285, opened 2023-09-12
Add dynamic update to adblocker engines

smalluban:
Based on `ghostery-common` adblocker engines manager: 
https://github.com/ghostery/common/blob/main/modules/adblocker/sources/manager.es

Where possible, the logic (and comments) of the update processs are exactly the same. The only difference in the algorithm is when one of the lists must be removed. The v8 loads from CDN as a last resort, so resetting to this state is fine (it will be still the last full engine). In v10 we pack a file into the extension, so resetting to this state would move it to the state, where it can be incompatible with the update process. In other words, in that case, we need to fetch a full file with the engine (and we do so).

Resetting (fetching a new engine) replaces object instances (not updates). Because of that, I had to introduce an API, where we get the current instance each time when we need it (as it might be switched) - clients must use `get()` method, not keep the instance somewhere.

The public API of `engines.js`:
* `init(name)` - schedule alarm and fetches engine from all resources (memory -> storage -> disk)
* `get(name)` - returns engine from the memory

The update process is triggered in two places:
1. Scheduled alarm each 24 hours
2. Initial load from the disk - it should be triggered only once, as the engine is saved in indexeddb, so even after restarting the extension, it should not load again from the disk


smalluban:
After a discussion with @chrmod I'll add a manual update button in our Dev Tools for easier debugging.

extension-manifest-v3/src/background/devtools.js
```diff
@@ -10,38 +10,58 @@
...
+      return true;
+
+    case 'updateEngines':
+      (async () => {
+        try {
+          await engines.updateAll();
+        } catch (e) {
+          sendResponse(`Error updating engines: ${e}`);
+        }
+        sendResponse(`Engines updated`);
```
philipp-classen, line 61: If an exception is thrown, the "Engines updated" message will also be send. Would it be better to move after "await engines.updateAll()"?

smalluban: Yes, you are right

extension-manifest-v3/src/background/utils/engines.js
```diff
@@ -0,0 +1,336 @@
...
+}
+
+async function update(name) {
+  try {
+    const urlName =
+      name === 'trackerdb'
+        ? 'trackerdbMv3'
+        : `dnr${__PLATFORM__ === 'firefox' ? '' : '-cosmetics'}-${name}`;
+
+    const data = await fetch(
```
philipp-classen, line 104: Since this polls for updates, caching can create problems since you can end up with an old version. If the update call here is guaranteed to be called not too often, you can disabling it with

```
fetch(url, {
  method: 'GET',
  cache: 'no-cache',
  credentials: 'omit',
}
```
  
(e.g. https://github.com/whotracksme/webextension-packages/blob/257b4f6dba519497119ecf9d16c1475ef435ff49/packages/reporting/src/patterns-updater.js#L193)

But if it is called frequently, disabling caching can be wasteful for the client. One compromise would be to introduce a "force" flag and only bypass the browser cache in force mode.

smalluban: This is a service, which is in our control. Why should we bypass our own cache settings? The update process is triggered once a day, so if the endpoint has an hour of cache, it should be fine. The only case is when you want to trigger the update process manually, but if we need to have the ability to fetch it more often, let's change the cache setting on the server side - clients normally don't call it more often than once a day (so in that case, the cache could be turned off completely).

philipp-classen: Regarding the server settings: serving a mutable file as "immutable" (even with only one hour) is indeed a bit contradicting. A release of the adblocker invalidates the CDN, which makes the new content accessible from the internet.

For the client: The interesting use case is indeed to force an update. But I think, we can stick with the current workflow, which is to install a fresh extension and test it with that. That is the only reliable way to force the latest block list (also for the current MV2 Ghostery).

Though we should keep in mind that forcing an update through the dev UI is unreliable.

philipp-classen (approved):

extension-manifest-v3/src/background/utils/trackerdb.js
```diff
@@ -54,9 +45,8 @@ export async function getMetadata(request) {
 
 const patterns = new Map();
 export async function getPattern(key) {
-  if (trackerDBStartupPromise) {
-    await trackerDBStartupPromise;
```
chrmod, line 58: this is a breaking change - why don't we wait for trackerdb to be ready?

extension-manifest-v3/src/background/utils/engines.js
```diff
@@ -0,0 +1,336 @@
...
+  ENGINE_VERSION,
+  getLinesWithFilters,
+  mergeDiffs,
+} from '@cliqz/adblocker';
+
+import { registerDatabase } from '/utils/indexeddb.js';
+
+const engines = new Map();
+
+function loadFromMemory(name) {
```
chrmod, line 24: are those functions (loadFromMemory/saveToMemory) needed? why not access the Map directly?

extension-manifest-v3/src/background/utils/engines.js
```diff
@@ -0,0 +1,336 @@
...
+  return engines.get(name);
+}
+
+function saveToMemory(name, engine) {
+  engines.set(name, engine);
+}
+
+const DB_NAME = registerDatabase('engines');
+
+async function getDB() {
```
chrmod, line 34: why not share with daily-stats?

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -20,37 +19,28 @@ import { observe, ENGINES } from '/store/options.js';
...
 let pausedDomains = [];
 
 const setup = asyncSetup([
   observe(null, (options) => {
-    ENGINES.forEach(({ name, option }) => {
-      const engine = adblockerEngines.find((e) => e.name === name);
-      engine.isEnabled = options.terms && options[option];
-    });
+    // Set enabled engines
+    enabledEngines = ENGINES.reduce((acc, { name, key }) => {
```
chrmod, line 32:
```js
enabledEngines = ENGINES
  .filter(({ key }) => options.terms && options[key])
  .map({ name }) => name);

chrmod:
looks like it is ready to merge, my comments are not blocking

extension-manifest-v3/src/background/utils/engines.js
smalluban: This is done only for the algorithm clarity - instead of `engines.set()` you have `saveToMemory()`. It keeps things clear in functions, what when happen (which kind of storage is being used).

extension-manifest-v3/src/background/utils/engines.js
smalluban: The common code is not that huge, as each db has its own upgrade and blocking implementation. Only the shell of the function is similar. I think it is not worth it.

extension-manifest-v3/src/background/utils/trackerdb.js
smalluban: I added await for the init promise.

chrmod (approved):

philipp-classen (approved):

PR #1300, opened 2023-09-21
Add opera notification for enabling extension on SERP

smalluban:
Relates to https://github.com/ghostery/journal/issues/180.

After discussion, the following logic was implemented:

* Danger extension icon and panel notification are shown until a user enables extension on SERP or ignores notification 4 weeks in a row, so we disable the warning entirely. 
* The notification is shown immediately (only for onboarded and never-consent active users) and continuously until the user interacts with the popup. Then we show it again after a week of pause.  If the user ignores it for the 4th time, we disable the warning.

Additionally, as we show notifications on every page (until interaction with the popup) - closing one popup clears out all of the popups on opened tabs (only the same popups - the URL is the id).

I am not sure about the naming of files/methods, so if you have a better proposition, you're welcome to change it.


extension-manifest-v3/src/background/onboarding.js
```diff
@@ -12,32 +12,45 @@
 import { observe } from '/store/options.js';
 import { sendShowIframeMessage } from './utils/iframe.js';
 import { tabStats } from './stats.js';
+import { showOperaSerpNotification } from '/utils/opera-serp.js';
```
chrmod, line 15: I think we can put in `notifications` folder (can be nested under some other folder but I would not call it `utils`, maybe `features`) - it is very likely we will have more of those. One that comes to mind would be a notification to enable Ghostery in incognito windows.

extension-manifest-v3/src/background/onboarding.js
```diff
@@ -12,32 +12,45 @@
...
+              details.tabId,
+              'pages/onboarding/iframe.html',
+            );
+          }
+        }, 1000);
       }
-    }, 1000);
+    } else if (__PLATFORM__ === 'opera') {
+      showOperaSerpNotification(() => {
+        sendShowIframeMessage(details.tabId, 'pages/onboarding/serp.html');
```
chrmod, line 52: shouldn't this logic be encapsulated inside `showOperaSerpNotification` - magic sting suggests it is wrong place

extension-manifest-v3/src/utils/opera-serp.js
```diff
@@ -0,0 +1,77 @@
...
+    // Run the callback function
+    if (cb) cb();
+
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+export async function showOperaSerpAlert() {
```
chrmod, line 72: this name suggests an action. how about to call it starting with `should`

extension-manifest-v3/src/pages/onboarding/iframe.js
```diff
@@ -12,22 +12,19 @@
...
 
-setupIframeSize();
+setupIframe();
 
 async function close(host, event) {
   await store.set(Options, {
     onboarding: event.type === 'ignore' ? { shownAt: Date.now() } : null,
   });
 
-  // Options notify other contexts about changes, but
```
chrmod, line 27: why this was removed? did Safari fix something? if so, what about older versions?

extension-manifest-v3/src/pages/onboarding/serp.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width" />
+    <title>More ad blocking available</title>
```
chrmod, line 6: is this copy final? shouldn't it be localised? (not that majority of Opera users are in easter europe)

extension-manifest-v3/src/pages/onboarding/serp.js
```diff
@@ -0,0 +1,53 @@
+/**
```
chrmod, line 1: how about naming this file `opera-serp-notification` ? 
btw. not sure if this should be considered an onboarding, to me it is rather a notification.

extension-manifest-v3/src/pages/onboarding/serp.js
```diff
@@ -0,0 +1,53 @@
...
+import '@ghostery/ui/onboarding';
+import { setupIframe, closeIframe } from '@ghostery/ui/iframe';
+
+import { openTabWithUrl } from '/utils/tabs.js';
+import Options from '/store/options.js';
+
+setupIframe(360);
+
+async function updateOptions() {
+  return store.resolve(Options).then((options) =>
```
chrmod, line 23: why not `await`, the function is already `async`

extension-manifest-v3/src/utils/opera-serp.js
```diff
@@ -0,0 +1,77 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { store } from 'hybrids';
+
+import Options from '/store/options.js';
+
+const NOTIFICATION_SHOW_LIMIT = 4;
+const NOTIFICATION_DELAY = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds
+const TEST_COOKIE_NAME = 'ghostery:opera:cookie:test';
```
chrmod, line 18: just to be on a safe side - how about we make this name random to prevent detection?

extension-manifest-v3/src/background/onboarding.js
smalluban: Do you mean `__PLATFORM__...` condition? It must be outside, so the bundler can remove this code (and all of the imports) from platforms, which don't need it.

extension-manifest-v3/src/pages/onboarding/iframe.js
smalluban: No, it was moved to the `closeIframe` general function, so the timeout is triggered always (to prevent situation, when it would be needed, but we forget about it).

https://github.com/ghostery/ghostery-extension/blob/feat-opera-serp-notification/packages/ui/src/utils/iframe.js#L113

extension-manifest-v3/src/pages/onboarding/serp.html
smalluban: Generally, the `<title>` from the head is not visible anywhere when the page is used in the iframe element. I added it just to unify with other HTML files (which also have titles).

extension-manifest-v3/src/background/onboarding.js
smalluban: ok, can be `notifications`.

extension-manifest-v3/src/pages/onboarding/serp.js
smalluban: It's in the `onboarding` path, as this is a part of onboarding notifications - if you look at new installation, this is a notification, which relates to checking terms, never-consent or renew, and it must relay on those features, so this is a notification in the **onboarding process**.

smalluban: About the path, I would go with `opera-serp` to match other file names, but the `notification` prefix is in my opinion not required, and makes a file name too long.

extension-manifest-v3/src/utils/opera-serp.js
smalluban: During the test, the cookie is immediately removed after successful creation, so I doubt that any other code can detect it (and it sets `httpOnly` - so no other JS can access it).

mdsandu:
@smalluban this should fix the UX issue --> pls replace the link to the blog/guide in context notification + panel with the below - rest remains the same in flow: https://www.ghostery.com/blog/block-search-engine-ads-on-opera#strong-remove-ads-on-search-results-pages-strong
thx

extension-manifest-v3/src/background/onboarding.js
chrmod: I've meant the page path `'pages/onboarding/serp.html'`

chrmod: the `__PLATFORM__` condition is fine to stay, I ask about it's contents

chrmod (approved):

packages/ui/src/modules/onboarding/serp.js
```diff
@@ -0,0 +1,53 @@
+/**
```
chrmod, line 1: wouldn't it make more sense to call this file `opera-serp`?

smalluban: Sure, I think we moved to `oper-serp` name after the initial code, so it is kind of leftover.

extension-manifest-v3/src/notifications/opera-serp.js
```diff
@@ -48,9 +50,9 @@ async function isSerpSupported() {
   }
 }
 
-export async function shouldShowOperaSerpNotification(cb) {
+export async function showOperaSerpNotification(tabId) {
```
chrmod, line 53: now the name could start with `maybe` :D

chrmod (approved):

PR #1305, opened 2023-09-22
Use Session store model for getting user's data

### LINKED ISSUE #1303: Missing or outdated token
**Description**
Error **_"401 - token is not valid"_** appears during the attempt to sync account settings. After logging into the same account, on the first smartphone everything is working as expected. But on the second smartphone error appears during any attempts of changing the settings. There is a chance that on the second smartphone token is missing but app is still thinking and showing that I'm logged into my account.

**Steps to Reproduce**
1. Log into the account on the **first** smartphone 
2. Open any website, for example https://cnn.com and open Ghostery panel
3. Click on the hamburger and open Ghostery Settings
4. Turn OFF for example Ad-Blocking
5. Log into the account on the **second** smartphone 
6. Open any website, for example https://cnn.com and open Ghostery panel
7. Click on the hamburger and open Ghostery Settings
8. Check the settings - Ad-Blocking should be turned off
9. Still on the **second** smartphone check few other websites
10. Turn OFF also Anti-Tracking and Never-Consent
11. Switch to the **first** smartphone and open Ghostery Settings
12. Check the status of Anti-Tracking and Never-Consent - both are turned ON

https://github.com/ghostery/user-agent-android/assets/40433708/8ee5d149-da33-44fb-a795-88d1d5384a02

**Expected results**
It should be working correctly as expected. If token is missing user should be logged automatically or token should be refreshed or user should be logged out. 

Versions

	Browser: Ghostery Private Browser 2023.9 (6d239e7 )
	OS: Android 11 
	GBE 10.1.4

	Browser: Ghostery Private Browser 2023.9 (#2015973547 )
	OS: Android 13 
	GBE 10.1.4

smalluban:
Fixes #1303.

I've found that `session()` call is made by multiple places, which may trigger it until another call ends. It's critical when `access_token` expires, so the first call after tries to refresh the token. However, currently another call can be made at "the same time" - it also tries to refresh the token. As the `refresh_token` is already used (it can be used only once) - the second call fails, so in the `catch ()` statement, we are clearing cookies (again, another bug, only half)... As a result, in that case, the session is broken, and cannot be refreshed (but access_token is there, so the code thinks, that the user is logged in correctly).

The solution is to make calls to `session()` only by the `Session` store model. All of the places should use the store model. The `store.resolve()` method, and generally store architecture protects from getting the model, which is currently fetched.


extension-manifest-v3/src/utils/api.js
```diff
@@ -11,7 +11,7 @@
 
 import jwtDecode from 'jwt-decode';
 
-const DOMAIN = 'ghostery.com';
+const DOMAIN = 'ghosterystage.com';
```
chrmod, line 14: please revert

chrmod (changes requested):
In general looks good. Wonder if we can prevent ourselves from using the session incorrectly, eg. can we make `session()` private? 

Please add some documentation in the session model.

smalluban:
There is no technical way to make this function private, other than putting it inside of the Session model file. However, I would avoid that, as for now API implementation is defined in its own place, which separates logic better.

I made a comment for the `session()` function, so if it would be used in the future, it should show up in the editor snippet.

chrmod (approved):

Issue #1306, opened 2023-09-22
Renew popup disappears in Incognito windows


GRadziejewski:
**Description**
On Incognito mode I had a Renew pop-up, but when I click the green button - the pop-up disappeared. Nothing more is happening.

**Steps to Reproduce**
1. Open Incognito window
2. Open any website, for example https://cnn.com/
3. Wait 1h - Renew pop-up should appear
4. Press on the "Renew Setup" button
5. Pop-up disappear and nothing else is happening

https://github.com/ghostery/ghostery-extension/assets/40433708/263d1136-d665-4c6c-916e-124ccafcc284

**Expected results**
When I press "Renew Setup" button, new tab with an Onboarding should be opened.

Versions

	Browser: Chrome 117.0.5938.92 
	OS: Windows 10 Pro 22H2 19045.3393
	GBE 8.11.0

philipp-classen:
In a test setup, where I reconfigured the timeouts to fire every minute, I could reproduce the problem. With the PR here, I could finish the onboarding (https://github.com/ghostery/ghostery-extension/pull/1307).

philipp-classen:
Independent of the bug fix, here is a workaround for all who are affected:

* Make sure to have a non-incognito window open
* Open the extension (see screenshots), click on "Enable Ghostery" (blue button) and then you get the onboarding screen
* In the onboarding, click on "Enable Ghostery" (green button) and you should be able to finish the onboarding

![step1](https://github.com/ghostery/ghostery-extension/assets/1616509/6b69c8bc-da1b-4825-9d8e-8b78c1b00082)
![step2](https://github.com/ghostery/ghostery-extension/assets/1616509/41c9943c-8a12-4597-ad15-058408e2522a)
![step3](https://github.com/ghostery/ghostery-extension/assets/1616509/05aded5e-7e3b-49d9-bfd0-6625ea1f5fef)
![step4](https://github.com/ghostery/ghostery-extension/assets/1616509/9f288540-5fa3-4c73-b955-37c499baaab6)

philipp-classen:
After further research, it looks like we are hitting a bug in Chrome, unfortunately. Opening extension links in an incognito window does not work (only non-extension links do open). Thus, opening the onboarding in a non-private window, looks like the best that we can do in Chrome.

PR #1307, opened 2023-09-22
Potential fix for the renew problem on Chrome in Incognito mode

philipp-classen:
In my tests, I could also reproduce that the error path is reached. With this fix, it should still open the onboarding tab, which allows the user to proceed.

refs https://github.com/ghostery/ghostery-extension/issues/1306

### DIFF
--- a/extension-manifest-v2/src/utils/utils.js
+++ b/extension-manifest-v2/src/utils/utils.js
@@ -18,11 +18,8 @@
  */
 import { debounce } from 'underscore';
 import { URL } from '@cliqz/url-parser';
-import { getBrowserInfo } from '@ghostery/libs';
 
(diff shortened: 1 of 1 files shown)


philipp-classen:
Updated the PR:
* The error path (when getActiveTab fails) seems to be only reachable when using breakpoints and having the debugger as the active window. However, using the simpler Firefox code as a fallback worked
* Going one step further, we can unify the code to only use the Firefox path

During debugging, we noticed that otherwise, Chrome runs into the issue that it refuses to open URLs within the active private window. With this change, it might open the onboarding in a new, non-private window.

philipp-classen:
New version came in https://github.com/ghostery/ghostery-extension/pull/1310

PR #1310, opened 2023-09-25
MV2: fix opening new tab

chrmod:
fix https://github.com/ghostery/ghostery-extension/issues/1306

### DIFF
--- a/extension-manifest-v2/src/utils/utils.js
+++ b/extension-manifest-v2/src/utils/utils.js
@@ -228,7 +228,7 @@ export function getActiveTab(callback, error) {
 }
 
 export async function openNewTab({ url, become_active: active = false } = {}) {
-	if (!url) { return; }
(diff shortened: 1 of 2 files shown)


extension-manifest-v2/src/utils/utils.js
```diff
@@ -229,60 +227,25 @@ export function getActiveTab(callback, error) {
...
-		} else {
-			chrome.tabs.create({
-				url: data.url,
-				active: data.become_active || false
-			});
-		}
-	}, (err) => {
-		log(`_openNewTab Error: ${err}`);
+export async function openNewTab({ url, become_active: active = false } = {}) {
+	if (!url) { return; }
```
philipp-classen, line 231: Is it being reach? Or would it be better to let it throw an error if no URL is given?

philipp-classen (approved):

PR #1325, opened 2023-10-05
Fix panel scrollbar and sizing

### LINKED ISSUE #1315: Extension Panel Quick Improvements
We want to fit more content in max height and fix scrollbar issues:
- reduce spacings: https://www.figma.com/file/GIWeYHGf3bUQzCd3LK9VmL/Ghostery-Panel?type=design&node-id=1810%3A2762&mode=design&t=uIcsQZhJwUQrNGot-1
- removed Ghostery Settings label
- improve scrollbar: https://www.figma.com/file/GIWeYHGf3bUQzCd3LK9VmL/Ghostery-Panel?type=design&node-id=1821%3A2631&mode=design&t=uIcsQZhJwUQrNGot-1
- improve trackers wheel

smalluban:
Fixes #1315.

After a few attempts at the scrolling container problem, I finally found a pure CSS solution. Might be accidentally :P At first, I thought about counting children's heights and checking if they could fit into the view. However, making this scrolling container for a shadow requires nesting it in the container element. It looks like if you create a proper nested structure of flex containers, the `overflow: scroll` does not make it 0px height (as it is if you just put an element directly).

About the scrollbar styling - @kalinme - I would leave it as if you style the scrollbar, it won't hide when it is not needed, which looks worse than the default one. As the above fix makes it visible only in the container, I think it looks more than fine on Windows 10 (checked Edge, Opera, and Firefox).

The PR also makes the notification alert smaller (as it won't fit in the header anymore), and removed the notification alert on top when the user is not onboarded - it makes the things too high, so scrolling appears (I think it is not needed there, as there are so many CTAs to start onboarding).

### Firefox macOS

Zrzut ekranu 2023-10-5 o 16 27 44
Zrzut ekranu 2023-10-5 o 16 27 40
Zrzut ekranu 2023-10-5 o 16 27 19
Zrzut ekranu 2023-10-5 o 16 27 12

### Firefox Windows 10

Zrzut ekranu 2023-10-5 o 16 30 42
Zrzut ekranu 2023-10-5 o 16 31 07
Zrzut ekranu 2023-10-5 o 16 31 16
Zrzut ekranu 2023-10-5 o 16 31 34


extension-manifest-v3/src/pages/panel/views/home.js
```diff
@@ -106,168 +75,137 @@ export default {
...
+                  Enable Ghostery
+                </a>
+              </gh-panel-button>
+            `}
+        <gh-panel-container>
+          ${stats.domain
+            ? html`
+                <ui-panel-stats
+                  domain="${stats.domain}"
+                  categories="${stats.topCategories}"
```
chrmod, line 130: white space?

extension-manifest-v3/src/pages/panel/components/container.js
```diff
@@ -0,0 +1,82 @@
...
+      el.scrollTop + el.clientHeight < el.scrollHeight,
+    );
+  } else {
+    shadowTop.classList.remove('show');
+    shadowBottom.classList.remove('show');
+  }
+}
+
+export default {
+  _: {
```
chrmod, line 23: that's just a convention to manage the attached side effects, right?

How about using a symbol instead ? ;) jk

chrmod (changes requested):

extension-manifest-v3/src/pages/panel/views/home.js
smalluban: Not sure what you mean, this is just prettier...

extension-manifest-v3/src/pages/panel/components/container.js
smalluban: This might be a missing feature of hybrids. There is no simple way to attach `connect/disconnect` callback directly to the `render/content` properties. They must be functions to work properly, so you must create another property with the `connect()` method. The value of this property is irrelevant in this case, so I just have used `_` for convenience. 

Symbols are not iterated using `Object.keys()`, and this is by design, as `[router.connect]` and other similar properties use symbols, to avoid generating properties on the host.

extension-manifest-v3/src/store/tab-stats.js
```diff
@@ -33,26 +33,24 @@ const Stats = {
...
-    }, {});
+    const counts = Object.entries(
+      categories.reduce((acc, category) => {
+        acc[category] = (acc[category] || 0) + 1;
+        return acc;
+      }, {}),
+    );
 
-    const sorted = Object.entries(counts).sort((a, b) => b[1] - a[1]);
+    if (counts.length < 6) return categories;
```
chrmod, line 43: categories have to be sorted by name in order for this to work correctly

chrmod (changes requested):

chrmod (dismissed):

kalinme (approved):

chrmod (approved):

PR #1333, opened 2023-10-12
Fix stats leaking & minor fixes

smalluban:
WIP.

Fixes:
* Leaking stats from the previous page when navigation occurs
* Paused icon on sites where trackers are not detected
* Minor fixes (look the commit log)

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -21,7 +21,7 @@ import Request from './utils/request.js';
 import asyncSetup from './utils/setup.js';
 import * as engines from './utils/engines.js';
 
-import { setupTabStats, updateTabStats } from './stats.js';
+import { updateTabStats } from './stats.js';
 
 let enabledEngines = [];
 let pausedDomains = [];
@@ -333,24 +333,15 @@ function isPaused(details, request) {
 if (__PLATFORM__ === 'firefox') {
   chrome.webRequest.onBeforeRequest.addListener(
     (details) => {
+      if (details.tabId < 0) return;
+
       const request = Request.fromRequestDetails(details);
-      const tabId = details.tabId;
 
-      // Update stats
-      if (tabId > -1 && (request.sourceDomain || request.sourceHostname)) {
-        if (request.isMainFrame() && !request.isHttp && !request.isHttps) {
-          return;
+      // INFO: request.source... is only available in Firefox
+      if (request.sourceDomain || request.sourceHostname) {
+        if (details.type !== 'main_frame') {
+          updateTabStats(details.tabId, [request]);
         }
-
-        Promise.resolve().then(
-          request.isMainFrame()
-            ? () =>
-                setupTabStats(
-                  tabId,
-                  request.sourceDomain || request.sourceHostname,
-                )
-            : () => updateTabStats(tabId, [request]),
-        );
 
         if (isPaused(details, request)) return;
 
@@ -358,7 +349,7 @@ if (__PLATFORM__ === 'firefox') {
           const engine = engines.get(name);
           if (!engine) continue;
 
-          if (request.isMainFrame()) {
+          if (details.type === 'main_frame') {
             const htmlFilters = engine.getHtmlFilters(request);
             if (htmlFilters.length !== 0) {
               filterRequestHTML(
--- a/extension-manifest-v3/src/background/reporting/index.js
+++ b/extension-manifest-v3/src/background/reporting/index.js
@@ -137,11 +137,13 @@ if (__PLATFORM__ === 'firefox') {
       if (event === 'observed') {
         return;
       }
-      const request = Request.fromRawDetails({
+
+      const request = Request.fromRequestDetails({
         url: state.url,
-        sourceUrl: state.tabUrl,
+        originUrl: state.tabUrl,
       });
       request.modified = true;
+
       updateTabStats(state.tabId, [request]);
     },
   });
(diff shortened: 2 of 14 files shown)


extension-manifest-v3/src/background/stats.js
```diff
@@ -95,91 +94,97 @@ const delayMap = new Map();
...
-      let tracker = stats.trackers.find((t) => t.id === pattern.id);
+    let sortingRequired = false;
 
-      if (!tracker) {
-        tracker = { ...pattern, requests: [] };
-        stats.trackers.push(tracker);
-      }
+    // Filter out requests that are not related to the current domain
+    // (e.g. requests on trailing edge when navigation to a new page is in progress)
+    requests = requests.filter(
```
smalluban, line 139: This protects from leaking requests from another domain, so the current one is different (using `initiator` property from request details).

extension-manifest-v3/src/background/stats.js
```diff
@@ -318,7 +326,16 @@ if (__PLATFORM__ !== 'safari' && __PLATFORM__ !== 'firefox') {
       if (details.error !== 'net::ERR_BLOCKED_BY_CLIENT') return;
 
       const stats = tabStats.get(details.tabId);
-      stats?.requestsBlocked.push(details.requestId);
+      if (!stats) return;
```
smalluban, line 329: `stats.requestsBlocked` was not used anywhere. I think it might be related to anti-tracking feature changes. Instead of passing those ids to the array, requests should be marked.

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -333,32 +333,23 @@ function isPaused(details, request) {
 if (__PLATFORM__ === 'firefox') {
   chrome.webRequest.onBeforeRequest.addListener(
     (details) => {
+      if (details.tabId < 0) return;
```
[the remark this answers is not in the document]
smalluban: It's not a change here in PR, just unified style. If this is is a case, please make an issue to split changes into separate PRs.

extension-manifest-v3/src/background/reporting/index.js
```diff
@@ -137,11 +137,13 @@ if (__PLATFORM__ === 'firefox') {
       if (event === 'observed') {
         return;
       }
-      const request = Request.fromRawDetails({
+
+      const request = Request.fromRequestDetails({
         url: state.url,
-        sourceUrl: state.tabUrl,
+        originUrl: state.tabUrl,
```
chrmod, line 143: not a comment to a name change but in general - not sure if we should relay in `state.tabUrl` as it comes from `page-store` of the wtm/reporting webrequest-pipeline and we know that this one was a subject of same kind of state leakage. Yes, the latest version should work better, but maybe we should find another mechanism that will be consistent between all features.

extension-manifest-v3/src/background/adblocker.js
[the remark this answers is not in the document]
chrmod: I see. All good

extension-manifest-v3/src/background/stats.js
```diff
@@ -265,15 +265,20 @@ if (__PLATFORM__ === 'safari') {
         // with the correct tabId (sometimes it is correct, sometimes it is 0).
         // Thus, let the content_script fire it.
         case 'onCommitted': {
-          const { domain, hostname } = parse(sender.url);
-          setupTabStats(sender.tab.id, domain || hostname);
+          const request = Request.fromRequestDetails({ url: sender.url });
+          setupTabStats(
```
chrmod, line 269: now I start to wonder if this is not too late for non-safari browsers. couldn't it be the some webRequests for this navigation would fire before `onCommited`, taking Service Workers into consideration

extension-manifest-v3/src/background/stats.js
```diff
@@ -265,15 +265,20 @@ if (__PLATFORM__ === 'safari') {
...
+              : undefined,
+          );
           break;
         }
         case 'updateTabStats':
           updateTabStats(
             sender.tab.id,
             msg.urls.map((url) =>
-              Request.fromRawDetails({ url, sourceUrl: sender.url }),
+              Request.fromRequestDetails({ url, originUrl: sender.url }),
```
chrmod, line 281: shouldn't `originUrl` point to `sender.tab.url` instead? as `sender.url` is likely the iframe url, but I'm not sure if that's true

extension-manifest-v3/src/background/stats.js
chrmod: requestsBlocked are for the ad-blocking. anti-tracking marks requests as modified

extension-manifest-v3/src/background/reporting/index.js
smalluban: I think it is not related to this PR. Please "save the case for later" in your mind, or make an issue about it to not forget the problem.

extension-manifest-v3/src/background/stats.js
smalluban: The implementation has changed, please make another round of review and make the comment again if it still applies.

extension-manifest-v3/src/background/stats.js
smalluban: It's just difference between `fromRawDetails` and `fromRequestDetails` - as the second does not have `sourceUrl` but `originUrl` or `documentUrl` (in Firefox). I recommend to use one method everywhere, so we are consistent about generating request meta data, so I have changed it here to `fromRequestDetails`.

chrmod (approved):

PR #1335, opened 2023-10-13
Shim originUrl outside of Firefox

chrmod:


chrmod:
Closing for now as current methods may be sufficient

PR #1337, opened 2023-10-16
Refactor TabStats & AutoSyncingMap to limit requests and use session storage

smalluban:
* Refactor `AutoSyncingMap` to use `storage.session` and limit ttls to 1 day, and limit entries to 1000 (equals 1k open tabs, so it should be sufficient)
* Limits `tracker.requests` to 10 for each group (observer, blocked, modified)
* Adds `trackersCount` property and`blocked` and `modified` flags to `tracker` model
* Updates Panel UI
* Minor fix of the icon update (sometimes it did not clear the wheel after the navigation to the page without trackers)

### DIFF
--- a/extension-manifest-v3/src/background/stats.js
+++ b/extension-manifest-v3/src/background/stats.js
@@ -64,19 +64,17 @@ async function refreshIcon(tabId) {
     );
   }
 
-  if (data.path || data.imageData) {
-    // Note: Even in MV3, this is not (yet) returning a promise.
-    chromeAction.setIcon({ tabId, ...data }, () => {
-      if (chrome.runtime.lastError) {
-        console.debug(
-          'setIcon failed for tabId',
-          tabId,
-          '(most likely the tab was closed)',
-          chrome.runtime.lastError,
-        );
-      }
-    });
-  }
+  // Note: Even in MV3, this is not (yet) returning a promise.
+  chromeAction.setIcon({ tabId, ...data }, () => {
+    if (chrome.runtime.lastError) {
+      console.debug(
+        'setIcon failed for tabId',
+        tabId,
+        '(most likely the tab was closed)',
+        chrome.runtime.lastError,
+      );
+    }
+  });
 
   if (Options.trackerCount) {
     try {
@@ -101,12 +99,10 @@ function updateIcon(tabId) {
         delayMap.delete(tabId);
         refreshIcon(tabId);
       },
-      // Firefox flickers when updating the icon, so we should expand the throttle
+      // Firefox flickers when updating the icon, so we should expand the debounce delay
       __PLATFORM__ === 'firefox' ? 1000 : 250,
     ),
   );
-
-  refreshIcon(tabId);
 }
 
 export function updateTabStats(tabId, requests) {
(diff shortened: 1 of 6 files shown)


extension-manifest-v3/src/background/stats.js
```diff
@@ -253,7 +269,7 @@ function setupTabStats(tabId, request) {
     tabStats.delete(tabId);
   }
 
-  updateIcon(tabId);
+  updateIcon(tabId, true);
```
philipp-classen, line 272: Is the "true" parameter intentional? From the function signature, it has only argument as far as I can tell.

smalluban: Sorry, it's a left over when I was fixing the issue with icon (didn't help and was not needed). I'll clear it up.

extension-manifest-v3/src/background/stats.js
```diff
@@ -147,12 +143,34 @@ export function updateTabStats(tabId, requests) {
...
-          modified: request.modified,
-        });
+        if (!tracker.requests.some((r) => r.url === request.url)) {
+          tracker.requestsCount = (tracker.requestsCount || 0) + 1;
+          tracker.blocked = tracker.blocked || request.blocked;
+          tracker.modified = tracker.modified || request.modified;
+
+          let blocked = Number(request.blocked);
+          let modified = Number(request.modified);
+          let observed = !blocked && !modified ? 1 : 0;
```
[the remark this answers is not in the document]
smalluban: It depends on the platform. On Safari we can only observe, but they might be blocked (we just don't know that). On chromium we have blocked, but not modified (antytracking doesn't work).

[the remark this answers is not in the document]
smalluban: In short, observed is our the last resort if anything else doesn't match.

extension-manifest-v3/src/background/utils/map.js
```diff
@@ -69,7 +69,11 @@ export default class AutoSyncingMap {
     // and log it. A potential improvement could be to treat the
     // in-memory map as the source of truth in that scenario.)
     this._pending = new Promise((resolve, reject) => {
-      chrome.storage.local.get([this.storageKey], (result) => {
+      // Migration to session storage
+      // TODO: Remove this code after a few releases
+      chrome.storage.local.remove(this.storageKey);
```
philipp-classen, line 74: I wonder if it could be better in this case to move the migration out of the class? Maybe to the place where the class is initialized?

https://github.com/ghostery/ghostery-extension/blob/5607209f8297e27206852d3158cce623a0c8aa47/extension-manifest-v3/src/background/stats.js#L27

We currently lack a good place where to put migration into and the code that cleans up the storage has to live somewhere. One advantage of moving it out of the class is that AutoSyncingMap is potentially reusable, but the migration is only needed for the specific case of removing the "tabStats:v1" key.

smalluban: Current solution might have only impact on the startup, as remove is called when map is initialized. We can move it to the `stats.js` root module, but it still will be called each time background process starts.

After all, any other solution will be just for different code style. For me.keeping in the map implementation is the right place.

philipp-classen: We can keep it here. Though it will be called at each service worker start in both cases, since it is in the code path of the constructor.

extension-manifest-v3/src/background/utils/map.js
```diff
@@ -69,7 +69,11 @@ export default class AutoSyncingMap {
...
+
+      chrome.storage.session.get([this.storageKey], (result) => {
```
philipp-classen, line 76: After switching to session storage, please update the comment on the class ("with local storage from time to time" -> "with session storage from time to time")

philipp-classen (approved):

extension-manifest-v3/src/background/utils/map.js
chrmod, line 76: what's the status of chrome.storage.session on safari? do we have to bump the minimal OS requirements?

chrmod (approved):
Looks very good. My only worry is if `chrome.storage.session` will work on older versions of Safari.

extension-manifest-v3/src/background/utils/map.js
philipp-classen: chrome.storage.session support came with Safari 16.4 (Released 2023-03-27)

smalluban: We can make a fallback to `chrome.storage.local` if the `chrome.storage.session` is not available.

PR #1340, opened 2023-10-17
Optional storage session for tab stats

smalluban:
Fixes https://github.com/ghostery/ghostery-extension/pull/1337#discussion_r1361787002

* Hides implementation details of `AutoSyncingMap` into the static method of the class
* As the map is used not only in the background anymore, the `map.js` moved to `/utils`

Tested in iOS 15.5 Simulator. Indeed, there is just no `chrome.storage.session` object.


chrmod (approved):

extension-manifest-v3/src/store/tab-stats.js
```diff
@@ -66,9 +68,7 @@ const Stats = {
       const tab = await chrome.runtime.sendMessage({
         action: 'getCurrentTab',
       });
-
-      const storage = await chrome.storage.session.get(['tabStats:v1']);
-      const stats = tab && storage['tabStats:v1']?.entries[tab.id];
+      const stats = tab && (await AutoSyncingMap.get('tabStats:v1', tab.id));
```
philipp-classen, line 71: I was not aware that we are having direct read access through the native API and the AutoSyncingMap next to each other. As long as it is only used for the UI, it is probably okay to fetch snapshots natively, but it is potentially racy to mix native and wrapped APIs.

Below, we also have to be careful:
```
if (__PLATFORM__ !== 'safari') {
  chrome.storage.onChanged.addListener((changes) => {
    if (changes['tabStats:v1']) {
      store.clear(Stats, false);
    }
  });
}
```

I wonder if did fire on changes (with chrome.storage.local)? And does it work after switching to chrome.storage.session?

smalluban: Yes, the `chrome.storage.onChanged` fires correctly. I tested it out.

philipp-classen (approved):

PR #1361, opened 2023-10-30
Upgrade Node.js to 20.9.0

philipp-classen:


smalluban:
With https://github.com/ghostery/ghostery-extension/pull/1380 changes update node to 20.9.0 passes tests in v10, so the PR should be ready to merged after #1380.

PR #1364, opened 2023-11-02
MV3: fix cosmetic filter injection

chrmod:
The injection got broken on Safari due to removal of `onCommited` content script message in https://github.com/ghostery/ghostery-extension/pull/1333
But the MV3 implementation was never complete as the `adblocker` content script was injected to main frame only.

### DIFF
--- a/extension-manifest-v3/src/content_scripts/adblocker-scriptlets.js
+++ b/extension-manifest-v3/src/content_scripts/adblocker-scriptlets.js
@@ -1,0 +1,5 @@
+// Should only be needed on Safari:
+// the tabId of the initial chrome.webNavigation.onCommitted
+// is not reliable. When opening bookmarks, it can happen that
+// the event is associated with a tabId of 0.
+chrome.runtime.sendMessage({ action: 'onCommitted' });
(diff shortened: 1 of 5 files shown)


extension-manifest-v3/src/content_scripts/adblocker-scriptlets.js
```diff
@@ -0,0 +1,5 @@
+// Should only be needed on Safari:
+// the tabId of the initial chrome.webNavigation.onCommitted
+// is not reliable. When opening bookmarks, it can happen that
+// the event is associated with a tabId of 0.
+chrome.runtime.sendMessage({ action: 'onCommitted' });
```
philipp-classen, line 5: May it lead to double-injections?

chrmod: no, as this is loaded for Safari only and on Safari we listen to this message instead of listening to webNavigation https://github.com/ghostery/ghostery-extension/blob/f4ca329c15edaf073bd7297e6e2dfaddbb710d0e/extension-manifest-v3/src/background/adblocker.js#L303

smalluban, line 5: As this is used only for adblocker scriptlets, what do you think about renaming that event to something more specific like `injectScriplets` or `resolveScriptlets`?

smalluban (approved):

extension-manifest-v3/src/content_scripts/adblocker-scriptlets.js
chrmod: good idea!

philipp-classen (approved):

PR #1367, opened 2023-11-02
Safari: use single DNR list

chrmod:
Safari allows only 75k rules and does not support many DNR features like `requestDomains` or `initiatorDomains`. 

To save some space we merge 3 DNR lists for ads, trackers and annoyances into one, so we don't have to add exclusions and fixes multiple times. 

For now the cosmetic filters are still split but it in next iteration we may want to merge them into single engine. 

With those changes we were able to get a decent score in the [d3ward test](https://d3ward.github.io/toolz/adblock.html)

Screenshot 2023-11-02 at 22 25 38


smalluban (changes requested):
main vs this branch on my macOS 14.0, Safari 17.0
Zrzut ekranu 2023-11-3 o 09 46 55
Zrzut ekranu 2023-11-3 o 09 45 26

smalluban (approved):

PR #1380, opened 2023-11-13
Update npm dependencies in v10

smalluban:
Updates dependencies for v10, but as some packages are shared with v8, they are updated in both versions.

Both:
* `adblocker` and `adblocker-*`
* `@duckduckgo/autoconsent`
* `hybrids`
* `tldts-experimental`
* `web-ext`

v10 only:
* `vite`
* `jwt-decode`
* `@github/relative-time-element`


extension-manifest-v3/scripts/build.js
```diff
@@ -136,7 +136,7 @@ if (manifest.declarative_net_request?.rule_resources) {
 }
 
 // generate license file
-import('../../tools/licenses.js');
+execSync('node ../tools/licenses.js', { stdio: 'inherit' });
```
smalluban, line 139: We need to be sure that this process ends before rollup starts, as it creates `node_modules` path inside of the v10 folder on the fly, and clears it out at the end.

extension-manifest-v3/scripts/download-engines/index.js
```diff
@@ -11,7 +11,6 @@
 
 import { writeFileSync } from 'fs';
 import { resolve } from 'path';
-import fetch from 'node-fetch';
```
smalluban, line 14: Fetch is supported in Node v19

extension-manifest-v3/src/content_scripts/whotracksme/reporting.js
```diff
@@ -1,4 +1,4 @@
-import { recordMouseDown } from '@whotracksme/webextension-packages/packages/reporting/content-script';
+import { recordMouseDown } from '@whotracksme/webextension-packages/packages/reporting/src/request/content-script.js';
```
smalluban, line 1: There is an issue with using nested `package.json` with the newest rollup. I suggest flattening the `@whotracksme/webextension-packages` project to a single package (containing communication and reporting) - for now we also send a lot of garbage data to npm, like tests, examples, etc.. (they are not cleared when main package is created).

extension-manifest-v3/src/utils/api.js
```diff
@@ -9,7 +9,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
-import jwtDecode from 'jwt-decode';
+import { jwtDecode } from 'jwt-decode';
```
smalluban, line 12: The new version requires using name import.

extension-manifest-v3/src/content_scripts/whotracksme/reporting.js
smalluban: I think we "hacked" a little, and they might change implementation of the module resolution in rollup, so the nested package.json stops working (beside the main import, as it still works). 

The error message when using exports:
```
[vite]: Rollup failed to resolve import "@whotracksme/webextension-packages/packages/reporting/content-script" from "/Users/dominik/Workspace/ghostery-extension/extension-manifest-v3/src/background/reporting/index.js".
This is most likely unintended because it can break your application at runtime.
```

smalluban: I checked, and it looks, that nested paths from a package are resolved to the path first, so there is no `reporting/content-script` and it fails. The `.../reporting` exists, and then it checks that there is `package.json`, so it uses it.

However, a package inside a package is not normally supported case, and we should flatten it. For now, we can import file directly, as I've changed it.

philipp-classen (approved):

chrmod (approved):

Bankrollshow (commented):
- [ ]

Bankrollshow (commented):
- [ ]

extension-manifest-v3/scripts/build.js
Bankrollshow: - [ ]

PR #1387, opened 2023-11-17
Add Youtube wall popup to v10

smalluban:
* Moves the UI for the Youtube wall popup into the `packages/ui`
* Unify logic for v8 and v10
* Adds a condition for `showIframe` to detect if the iframe is already shown
* Adds `detectWall` function
* Hides popup when "Open YouTube in a private window" is clicked

The current logic detects changes in the DOM and checks if the URL has changed. If so, selectors for the wall are checked if exists, and then we trigger showing up the iframe. As the iframe blocks displaying more than once, it triggers each time the selectors are found.

For the notice, Safari `chrome.windows.create()` silently returns a resolved promise if the extension is not allowed in private mode. The `chrome.extension.inIncognitoContext` also always returns `false`.


smalluban:
Initial @kalinme testing failed.

extension-manifest-v3/src/content_scripts/youtube.js
```diff
@@ -0,0 +1,55 @@
...
+
+// INFO: Safari always returns false for `inIncognitoContext`
+if (!chrome.extension.inIncognitoContext) {
+  (async () => {
+    if (await isFeatureDisabled()) return;
+
+    window.addEventListener('yt-navigate-start', () => closeIframe(), true);
+
+    detectWall(async () => {
+      if (await isFeatureDisabled()) return;
```
smalluban, line 43: YT page is long living app, so the condition must be checked each time - not only at startup (applies to v8 & v10).

philipp-classen (approved):

chrmod (approved):
Looks good!

Bankrollshow (commented):
- [ ]

extension-manifest-v3/src/background/helpers.js
```diff
@@ -16,6 +16,12 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
...
       return true;
+    case 'openTabWithUrl':
+      chrome.tabs.create({ url: msg.url });
+      break;
+    case 'openPrivateWindowWithUrl':
+      chrome.windows.create({ url: msg.url, incognito: true });
+      break;
   }
 
   return false;
```
Bankrollshow, line 27: - [ ]

extension-manifest-v2/app/youtube/index.js
```diff
@@ -8,10 +8,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
+
```
Bankrollshow, line 11: -

Bankrollshow (commented):
- [ ]

PR #1424, opened 2023-12-18
Fix source url check for updating stats on Firefox

### LINKED ISSUE #1423: The panel doesn't display blocked or modified trackers
**Description:**
The Ghostery panel doesn't show blocked and/or modified trackers when a website is first accessed. When the website is reloaded, the panel shows the number of blocked and/or modified trackers.

**Steps to Reproduce:**
1. Open Ghostery Private Browser
2. Open website, for example https://www.wowhead.com/
3. Open Ghostery panel
image
4. Reload the website
5. Open Ghostery panel
image

**Expected Results:**

**Versions:**

	Browser: Ghostery Private Browser 2023.9 (#3e8eed2)
	OS: Android 14
	GBE 10.2.2

	Browser: Ghostery Private Browser 2023.9 (#3e8eed2)
	OS: Android 11
	GBE 10.2.2

smalluban:
Fixes #1423 

On Firefox when you type the URL by hand and navigate, for example, `onet.pl` - the browser navigates at first to `http://onet.pl`, which is saved as the URL of the tab (when stats are initialized). However, the final address is `https://www.onet.pl`. It has not only a different protocol, but also a subdomain added. 

When requests are checked after that navigation, the `request.isFromSourceUrl()` returns false, as those requests have the final URL as the source. After reloading the page, the `stats.url` is set to correct value, as the tab is already at the "final" URL.

~~This PR fixes the issue by loosening the condition, to check if the origin (domain or hostname) is a part of the source URL (ancestors, initiator, or sourceUrl). I am aware that this is not 100% correct, as the pathname or query might contain the domain, but in this case, this is just a protection to not pass through requests from the "previous page" navigation - which is very unlikely to happen. Even using parsed URLs might not be sufficient, as the case with added subdomains would fail (like the example of `onet.pl` -> `www.onet.pl`).~~

EDIT:

The final solution chooses a URL to check, parses it, and then compares the domain or hostname. It should now be 100% complaint. 

The URL comes from a different part of the original request, so it is not parsed before, we have to do it within the function.

### DIFF
--- a/extension-manifest-v3/src/background/utils/request.js
+++ b/extension-manifest-v3/src/background/utils/request.js
@@ -12,6 +12,24 @@
 import { Request } from '@cliqz/adblocker';
 import { parse } from 'tldts-experimental';
 
+const PARSE_CACHE_LIMIT = 1000;
+const parseCache = new Map();
+
+function parseWithCache(url) {
+  if (parseCache.has(url)) {
+    return parseCache.get(url);
+  }
+
+  if (parseCache.size > PARSE_CACHE_LIMIT) {
+    parseCache.clear();
+  }
+
+  const parsed = parse(url);
+  parseCache.set(url, parsed);
+
+  return parsed;
+}
+
 export default class ExtendedRequest extends Request {
   static fromRequestDetails(details) {
     const isMainFrame = details.type === 'main_frame';
@@ -19,8 +37,8 @@ export default class ExtendedRequest extends Request {
       ? details.url
       : details.originUrl || details.documentUrl || '';
 
-    const parsedUrl = parse(details.url);
-    const parsedSourceUrl = isMainFrame ? parsedUrl : parse(sourceUrl);
+    const parsedUrl = parseWithCache(details.url);
+    const parsedSourceUrl = isMainFrame ? parsedUrl : parseWithCache(sourceUrl);
 
     return new ExtendedRequest({
       requestId: details.requestId,
@@ -76,7 +94,7 @@ export default class ExtendedRequest extends Request {
     }
 
     if (url) {
-      const parsedUrl = parse(url);
+      const parsedUrl = parseWithCache(url);
       return parsedUrl.domain === domain || parsedUrl.hostname === domain;
     }
 
--- a/extension-manifest-v3/src/background/utils/request.js
+++ b/extension-manifest-v3/src/background/utils/request.js
@@ -53,27 +53,31 @@ export default class ExtendedRequest extends Request {
     this.sourceHostname = data.sourceHostname;
   }
 
-  isFromOriginUrl(url) {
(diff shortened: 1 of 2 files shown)


smalluban:
@philipp-classen Please review this PR, @chrmod had some doubts about loosening the condition, but truthfully I cannot find another working solution.

extension-manifest-v3/src/background/utils/request.js
```diff
@@ -53,17 +53,17 @@ export default class ExtendedRequest extends Request {
...
     /* Firefox APIs */
 
     if (frameAncestors && frameAncestors.length > 0) {
-      return url === frameAncestors[frameAncestors.length - 1].url;
+      return frameAncestors[frameAncestors.length - 1].url?.includes(origin);
     }
 
     if (this.sourceUrl) {
-      return url === this.sourceUrl;
+      return this.sourceUrl.includes(origin);
```
philipp-classen, line 66: Would `startsWith` be still correct (and more precise)?

smalluban: Sadly nope, I even gave an example in the description, which was a case where the bug was found. The sourceUrl is `https://www.onet.pl`, but the `url` from the original code was `http://onet.pl` (where we saved it as `onet.pl`). This is way we cannot compare full URLs as they are different (in case of Firefox, where it dispatch an navigation event even for the redirect).

chrmod (changes requested):
How about we compare eTLD+1 instead?

smalluban:
> How about we compare eTLD+1 instead?

No idea what you mean, does it return `true` for a pair of `http://onet.pl` and `https://www.onet.pl` (mind a different protocol and subdomain)?

chrmod:
yes, effective top level domain for those two urls is the same - `onet.pl`.

chrmod:
you can use `tldts` or `cliqz/fast-url-parser`

extension-manifest-v3/src/background/utils/request.js
```diff
@@ -53,27 +53,31 @@ export default class ExtendedRequest extends Request {
...
-    /* Chrome APIs */
-
-    const { frameType, initiator } = this._originalRequestDetails;
-
-    // For frameType 'sub_frame', we can't determine the origin URL
-    // as it might be the iframe itself or any of its ancestors
-    if (frameType === 'outermost_frame' && initiator) {
-      return url.startsWith(initiator);
+    if (url) {
+      const parsedUrl = parse(url);
```
chrmod, line 79: we will likely parse the same URL over and over again. One way to do it would be to add some in-memory cache. This is what we did in V8 - https://github.com/ghostery/common/blob/main/modules/core/sources/LRU.ts

chrmod (approved):

PR #1440, opened 2024-01-11
Release 10.2.9

smalluban:
2ce762916653771cf45686b92a62c61fb016e82e Add workaround to start background process in Safari (#1439)
67e021808c46a5e339598cf176d88f1f589c08cd Fix pin-it position for Opera One UI redesign (#1438)

@GRadziejewski Please make a fast sanity check on iOS and macOS

We have found that resetting Safari is a sufficient way to break not running background process. the #1439 introduced a hacky way to fix this (temporary solution).

TestFlight: v10.2.9 (140)


GRadziejewski (approved):
Tests results: https://github.com/ghostery/ghostery-extension/issues/1441#issuecomment-1887550189

PR #1448, opened 2024-01-29
Update autoconsent to 9.4.0

seia-soto:
There's no breaking changes from version 8.1 to 9.4.

- https://github.com/duckduckgo/autoconsent/compare/v8.1.0...v9.4.0

The configuration change includes the following PR but GBE doesn't require any change:

- https://github.com/duckduckgo/autoconsent/pull/332

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
         "@cliqz/adblocker": "^1.26.12",
         "@cliqz/adblocker-webextension": "^1.26.12",
         "@cliqz/url-parser": "^1.1.5",
-        "@duckduckgo/autoconsent": "^8.1.0",
+        "@duckduckgo/autoconsent": "^9.4.0",
         "@ghostery/libs": "^1.0.0",
         "@ghostery/trackers-preview": "^1.0.0",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -865,11 +875,6 @@
         "node": ">=10.0.0"
       }
     },
-    "node_modules/@duckduckgo/autoconsent": {
```
chrmod, line 868: please ensure, the autoconsent is included once in the package-lock.json.

if you don't find a better way with npm cli, then you can always update package-lock by hand. in this case it should be easy

chrmod (changes requested):

seia-soto:
Just a note for future contribution: running the same command twice will optimize the dependency tree.

```sh
npm install @duckduckgo/autoconsent@9.4.0 --workspace extension-manifest-v3 --workspace extension-manifest-v2 --save
```

package-lock.json
seia-soto: Successfully optimized the tree.

chrmod (approved):

PR #1512, opened 2024-03-25
Selective blocking

smalluban:
WIP.

Usage of the `TrackerException` model in the background:

```js
import TrackerException from '/store/tracker-exception.js';

// Resolve a full list of the models
await store.resolve([TrackerException])

// Resolve model by id
// The definition always returns a model with predefined values,
// but they are not saved until there is a `store.set()` call
// Still, I think this case should be used rarely in the BG
await store.resolve(TrackerException, 'trackerId...')
```


extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -62,3 +91,14 @@ export async function getPattern(key) {
 
   return patterns.get(key);
 }
+
+export function isCategoryBlockedByDefault(categoryId) {
+  switch (categoryId) {
```
chrmod, line 96: the blocking categories are:
* advertising
* porn_advertising
* email
* site_analytics

smalluban: This condition is the opposite - by default all categories are blocked except from listed here. Is there missing there something? Essential and CDN? Something more?

chrmod: yes, quite a few
https://github.com/ghostery/trackerdb/tree/main/db/categories

Not sure what's better, but I think it is more likely we add a new non-blocking category.

chrmod:
@smalluban please check if selective blocking UI works for trackers from unidentified category.

chrmod (dismissed):
Amazing!

I'm all for merging it in this state as it is already too big to handle. 

Few things will require to be addressed before it can be released, for example Safari support.

extension-manifest-v3/src/utils/offscreen.js
```diff
@@ -9,13 +9,14 @@
...
-
-// Components
-define.from(import.meta.glob('./**/*.js', { eager: true, import: 'default' }), {
-  prefix: 'ui-settings',
-  root: 'components',
-});
+export async function requestPermission() {
+  if (!(await chrome.permissions.contains(permissionRequest))) {
+    if (!(await chrome.permissions.request(permissionRequest))) {
+      throw new Error('Ghostery requires "offscreen" permission');
```
philipp-classen, line 19: Should this throw if both are false (like it is now) or if one of condition is not met?

extension-manifest-v3/tests/utils/trackerdb.test.js
```diff
@@ -0,0 +1,100 @@
...
+import { isTrusted } from '../../src/utils/trackerdb.js';
+
+describe('/utils/trackerdb.js', () => {
+  describe('#isTrusted', () => {
+    let exception;
+
+    beforeEach(() => {
+      exception = {
+        blocked: false,
+        blockedDomains: ['foo.org'],
```
philipp-classen, line 24: This is a bit of a weird setup, having a domain trusted and blocked at the same time. Is such a case possible to reach in production?

philipp-classen (approved):

PR #1519, opened 2024-04-02
Adblocker preprocessor support

chrmod:
Requires: https://github.com/ghostery/common/pull/96


extension-manifest-v3/src/utils/engines.js
```diff
@@ -191,7 +203,7 @@ async function update(name) {
 
       const engineBytes = new Uint8Array(arrayBuffer);
       engine = FiltersEngine.deserialize(engineBytes);
-
+      engine.updateEnv(ENV);
```
smalluban, line 206: As we add some layers to the `FiltersEngine`, I would rather create a custom `deserialize` function, and put `updateEnv` call there. It would be then only in one place.

smalluban (changes requested):

extension-manifest-v3/src/utils/engines.js
```diff
@@ -22,6 +22,17 @@ import { registerDatabase } from '/utils/indexeddb.js';
 
 export const CUSTOM_ENGINE = 'custom-filters';
 
+const checkUserAgent = (pattern) => navigator.userAgent.indexOf(pattern) !== -1;
+
+const ENV = new Map([
```
smalluban, line 27: This can be then inlined rather than global in the module scope.

chrmod: it is used after `deserialise` and for `updateFromDiff`

extension-manifest-v3/src/utils/engines.js
```diff
@@ -22,6 +22,17 @@ import { registerDatabase } from '/utils/indexeddb.js';
...
+  ['ext_ghostery', true],
+  ['cap_html_filtering', true],
+  ['env_firefox', checkUserAgent('Firefox')],
+  ['env_chromium', checkUserAgent('Chrome')],
+  ['env_edge', checkUserAgent('Edg')],
+  ['env_mobile', checkUserAgent('Mobile')],
```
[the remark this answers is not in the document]
chrmod: it's correct https://github.com/gorhill/uBlock/blob/98a600698e25533ea1bf9c228885791cee5fa636/platform/common/vapi-common.js#L173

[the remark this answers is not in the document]
smalluban: Ok

extension-manifest-v3/src/utils/engines.js
```diff
@@ -22,6 +22,23 @@ import { registerDatabase } from '/utils/indexeddb.js';
...
+const ENV = new Map([
+  ['ext_ghostery', true],
+  ['cap_html_filtering', checkUserAgent('Firefox')],
+  ['env_firefox', checkUserAgent('Firefox')],
+  ['env_chromium', checkUserAgent('Chrome')],
+  ['env_edge', checkUserAgent('Edg')],
+  ['env_mobile', checkUserAgent('Mobile')],
+]);
+
+const deserializeEngine = (engineBytes) => {
```
smalluban, line 36: Please use `function` syntax, it's not a variable.

extension-manifest-v3/src/utils/engines.js
```diff
@@ -378,6 +393,7 @@ export async function createCustomEngine(filters = '') {
     enableHtmlFiltering: true,
   });
   const engine = FiltersEngine.parse(filters, config);
+  engine.updateEnv(ENV);
```
smalluban, line 396: I would create another `parseEngine` function to use the same pattern, that updateEnv is not used directly.

smalluban (changes requested):

smalluban (approved):

PR #1529, opened 2024-04-09
Selective blocking: offscreen document

chrmod:


extension-manifest-v3/src/utils/dnr-converter.js
```diff
@@ -9,45 +9,114 @@
...
+        action: 'offscreen:urlfitler2dnr:convert',
+        filter,
+      })) || { errors: ['failed to innitiate offscreen document'], rules: [] }
+    );
+  }
+
+  return convert;
 }
+
+export default chrome && chrome.runtime.getManifest?.().manifest_version === 3
```
smalluban, line 120: The idea is perfect, but it should be done on the build step, and it is not strictly related to the `manifest_version`, rather the platform. Only chromium has a permission to `offscreen` so I suppose, that only it supports it. The Safari supports version 3, which I would like to migrate to (and I am not sure if you will have offscreen document there).


```suggestion
export default __PLATFORM__ !== 'safari' && __PLATFORM__ !== 'firefox'
  ? createOffscreenConverter()
  : createDocumentConverter()
```

Mind the new function naming - now they are not related to mv... rather what they actually do...

extension-manifest-v3/src/background/exteptions.js
```diff
@@ -0,0 +1,90 @@
+/**
```
smalluban, line 1: Name has a typo - `exteptions` -> `exceptions`

extension-manifest-v3/src/utils/dnr-converter.js
chrmod: cannot be done on a build as this is used inside of the offscreen document, but I will update the check

smalluban: It does not matter, platform is a primary condition, and it's always available.. and it's a build time check.

extension-manifest-v3/scripts/build.js
```diff
@@ -183,6 +183,20 @@ if (manifest.browser_action?.default_popup) {
...
 
+// offscreen documents
+if (
+  manifest.permissions.includes('offsceen') ||
+  manifest.optional_permissions.includes('offsceen')
+) {
+  readdirSync(join(options.srcDir, 'pages', 'offscreen'), {
+    withFileTypes: true,
+  })
+    .filter((dirent) => dirent.isDirectory() && !dirent.name.startsWith('.'))
```
philipp-classen, line 194: What should be filtered by starsWith('.')? Is it only to exclude `.` and `..`?

(I'm not sure if it specified if you get directories like this `path/...` or that `./path/...` without normalizing. That might depend on factors like the OS and the current working directory.)

extension-manifest-v3/src/background/exceptions.js
```diff
@@ -0,0 +1,90 @@
...
+    // * blocking without overwrite = nothing
+    // * non-blocking with overwrite = nothing
+    // * non-blocking without overwrite = block
+    // website allowed:
+    // * blocking with overwrite = nothing
+    // * blocking without overwrite = allow
+    // * non-blocking with overwrite = allow
+    // * non-blocking without overwrite = nothing
+
+    // WIP
```
philipp-classen, line 65: Will this change, or is it an outdated comment now?

extension-manifest-v3/src/pages/offscreen/urlfilter2dnr/index.js
```diff
@@ -0,0 +1,9 @@
+import convert from '/utils/dnr-converter.js';
+
+chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
+  if (msg.action === 'offscreen:urlfitler2dnr:convert') {
+    convert(msg.filter).then((result) => sendResponse(result));
```
philipp-classen, line 5: What if "convert" throws?

(If there is nothing better, then adding a ".catch(console.error)" could be useful. Even if convert should never fail.)

extension-manifest-v3/src/background/exceptions.js
chrmod: this will be addressed in next PRs

extension-manifest-v3/src/pages/offscreen/urlfilter2dnr/index.js
chrmod: it never throws, returns `errors` instead

extension-manifest-v3/scripts/build.js
chrmod: it's meant to ignore "hidden" folders. `..` is not returned by `readdirSync`

PR #1531, opened 2024-04-10
Ghostery 10: staging setup

### LINKED ISSUE #1521: Ghostery 10: staging setup
Ghostery 8 offered developer build that connected to staging infrastructure for Ghostery backends. That includes:
* website: www.ghosterystage.com
* accounts: accounts.ghosterystage.com and account.ghosterystage.com
* cdn: staging-cdn.ghostery.com
* search: staging.ghosterysearch.com

The choice of the backend url was handled in runtime by checking for `debug` flag in the extension manifest, so the setup was forced at the build time. 

If possible, a true runtime check would be more beneficial for testing purposes - a user (tester) should be able to switch the backends to the staging environment at any time.

chrmod:
fix #1521 

As Manifest V3 requires ruleset to be part of the build, we are unable to update them in the runtime. That means, staging build needs to include rulesets from the staging CDN.


packages/ui/src/modules/onboarding/views/main.js
```diff
@@ -16,8 +16,9 @@ import Skip from './skip.js';
 import OutroSkip from './outro-skip.js';
 import OutroSuccess from './outro-success.js';
 
-const TERMS_AND_CONDITIONS_URL =
-  'https://www.ghostery.com/privacy/ghostery-terms-and-conditions?utm_source=gbe';
+const TERMS_AND_CONDITIONS_URL = `https://www.${
+  chrome.runtime.getManifest().debug ? 'ghosterystage' : 'ghostery'
```
chrmod, line 20: @smalluban this patter repeats. Do you think it is worth to encapsulate this it?

extension-manifest-v3/package.json
```diff
@@ -4,11 +4,12 @@
   "version": "10.2.15",
   "type": "module",
   "scripts": {
-    "postinstall": "npm run download-engines && npm run download-whotracksme-bloomfilter",
```
chrmod, line 7: this change makes download a part of the build. If we want to skip the download, it's probably best to check checksums of the local resources against the remote ones. Let me know if you feel this should be addressed in this PR.

extension-manifest-v3/scripts/build.js
```diff
@@ -67,6 +67,14 @@ if (argv.target === 'safari') {
   options.assets.push('background/safari');
 }
 
+if (argv.staging) {
+  execSync('npm run download.staging', { stdio: 'inherit' });
```
smalluban, line 71: Import and run download explicitly, not as an external command

extension-manifest-v3/scripts/build.js
```diff
@@ -67,6 +67,14 @@ if (argv.target === 'safari') {
...
+
+  manifest.debug = true;
+} else {
+  execSync('npm run download', { stdio: 'inherit' });
```
smalluban, line 75: Run this as an implicit download function - not an external command.

extension-manifest-v3/package.json
```diff
@@ -4,11 +4,12 @@
...
-    "build": "node scripts/build.js",
-    "start": "npm run build -- --watch",
     "download-engines": "node scripts/download-engines/index.js",
     "download-whotracksme-bloomfilter": "node scripts/download-whotracksme-bloomfilter.js",
+    "download": "npm run download-engines && npm run download-whotracksme-bloomfilter",
```
smalluban, line 9: I am fine with moving downloading into the build step. As we will have it there, I would strongly advice to refactor those scripts, so they will be "run" inside of the JS, rather than commands - they are no longer needed here. 

As you added `staging` argument to build command, we can also add `--skip-download`, or even flip it over, and add `--download` - so we run this excplicitly on CI, or `release` command, but not all the time.

extension-manifest-v3/src/utils/api.js
```diff
@@ -11,7 +11,9 @@
 
 import { jwtDecode } from 'jwt-decode';
 
-const DOMAIN = 'ghostery.com';
+const DOMAIN = chrome.runtime.getManifest().debug
```
smalluban, line 14: To fix repetition problem, I suggest to change it to

```suggestion
export GHOSTERY_DOMAIN = chrome.runtime.getManifest().debug
  ? 'ghosterystage.com'
  : 'ghostery.com';
```

Then we can just use it where it is needed:

```js
import { GHOSTERY_DOMAIN } from "/utils/api";
```

smalluban (changes requested):
Please utilize `GHOSTERY_DOMAIN` and simplify build script (clean up download commands)

extension-manifest-v3/package.json
chrmod: I would keep them as independent scripts for sake of local development

extension-manifest-v3/scripts/build.js
chrmod: I'm fine with having calling those as javascript function but lets address this in separate PR.

extension-manifest-v3/scripts/build.js
smalluban: Please add an option to skip download or enable it for build (I think the latter is better)

smalluban (changes requested):

extension-manifest-v3/src/background/telemetry.js
```diff
@@ -76,10 +76,12 @@ chrome.runtime.onMessage.addListener((msg) => {
 
 (async () => {
   const storage = await loadStorage();
-
+  const { version, debug } = chrome.runtime.getManifest();
   telemetry = new Telemetry({
-    METRICS_BASE_URL: 'https://d.ghostery.com',
-    EXTENSION_VERSION: chrome.runtime.getManifest().version,
+    METRICS_BASE_URL: debug
+      ? 'staging-d.ghostery.com'
```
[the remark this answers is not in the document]
chrmod: good catch!

chrmod:
@smalluban instead of optional downloads I've made the checksum check so we don't download if there are no changes. This will also ensure we don't use staging rules in production.

Issue #1523, opened 2024-04-05
[QA] Tests of Release 10.2.16


GRadziejewski:
Tests of: https://github.com/ghostery/ghostery-extension/pull/1522

GRadziejewski:
## Smoke tests Iteration no 1:


- 

## Issues:
-  

## Fixed:
- 



**Tests are finished at 14:15 28.03.2024.**

|| Feature 									| Browser | Browser | Browser | 
|:--- | 	 :---								| :---: | :---: | :---: | 
| NO | X  										|  Windows 10 </br> Opera | macOS </br> Safari |   iOS </br> Safari |
|01| Install fresh (no profiles etc) 					|
|02| Update 									|
|03| Ghostery cannot work before onboarding  		|
|04| Onboarding 								|
|05| Enable/Disable Ghostery 					|️
|06| Enable/Disable Ad-blocking  					|
|07| Enable/Disable Anti-tracking  					|
|08| Pause Ghostery  							|
|09| Never Consent  							|
|10| Trackers Preview  							|
|11| Restart the browser							|
|12| Licences 									|
|13| Language (ENG, FR, etc)						|
|14| Uninstall survey							|
|15| Sign in/out  								|
|16| &emsp; Sign in and then update  				|
|17| &emsp; Check Sync from / into another device	|
|18| &emsp;&emsp; Check Ghostery Settings logs	|
|19| &emsp;&emsp; Check Ghostery Background logs 	|
|20| &emsp; Check Ghostery account settings		|
|22| Check settings								|
|23| Log into Google account and open google.com 	|
|24| &emsp; Try to search for anything				|
|25| &emsp; Try to search something with video		|

```
Browser: Opera One 109.0.5097.24
OS: Windows 10 Pro
GBE 10.2.16

Browser: Safari 17.4.1 (19618.1.15.11.14)
OS: macOS Ventura 14.4.1
GBE 10.2.15 (152)

Browser: Safari  
OS: iOS 17.4.1 (iPhone 13)
GBE 10.2.15 (152)
```

Legend:
- ✔️ - Checked and working
- ❌ - Issue
- 🟦 - Not available 
- 🟨 - Not tested

GRadziejewski:
## Smoke tests Iteration no 2:


- 

## Issues:
-  

## Fixed:
- 



**Tests are finished at 10:35 08.04.2024.**

|| Feature 									| Browser | Browser | Browser | 
[...]

PR #1537, opened 2024-04-15
Remove duplicated ghostery domain suffix in use

smalluban:
Fixes https://github.com/ghostery/ghostery-extension/issues/1523#issuecomment-2056449696


smalluban:
Already done in #1536

PR #1538, opened 2024-04-15
Adblocker: integrate fixes ruleset and engines

chrmod:


extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -11,19 +11,23 @@
...
-let promise = engines.init('trackerdb').then(() => {
+let promise = engines.init(engines.TRACKERDB_ENGINE).then(() => {
   promise = null;
 });
 
-export async function getMetadata(request) {
-  if (promise) await promise;
+export function getMetadata(request) {
+  if (promise) {
+    console.warn('TrackerDB not ready yet');
```
philipp-classen, line 20: Good to have a warning on that. Do you expect it to be triggering often? (And if so, what functionality is impacted during that time?)

extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -94,11 +99,12 @@ export async function getPattern(key) {
 
 export function isCategoryBlockedByDefault(categoryId) {
```
smalluban, line 100: Are you sure that only 4 categories are blocked by default?

extension-manifest-v3/src/utils/engines.js
```diff
@@ -21,6 +21,8 @@ import {
 import { registerDatabase } from '/utils/indexeddb.js';
 
 export const CUSTOM_ENGINE = 'custom-filters';
+export const FIXES_ENGINE = 'fixes';
+export const TRACKERDB_ENGINE = 'trackerdb';
```
smalluban, line 25: Not sure if we need to specify trackerDB name as an export. It is used only in `trackerdb.js`. Not here at all.

extension-manifest-v3/src/utils/trackerdb.js
chrmod: yups, I'm positive

extension-manifest-v3/src/utils/trackerdb.js
chrmod: all engines should start as soon as possible. On MV3 the consequence of delayed trackerdb startup would be lack of detailed stats in the panel in MV2 some trackers may not be blocked - but there is nothing we can do about it.

extension-manifest-v3/src/utils/engines.js
smalluban: Ok

PR #1541, opened 2024-04-16
Selective blocking for Firefox

chrmod:


extension-manifest-v3/src/background/exceptions.js
```diff
@@ -30,15 +30,23 @@ function negateFilter(filter) {
...
+chrome.storage.local.get(['exceptions']).then((result) => {
+  exceptions = new Map(Object.entries(result.exceptions));
+});
+
 chrome.storage.onChanged.addListener(async (changes) => {
   if (!changes['exceptions']) {
     return;
   }
 
+  exceptions = new Map(Object.entries(changes['exceptions'].newValue || {}));
```
smalluban, line 44: This is overkill, as the exceptions is a plain object with keys, no need to transfer it to the Map instance.

extension-manifest-v3/src/background/exceptions.js
```diff
@@ -30,15 +30,23 @@ function negateFilter(filter) {
...
+
   const networkFilters = [];
   const cosmeticFilters = [];
 
-  for (const exception of changes['exceptions'].newValue || []) {
+  for (const exception of Object.values(changes['exceptions'].newValue || {})) {
```
smalluban, line 49: Use `exceptions` here which you defined above

extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -103,6 +103,8 @@ export function isCategoryBlockedByDefault(categoryId) {
     case 'adult_advertising':
     case 'email':
     case 'site_analytics':
+    case 'unidentified':
```
smalluban, line 106: I am not sure, but I think we removed that category, and everything, which does not fit its set to `misc`

smalluban (changes requested):

extension-manifest-v3/src/utils/trackerdb.js
chrmod: the extension code sets the `unidentified` category for unknown trackers

PR #1544, opened 2024-04-17
Ghostery 8: fix for missing ghostery_id

chrmod:
more and more TrackerDB patterns come without field `ghostery_id` - Ghostery 8 did not take it into account and was considering all of them alike as having empty value. This could manifest as https://github.com/ghostery/trackerdb/issues/266


philipp-classen (approved):

philipp-classen:
Confirmed in a local build that it fixes the problem in https://github.com/ghostery/trackerdb/issues/266

PR #1546, opened 2024-04-18
Update metadata creation for unidentified trackers

smalluban:


extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -32,10 +34,48 @@ export function getMetadata(request) {
...
+    matches = [
+      {
+        pattern: { key: exception.id, name: exception.id },
+        category: { key: 'unidentified' },
+      },
+    ];
   }
 
   const { category, pattern, organization } = matches[0];
+  const blockByDefault = isCategoryBlockedByDefault(category.key);
```
smalluban, line 49: I started with updating stats, but after all this place generates metadata for identifying tracker, and it was also used to now in `adblocker.js` if should block the request, so I moved it here.

If the `shouldBlock` is not required for DNR platforms, we can add here a platform check (not sure if we don't need it there).

smalluban: After all each request was checked in `shouldBlockRequest` in the previous version.

extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -32,10 +34,48 @@ export function getMetadata(request) {
   }
 
   if (matches.length === 0) {
-    return null;
+    exception = getException(request.domain);
+    if (!exception) return null;
+
+    matches = [
+      {
+        pattern: { key: exception.id, name: exception.id },
```
chrmod, line 42: would rather use `request.domain` instead of `exception.id`

smalluban: Ok

chrmod (approved):

PR #1547, opened 2024-04-18
MV2: Trust logic

chrmod:
The trust logic is following:
* trackers are blocked with adblocker filters
* when trusted, matching by domain will also be used to let them pass
* non trackers, matched by domain and adblocker filters are not blocked
* when blocked, only filters are applied

### DIFF
--- a/extension-manifest-v3/src/utils/trackerdb.js
+++ b/extension-manifest-v3/src/utils/trackerdb.js
@@ -9,14 +9,44 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
-import * as engines from './engines.js';
-import { getException } from '/background/exceptions.js';
-
 import { order as categoryOrder } from '@ghostery/ui/categories';
+
+import * as engines from './engines.js';
+import { getException } from '../background/exceptions.js';
 
 let promise = engines.init(engines.TRACKERDB_ENGINE).then(() => {
   promise = null;
 });
+
+export function isCategoryBlockedByDefault(categoryId) {
+  switch (categoryId) {
+    case 'advertising':
+    case 'pornvertising':
+    case 'email':
+    case 'site_analytics':
+    case 'unidentified':
+    case undefined:
+      return true;
+    default:
+      return false;
+  }
+}
+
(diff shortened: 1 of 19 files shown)


extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -50,29 +50,30 @@ export function getMetadata(request) {
...
-      shouldBlock = false;
+      isTrusted = true;
     }
 
     // If domain is blocked and global status is to allow
-    // set shouldBlock to true
+    // set isTrusted to false
     if (
-      (shouldBlock === false || !blockByDefault) &&
+      (isTrusted === false || !blockByDefault) &&
```
philipp-classen, line 73: This looks at least redundant, potentially wrong:

```
if ((isTrusted === false || <...>) && <...>) { 
  isTrusted = false;
}
```

philipp-classen: It could be simplified to:

```
if (!blockByDefault && exception.blocked.includes(request.tab.domain)) {
  isTrusted = false;
}
```

smalluban (approved):

extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -9,15 +9,45 @@
...
+      return false;
+  }
+}
+
+export function isTrusted(request, category, exception) {
+  const blockByDefault = isCategoryBlockedByDefault(category);
+
+  let isTrusted = !blockByDefault;
+
+  if (exception) {
```
philipp-classen, line 40: I find it hard to understand this logic.

overwriteStatus is a boolean, from what I see. To me, it is not intuitive what `overwriteStatus=true` or `overwriteStatus=false` means.

The semantic of "isTrusted" might also probably be rather "should not block". (I would expect that "trusted" implies that there was an action from the user before, namely to trust the tracker or the page.)

chrmod: I'm also not super happy with the naming - lets have a discussion.

PR #1549, opened 2024-04-19
Mv3-feat-selective-blocking-is-trusted

smalluban:
* Request was only used to access `.tab.domain` so we can pass it instead - it simplifies a much tests and eliminates import problem
* The logic was still not quire right...

### DIFF
--- a/extension-manifest-v3/src/utils/trackerdb.js
+++ b/extension-manifest-v3/src/utils/trackerdb.js
@@ -32,20 +32,16 @@ export function isCategoryBlockedByDefault(categoryId) {
   }
 }
 
-export function isTrusted(request, category, exception) {
-  const blockByDefault = isCategoryBlockedByDefault(category);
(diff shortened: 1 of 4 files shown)


extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -32,20 +32,16 @@ export function isCategoryBlockedByDefault(categoryId) {
...
-
-  if (exception) {
-    if (exception.overwriteStatus) {
-      isTrusted = blockByDefault === exception.overwriteStatus;
-    }
-    isTrusted = isTrusted
-      ? !exception.blocked.includes(request.tab.domain)
-      : exception.allowed.includes(request.tab.domain);
+export function isTrusted(domain, category, exception) {
+  const blockedByDefault =
```
chrmod, line 36: this name should rather be `shouldBlock`

chrmod (approved):

PR #1555, opened 2024-04-25
Mv3-feat-selective-blocking-offscreen

smalluban:
From my understanding:

* The Offscreen converter is only required in Chromium in the background process. Otherwise, we can safely create an iframe (there are a number of document-based contexts)
* We can avoid messing up with building offscreen documents where it is not required (for example, it was like this before the change in the settings page)
* As it is related to the platform we can simplify access to converter, and clean up requiring permissions when it can be skipped

### DIFF
--- a/extension-manifest-v3/src/pages/offscreen/urlfilter2dnr/index.js
+++ b/extension-manifest-v3/src/pages/offscreen/urlfilter2dnr/index.js
@@ -1,4 +1,6 @@
-import convert from '/utils/dnr-converter.js';
+import { createDocumentConverter } from '/utils/dnr-converter.js';
+
+const convert = createDocumentConverter();
 
 chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
   if (msg.action === 'offscreen:urlfitler2dnr:convert') {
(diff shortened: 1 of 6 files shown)


extension-manifest-v3/src/pages/offscreen/urlfilter2dnr/index.js
```diff
@@ -1,4 +1,6 @@
-import convert from '/utils/dnr-converter.js';
+import { createDocumentConverter } from '/utils/dnr-converter.js';
+
+const convert = createDocumentConverter();
```
smalluban, line 3: This is a offscreen page which needs to create a document converter, no case where offscreen document should create another offscreen document

extension-manifest-v3/src/pages/settings/components/custom-filters.js
```diff
@@ -12,10 +12,12 @@
 import { html, store } from 'hybrids';
 import { detectFilterType } from '@cliqz/adblocker';
 
-import convert from '/utils/dnr-converter.js';
+import { createDocumentConverter } from '/utils/dnr-converter.js';
```
smalluban, line 15: Settings pages should always use Document Converter

chrmod (approved):
Looks good

PR #1580, opened 2024-05-07
Create fallback tracker in the trackerdb utility

smalluban:
* Passes through requests without exception where category should not be blocked
* Moves fallback tracker for unidentified trackers to trackerdb - There is a bug when you add exception to the request, which is not in the trackerdb - then the metadata must be created there, as from now the metadata is returned (but without required info)

### DIFF
--- a/extension-manifest-v3/src/utils/trackerdb.js
+++ b/extension-manifest-v3/src/utils/trackerdb.js
@@ -55,6 +55,8 @@ export function getMetadata(request) {
 
   let isFilterMatched = true;
   let exception = null;
+  let tracker;
+
   let matches = engine.getPatternMetadata(request);
 
   if (matches.length === 0) {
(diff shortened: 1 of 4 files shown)


extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -64,30 +66,35 @@ export function getMetadata(request) {
...
-        category: { key: 'unidentified' },
-      },
-    ];
+    if (!exception && !request.blocked && !request.modified) {
+      return null;
+    }
+
+    tracker = {
+      id: request.domain,
+      name: request.domain,
```
chrmod, line 76: should those be domain or hostname in case of IP addresses?

smalluban: I'll add fallback in the `domain` property of the `ExtendedRequest`

chrmod (approved):

PR #1583, opened 2024-05-09
DNR feedback on Safari

smalluban:
From my testing, it looks like matched rules are only returned for blocked requests (rules can also match and allow requests). I assumed that this was right, so every matched rule is treated as if it was blocked.

* Refactor of the reporting file structure - not required, but done while developing - it would be a waste to undo it
* Adds check of `getMatchedRules` after stats are updated in Safari


smalluban:
@chrmod Please reset to the new version of the branch if you want to work on this branch. I made a rebase to the current `main`.

extension-manifest-v3/src/background/stats.js
```diff
@@ -186,12 +186,40 @@ export function updateTabStats(tabId, requests) {
...
+          const { rulesMatchedInfo } =
+            await chrome.declarativeNetRequest.getMatchedRules({
+              tabId,
+              minTimeStamp: stats.timestamp,
+            });
+
+          for (const info of rulesMatchedInfo) {
+            for (const tracker of stats.trackers) {
+              for (const request of tracker.requests) {
+                if (request.url === info.request.url) {
```
chrmod, line 204: this is great

please make it even better and add update stats for all urls that haven't matched already observed urls

chrmod (changes requested):

philipp-classen (approved):
Tested the refactoring with Chrome (reporting still works).

philipp-classen:
To fix the merge conflict: this is what has been added
https://github.com/ghostery/ghostery-extension/commit/9e9f9620de3ab1483aa7f09e5f8efb8768b81c86

smalluban:
@philipp-classen I Added config for Edge, please scan if it is right now - https://github.com/ghostery/ghostery-extension/pull/1583/files#diff-0077760dac31d77155faff549d98ea292441ec2f4f177fdd0ceee1d7620e4899R72

philipp-classen (dismissed):

philipp-classen:
> @philipp-classen I Added config for Edge, please scan if it is right now - https://github.com/ghostery/ghostery-extension/pull/1583/files#diff-0077760dac31d77155faff549d98ea292441ec2f4f177fdd0ceee1d7620e4899R72

Looks good!

smalluban:
@chrmod I think I made it as close to a 100% match as possible. The `getMatchedRules` must take a timestamp, as without it, it will return blocked requests for the previous page in the same tab. 

Safari has a known bug, that if you pass the URL directly in the new tab (add tab and write URL) - then the `onBeforeNavigate` event is not triggered. Then, we use a message from the content script to set up stats. In this case and only, the timestamp will be "too late" - so some of the DNR feedback might be cut off.

@chrmod Please re-review.

chrmod (approved):

PR #1611, opened 2024-05-24
Pausing happens on a hostname level

chrmod:
fix https://github.com/ghostery/broken-page-reports/issues/660

Current stats logic works on the domain level, so for url https://ads.google.com, the stats are saved on domain `google.com`.
Screenshot 2024-05-24 at 11 28 14

This does not work well with the pausing logic which rely on stats domain name. When pausing at ads.google.com, the exception is added for `google.com` and not for the subdomain `ads.google.com`


smalluban:
Closing in favor of #1615

PR #1615, opened 2024-05-28
Refactor domain stats to hostname

smalluban:
* Refactor `domain` in stats to use `hostname`
* To support already added pages to paused/exceptions, the hostname is clean up from the `www` subdomain - which is the most common case. Without it already added pages won't be selected properly

Minor fixes:
* `fixes` DNR rules must be added only when terms are accepted

The `tracker-exception.js` code is left as it is in purpose - renaming fields would require data migration. Also, this is likely that this part will be touched with refactor selective blocking.

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -24,7 +24,7 @@ import asyncSetup from './utils/setup.js';
 import { updateTabStats } from './stats.js';
 
 let enabledEngines = [];
-let pausedDomains = [];
+let pausedHostnames = [];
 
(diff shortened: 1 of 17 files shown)


chrmod (commented):
Looks good code wise.

@GRadziejewski we definitely have to test it before merging.

smalluban:
@chrmod @GRadziejewski My tests passed ;) It looks that it works fine with the refactor.

GRadziejewski (approved):
Test results: https://github.com/ghostery/ghostery-extension/issues/1616#issuecomment-2137321254.
LGTM

PR #1628, opened 2024-06-04
Clean up devtools mode

### LINKED ISSUE #1625: DNR lists are still set to work after disabling all 3 features (Ad-Blocking, Anti-Tracking, Never-Consent)
**Description**
DNR lists are still set to works, even if all 3 features (Ad-Blocking, Anti-Tracking, Never-Consent) are disabled. 

**Steps to Reproduce**

1. Open Ghostery settings, click 3 times on the version number to open dev tools
2. Disable all 3 features (Ad-Blocking, Anti-Tracking, Never-Consent)
3. Open any other tab, for example Websites, and go back to Privacy protection tab
4. Open dev tools once more
5. Look at DNR rulesets
image

6. Open any website, for example https://aarp.org and open Ghostery panel
image

**Expected results:**
When all 3 features (Ad-Blocking, Anti-Tracking, Never-Consent) are disabled, it should work just like Pause. Nothing should be blocked.

**Versions:**
```
Browser: Safari 17.5
OS: MacOS 14.5
GBE 10.3.4

Browser: Safari
OS: iOS 17.5.1
GBE 10.3.4

Browser: Edge 125.0.2535.79
OS: Windows 10
GBE 10.3.4
```

smalluban:
* Events are actions, not state names - for example, we have "click" event, not "is-clicked", so we should use the name in the sense of what happens, not what is a current state - "shown", not "is-shown"
* Visibility is a bool value from its name - a thing can be visible or not, so the "is" prefix is not needed, it is only helpful with functions, which usually don't have obvious returning value - "is" helps then, that a function return a bool value
* You can pass the `devMode` to the component, so after refresh the dev mode component is visible
* Fixes the issue of keeping DNR lists when all of the options are disabled (as the PR touches that part)

Fixes #1625

### DIFF
--- a/extension-manifest-v3/src/pages/settings/views/privacy.js
+++ b/extension-manifest-v3/src/pages/settings/views/privacy.js
@@ -46,10 +46,6 @@ function toggleNeverConsent({ options }) {
   });
 }
 
-function onDevModeEnabled(host) {
-  host.devMode = true;
-}
-
 export default {
   options: store(Options),
   session: store(Session),
@@ -180,7 +176,8 @@ export default {
           </section>
 
           <gh-settings-devtools
-            onis-shown=${onDevModeEnabled}
+            onshown="${html.set('devMode', true)}"
+            visible="${devMode}"
           ></gh-settings-devtools>
         `}
         ${store.ready(session) &&
--- a/extension-manifest-v3/src/pages/settings/views/privacy.js
+++ b/extension-manifest-v3/src/pages/settings/views/privacy.js
@@ -83,7 +83,7 @@ export default {
                   </ui-text>
                 </div>
                 <ui-toggle
-                  disabled=${!devMode}
+                  disabled=${options.blockAds && !devMode}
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/src/pages/settings/views/privacy.js
```diff
@@ -180,7 +176,8 @@ export default {
           </section>
 
           <gh-settings-devtools
-            onis-shown=${onDevModeEnabled}
+            onshown="${html.set('devMode', true)}"
```
smalluban, line 179: For that one-liners you can use `html.set` helper with custom value as a second argument. No need for custom callback function here.

chrmod (approved):

smalluban:
@chrmod I assume that you want @GRadziejewski to test it before merging?

smalluban:
I think this is not worth to test it separately, as we have upcoming release. I will merge it, and we can test it out within tests of the 10.3.5.

PR #1647, opened 2024-06-10
Prevent double injection of scriptlets

chrmod:
Includes https://github.com/ghostery/adblocker/pull/4016

### DIFF
--- a/extension-manifest-v3/src/background/debug.js
+++ b/extension-manifest-v3/src/background/debug.js
@@ -1,0 +1,11 @@
+/**
+ * Ghostery Browser Extension
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+globalThis.ghostery = {};
--- a/extension-manifest-v3/src/background/reporting/index.js
+++ b/extension-manifest-v3/src/background/reporting/index.js
@@ -115,8 +115,6 @@ if (__PLATFORM__ !== 'safari') {
   chrome.webNavigation.onHistoryStateUpdated.addListener(onLocationChange);
 }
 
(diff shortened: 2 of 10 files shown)


extension-manifest-v3/src/background/reporting/index.js
```diff
@@ -115,8 +115,6 @@ if (__PLATFORM__ !== 'safari') {
   chrome.webNavigation.onHistoryStateUpdated.addListener(onLocationChange);
 }
 
-// for debugging service-workers
-globalThis.ghostery = globalThis.ghostery || {};
 globalThis.ghostery.WTM = {
```
smalluban, line 118:
```suggestion
import debug from '/utils/debug.js`

debug.WTM = {
```

extension-manifest-v3/src/utils/engines.js
```diff
@@ -428,3 +428,5 @@ chrome.alarms.onAlarm.addListener((alarm) => {
     });
   }
 });
+
+globalThis.ghostery.engines = { get };
```
smalluban, line 432:
```suggestion
import debug from '/utils/debug.js'`;

debug.engines = { get };
```

extension-manifest-v3/src/background/debug.js
```diff
@@ -0,0 +1,11 @@
...
+ * Ghostery Browser Extension
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+globalThis.ghostery = {};
```
smalluban, line 11:
```suggestion
const debug = {};
globalThis.ghostery = debug;

export default debug;
```

Move file to `/utils/debug.js`

smalluban (changes requested):
Introducing `debug.js` you added  an important side-effect - as other files relay that `globalThis.ghostery` exists. If we need that global object it is much better to introduce a separate module `debug.js` which provides that global object.

smalluban (approved):

PR #1657, opened 2024-06-12
Add Block-All mode

smalluban:
Adds `Block-All` mode toggle to the trackers page. 

Switching it on:
* disables TrackerDB DNR ruleset
* Forces `isCategoryBlockedByDefault` to return true

It means that TrackerDB is removed from allowing/blocking anything by default (if there is no exception on the tracker).

The copy needs a review, it is made by me ;)

Zrzut ekranu 2024-06-12 o 15 16 14


GRadziejewski:
Checked and, in my opinion, it works good. No breakage spotted (checked with a few [commonly visited pages](https://hackmd.io/rtCQmOgDTySbIjSJUPbAHQ?view).

PR #1659, opened 2024-06-12
Update block by default & trackers list UI

smalluban:


extension-manifest-v3/src/pages/settings/views/trackers.js
```diff
@@ -65,8 +65,15 @@ function clearCategory(id) {
             store.ready(t.exception) &&
             t.exception.blocked !== category.blockedByDefault,
         )
-        .map((tracker) =>
-          store.set(tracker.exception, { blocked: category.blockedByDefault }),
+        .map(({ exception }) =>
+          store.set(
```
smalluban, line 69: Much better, as when there is no exceptions, the whole model is removed, so in the case of being blocked by default, it does not introduce harmful filters.

extension-manifest-v3/src/store/options.js
```diff
@@ -12,7 +12,7 @@
 import { store } from 'hybrids';
 import { deleteDB } from 'idb';
 
-import { getUserOptions, setUserOptions } from '/utils/api.js';
+import { getUserOptions, setUserOptions } from '../utils/api.js';
```
smalluban, line 15: Required for tests to pass.

chrmod (approved):

PR #1660, opened 2024-06-13
Minor fixes to release script

smalluban:
* Renames branch name to `release`
* Fixes locales command - there is a new `src/messages.json` file, which contains phrases, which are not set in the source code
* Adds build number to commit message

### DIFF
--- a/extension-manifest-v3/scripts/release.sh
+++ b/extension-manifest-v3/scripts/release.sh
@@ -5,8 +5,8 @@
 # Create release branch from latest main
 git checkout main
 git pull
-git checkout -b mv3-release
-git checkout mv3-release
+git checkout -b release
+git checkout release
 
 # Remove local node_modules if exists
 rm -rf ./node_modules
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/scripts/release.sh
```diff
@@ -5,8 +5,8 @@
 # Create release branch from latest main
 git checkout main
 git pull
-git checkout -b mv3-release
-git checkout mv3-release
+git checkout -b release
+git checkout release
```
chrmod, line 9: can we change this to a version number? eg. `10.3.8` while the tag will be `v10.3.8`

chrmod (changes requested):

extension-manifest-v3/scripts/release.sh
smalluban: It's like to have information before we now it. The version is bumped by the `npm version`, but it must be run after we create branch. 

I am simplifying it as soon we will have only v10. We also should never have two open release branches, it then must be something wrong. 

For v8, as I remember, you where creating a branch name as a version already. However, I dislike this pattern too, as it looks like a tag, which is not a tag (tag is fixed commit, branch is a head which moves - there is a difference).

chrmod (approved):

PR #1662, opened 2024-06-13
fix: use Set over the Array instance for paused hostnames

### LINKED ISSUE #1651: Avoid unnecessary overhead if lots of sites are being paused
Currently, we keep it in an array data structure. By switching to a set, it would provide better protection against situations where lots of sites are paused and linear scans add up.

smalluban:
Fixes #1651

Still, I am not sure if this is better. The real question is if `Array.prototype.includes` is much slower than `Set.prototype.has`... The list has unique IDs (so no benefit here), and it must be cleared each time, as otherwise, we would need to check if the user removed the domain (and it might be then even slower).

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -24,7 +24,7 @@ import asyncSetup from './utils/setup.js';
 import { updateTabStats } from './stats.js';
 
 let enabledEngines = [];
-let pausedHostnames = [];
+let pausedHostnames = new Set();
 
 const setup = asyncSetup([
   observe(null, (options) => {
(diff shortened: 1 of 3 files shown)


chrmod (approved):

philipp-classen:
The back-and-forth converting could be avoid, I assume, but modeling the underlying data structure:

```
paused = [{ id: "foo", ...rest1 }, { id: "bar", ...rest2 }, ...]
```

as

```
paused = {
  foo: { ...rest1 },
  bar: { ...rest2 },
  ...
}
```

But I'm also fine with keeping the array structure if it is difficult to change.

smalluban:
The dynamic keys are not supported in [Store](https://hybrids.js.org/#/store/model).

philipp-classen (approved):

PR #1663, opened 2024-06-13
Experimental filters

chrmod:
Allow users to enable experimental filters which may help with some websites like currently Youtube.com

Experimental filters, by definition are not well tested so may cause breakage, that's why they cannot be enabled by default. 

Screenshot 2024-06-14 at 20 02 22

### DIFF
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -29,7 +29,7 @@ export const SYNC_OPTIONS = [
   'trackerCount',
   'wtmSerpReport',
   'serpTrackingPrevention',
-  'experimentalCosmeticFilters',
+  'experimentalFilters',
   'panel',
 ];
 
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -29,6 +29,7 @@ export const SYNC_OPTIONS = [
   'trackerCount',
   'wtmSerpReport',
   'serpTrackingPrevention',
+  'experimentalCosmeticFilters',
   'panel',
 ];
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/src/store/options.js
```diff
@@ -29,6 +29,7 @@ export const SYNC_OPTIONS = [
   'trackerCount',
   'wtmSerpReport',
   'serpTrackingPrevention',
+  'experimentalCosmeticFilters',
```
smalluban, line 32: I would name this option just `experimental` - as this not mean only cosmetics or filters (at all for DNR version)

smalluban, line 32: I would name this option just `experimental` - as this not mean only cosmetics or filters (at all for DNR version)
You have even a bug here, as this key does not match the below one ;)

extension-manifest-v3/src/utils/engines.js
```diff
@@ -430,4 +432,11 @@ chrome.alarms.onAlarm.addListener((alarm) => {
   }
 });
 
+observe('experimentalFilters', (experimentalFilters) => {
```
smalluban, line 435: You are using this in line 42, but this observe is in 435. Please move it close to where the key is used.

smalluban (changes requested):

extension-manifest-v3/src/store/options.js
chrmod: Prefer to call it `experimentalFilters` as I expect we can have different experiments in near feature

smalluban (approved):

PR #1677, opened 2024-06-18
Force update engines when experimental filters have changed

smalluban:
There might be a case, where a user turns on the experimental filters, but they won't work as expected. This will happen if the user is in the 60-minute window between automatic updates of the engines.

To fix the issue we need to force the update.

### DIFF
--- a/extension-manifest-v3/src/utils/engines.js
+++ b/extension-manifest-v3/src/utils/engines.js
@@ -40,6 +40,10 @@ const ENV = new Map([
 
 observe('experimentalFilters', (experimentalFilters) => {
   ENV.set('env_experimental', experimentalFilters);
+
+  // As engines on the server might have new filters, we need to update them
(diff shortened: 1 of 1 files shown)


extension-manifest-v3/src/utils/engines.js
```diff
@@ -40,6 +40,10 @@ const ENV = new Map([
 
 observe('experimentalFilters', (experimentalFilters) => {
   ENV.set('env_experimental', experimentalFilters);
+
+  // As engines on the server might have new filters, we need to update them
+  updateAll().catch(() => null);
```
chrmod, line 45: should't this be awaited?

smalluban: Don't think so. The effect will be the same.

chrmod (approved):

Issue #1680, opened 2024-06-18
"Trusting" a page does not work on subdomains


philipp-classen:
Example: https://promo.pch.com/ (see https://github.com/ghostery/broken-page-reports/issues/694)

philipp-classen:
It turns out the ` initiatorDomains` is not enough to match the original request that loads the site:
https://github.com/ghostery/ghostery-extension/blob/f01ba094766bdd361a6fba9666d722eee348154f/extension-manifest-v3/src/background/dnr.js#L96

Potential fix: https://github.com/ghostery/ghostery-extension/pull/1681

PR #1681, opened 2024-06-18
If a page is trusted, unblock also the initial load

philipp-classen:
If a page is trusted, only the requests that originated from the domain were allowed, but not the initial request to load the site.

fixes https://github.com/ghostery/ghostery-extension/issues/1680
refs https://github.com/ghostery/broken-page-reports/issues/694


smalluban (approved):
LGTM. Only the `await` is not required, as we don't make any calls after updating rules, and we are not waiting outside of the function to finish.

smalluban:
One notice - the code will have an effect only when the user touches trusted (paused) sites. Only then the dynamic rules will be reloaded.

PR #1685, opened 2024-06-21
Move `offscreen` permission from optional to required

smalluban:
The `chrome.permissions.request({ permissions: ['offscreen'] })` throws in Kiwi Browser (Android, chromium):

Zrzut ekranu 2024-06-21 o 09 25 20

Even though this must be a browser bug, we can address the problem by moving optional `offscreen` permission to the required list.

I have tested the update process of the extension in Chrome, using the instructions from https://developer.chrome.com/docs/extensions/develop/concepts/permission-warnings#update-tool

To be sure that the process works, I also tested adding `geolocation` permission (in the same way as the `offscreen`). For the `geolocation` I have seen the prompt, but for `offscreen` I have not.

Zrzut ekranu 2024-06-21 o 10 41 57

It means that adding `offscreen` required permission is safe, and should not trigger a prompt about disabled extension.


chrmod (changes requested):
@smalluban can we try to release it into chrome web store beta channel to ensure it is safe? Please coordinate with @GRadziejewski

smalluban:
So, please approve, we will prepare the next release, and release it first to the beta channel. I am sure, that we can't do more than I already did. @GRadziejewski can do the same on his side to be 100% sure (I mean to use that testing tool from Chrome).

How that rejected review is related to the changes in the code?

chrmod (approved):

Issue #1690, opened 2024-06-24
scriptlet are injected in iframes of "trusted" pages


philipp-classen:
Examples:
* https://github.com/ghostery/broken-page-reports/issues/697
* https://github.com/ghostery/broken-page-reports/issues/700

If the page is trusted, it will stop injecting for the main frame, but not subframes. This is the listener that triggers:
https://github.com/ghostery/ghostery-extension/blob/12cbbecccd09ab86affbff566da0408e84b3f747/extension-manifest-v3/src/background/adblocker.js#L328

The `details` object can look like that (triggered from the main frame `https://www.thesignatureedge.com/gallery):
```
{
  "documentId": "9918491B2AE2AA6D616759A6CC33DBA9",
  "documentLifecycle": "active",
  "frameId": 211,
  "frameType": "sub_frame",
  "parentDocumentId": "777835FC67713375564AD5A967B50B12",
  "parentFrameId": 0,
  "processId": 91,
  "tabId": 589212987,
  "timeStamp": 1719234441683.585,
  "transitionQualifiers": [],
  "transitionType": "auto_subframe",
[...]

PR #1691, opened 2024-06-24
Skip injections for subframes of trusted pages

philipp-classen:
Heuristic to detect navigations of subframes of trusted pages and skip them.

refs https://github.com/ghostery/ghostery-extension/issues/1690
refs https://github.com/ghostery/broken-page-reports/issues/697

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -56,6 +56,19 @@ function isPaused(hostname) {
   );
 }
 
+// This function is only an approximation: the async chrome.tabs introduces a race
+// (maybe it could be mitigated by extra bookkeeping of subframes with a pages).
+// Note: using this function makes only sense for subframes. Otherwise, you should
+// have synchronous access to URL. In that case, you should use "isPaused".
+async function isPausedTab(tabId) {
+  const tab = await chrome.tabs.get(tabId);
+  if (!tab?.url) {
+    return false;
+  }
+  const { hostname } = parse(tab.url);
+  return isPaused(hostname);
+}
+
 function adblockerInjectStylesWebExtension(
   styles,
   { tabId, frameId, allFrames = false },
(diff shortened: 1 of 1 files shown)


extension-manifest-v3/src/background/adblocker.js
```diff
@@ -326,7 +339,10 @@ if (__PLATFORM__ === 'safari') {
   });
 } else {
   chrome.webNavigation.onCommitted.addListener(async (details) => {
-    injectScriptlets(details.tabId, details.url);
+    const isMainFrame = details.parentFrameId === -1;
+    if (isMainFrame || !(await isPausedTab(details.tabId))) {
```
smalluban, line 343: The `details` object contains the `url`. Why do you need to use `chrome.tabs` API? We can simply call `isPaused(parse(details.url).hostname)`

smalluban (changes requested):

extension-manifest-v3/src/background/adblocker.js
smalluban: Reference: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/onCommitted#url

smalluban, line 343: I have an idea to use `tabStats` here synchroniusly. As the mainframe must call this event before subframes, we can find stats for the `details.tabId` and have the `stats.hostname` here.

smalluban:
```suggestion
    if (isMainFrame || !(await isPaused(tabStats.get(details.tabId)?.hostname) {
```

smalluban (changes requested):

extension-manifest-v3/src/background/adblocker.js
philipp-classen: This is a follow-up that tries out the idea: https://github.com/ghostery/ghostery-extension/pull/1692

philipp-classen:
Closed in favor of https://github.com/ghostery/ghostery-extension/pull/1692

PR #1692, opened 2024-06-24
Detect navigations of subframes of trusted pages and skip them.

philipp-classen:
Follow-up of https://github.com/ghostery/ghostery-extension/pull/1691 avoiding the asynchronous call to the tab API.

refs #1690
refs ghostery/broken-page-reports#697
refs https://github.com/ghostery/broken-page-reports/issues/700

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -21,7 +21,7 @@ import * as engines from '/utils/engines.js';
 import Request from './utils/request.js';
 import asyncSetup from './utils/setup.js';
 
-import { updateTabStats } from './stats.js';
+import { tabStats, updateTabStats } from './stats.js';
 
 let enabledEngines = [];
 let pausedHostnames = new Set();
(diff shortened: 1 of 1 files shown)


smalluban (approved):

chrmod (commented):
What about Safari? that platform uses a different injection path due to not reliable `onCommited` events.

smalluban (changes requested):
Good point from @chrmod. Please move the condition to `injectScriptlets()` function, so all the platforms will have that.

philipp-classen:
I tried now with Safari (MacOS), but could not reproduce the bug that this pull requests is intended to cover. It is also not intuitive to me why we need to change the Safari path, since I see it as a problem of the Non-Safari path only:

1. Non-Safari platforms use onCommitted to inject. onCommitted triggers for frames and subframes. This is why pausing did not cover subframes (inteded to be fixed by this pull request)
2. Safari: it does not use onCommitted, but injects a script via "content_scripts/adblocker-scriptlets.js" (https://github.com/ghostery/ghostery-extension/blob/1e74ce98523e4dba49ce031feb728a3eb1b8195e/extension-manifest-v3/src/content_scripts/adblocker-scriptlets.js). This sends back one message (with the original page). The existing pause mechanism in here https://github.com/ghostery/ghostery-extension/blob/1e74ce98523e4dba49ce031feb728a3eb1b8195e/extension-manifest-v3/src/background/adblocker.js#L279 should handle it correctly.

smalluban:
Yes, Safari uses a different approach, but still, the code uses `sender.url` which is the subframe URL, not the tab URL. 

We want to have unified logic as much as possible. Today it might not be an issue for Safari, but tomorrow it might. Why not allowing subframes is fine on Safari, but it doesn't on Chromium?

philipp-classen:
A follow-up that handles scenarios in Safari when the sender URL differs from the origin, is here: https://github.com/ghostery/ghostery-extension/pull/1695

I do not have an example it can happen. In the actual examples that break Chrome, the extra path is not being reached, but it also does not break something.

smalluban:
Fixed in https://github.com/ghostery/ghostery-extension/pull/1711

PR #1697, opened 2024-06-26
Release v10.3.11-178

smalluban:
c3a4bcdb77e420fc61f376f5fd3f1b7de8e966a2 Fix sending image src from setter (#1696)
8055753e561ffab34a2626152ee17928141a9e57 Move `offscreen` permission from optional to required (#1685)

Testing focus:

1. #1696 - only affects Safari
2. #1685 - only affects Chromium browsers

After usual testing, we will push it first to the beta channel for Chrome, to ensure there is no problem with changed permissions.


GRadziejewski (approved):
Tests results: https://github.com/ghostery/ghostery-extension/issues/1698#issuecomment-2191213086

PR #1704, opened 2024-07-01
Never-Consent: avoid scriptlet conflicts

chrmod:
There are adblocker filters that provide functionality similar to Never-Consent. This change, prevents scriptlets execution on domains belonging to Consent Managers.

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -274,6 +274,25 @@ ${scripts.join('\n\n')}}
   );
 }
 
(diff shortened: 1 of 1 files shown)


smalluban:
@chrmod Do you have a use case where I can test out the issue? Do those CMP lists from TrackerDB match all of the autocosent coverage? I am not sure also how it helps, as it will only apply to CMPs which opens up the iframe? Is this a common practice? What about CMPs loaded in the main frame as a script?

chrmod:
closing until proven useful

Issue #1718, opened 2024-07-04
Regional filters


chrmod:
Ghostery 8 used to auto detect users region and apply additional block lists. In Ghostery 10 we can do the same until https://github.com/ghostery/journal/issues/366 is implemented. 

The lists are already prepared by the adblocker-filters projects. Question is how they should be integrated with the extension. DNR is pretty clear as we can simply toggle a ruleset. The adblocker engines are not so obvious as we likely don't want to have multiple additional engines due to aspects as exceptions sharing, custom filter application. On the other hand it is not clear how to provide updates if we continue with Ghostery 8 strategy (to download, parse and updateDiffs). 

This needs investigation.

PR #1738, opened 2024-07-12
Regional filters

smalluban:
Fixes [#1718](https://github.com/ghostery/ghostery-extension/issues/1718)

### Overview

Adds regional engine and DNRs. The implementation differs according to the target platform:
* Firefox - full engines with network and cosmetics
* Chromium - DNR lists and engine with cosmetics
* Safari - only cosmetics - NO NETWORK FILTERS (this is caused by the strong limit of 75k rules, which we already use)

The feature will be turned on by default with a pre-selected list of regions detected by the accepted languages (`navigator.languages`).

### Testing 

To test if the lists work, the best way is to check one of the cosmetic/network rules from regional lists, like https://easylist-downloads.adblockplus.org/easylistpolish.txt for polish. For example `cowwilanowie.pl##.a-re` hides the top section with ads.

## UI

@mdsandu Please take a look at the new section on the settings page.

Zrzut ekranu 2024-08-7 o 09 15 11


extension-manifest-v3/src/background/index.js
```diff
@@ -22,9 +22,13 @@ import './serp.js';
 import './helpers.js';
 import './external.js';
 
+import './filters/custom-filters.js';
```
chrmod, line 25: how about we call the folder `adblocker` instead?

smalluban: I think it will be gone, as it might be not needed.

smalluban: This is WIP

mdsandu:
update to:
Regional Filters
Blocks additional ads, trackers, and pop-ups specific to your region. Enable only the regions you need to avoid slowing down your browser.

smalluban:
@GRadziejewski I've found a minor bug with the checkbox component - it was still possible to change its value when Ghostery was globally disabled.  There is no feature change.

GRadziejewski:
Test results: https://github.com/ghostery/ghostery-extension/issues/1791#issuecomment-2275578137

GRadziejewski (approved):

extension-manifest-v3/src/utils/engines.js
```diff
@@ -342,6 +387,11 @@ async function update(name) {
 
     if (updated) {
       console.log(`Engine "${name}" updated`);
+
+      // Notify listeners
+      const fns = updateListeners.get(name);
+      fns?.forEach((fn) => fn());
```
chrmod, line 393: listener callbacks should be guarded by try/catch - in case any of them throw, the saveToStorage should still be called

chrmod (approved):

PR #1743, opened 2024-07-16
Enable dependabot

chrmod:


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: npm
```
smalluban, line 8: Can we scope this to only a limited number of packages? If not, I think we can be flooded with updates for v8.. but I am not sure if this is going to work, as I expect that it will only look at the root dependencies. Does it support mono-repos and nested package.json files?

smalluban (changes requested):
Not sure if the config is done right.

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
smalluban, line 1: Remove the comment. I suppose it has sense only in the template, not when used.

.github/dependabot.yml
chrmod: Good questions.

I've took it from https://github.com/ghostery/adblocker/blob/master/.github/dependabot.yml
so it does support monorepos. 

Also as far I can can tell it cover root dependencies only.

chrmod:
@smalluban in case v8 would start to cause problems we can always tweak the dependabot behaviour.
It offers those commands on github comments 
Screenshot 2024-07-16 at 10 27 39

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/"
```
smalluban, line 4: Is this right? Does it apply to all nested packages?

smalluban (approved):
Still I think there will be a lot of PRs from it if we allow to update everything.

chrmod:
it's getting tricky. our workspace setup requires us to run `npm dedupe` when updating shared dependencies like `adblocker` or `autoconsent`. Lets wait with introduction of dependabot until we sunset Ghostery 8 later this summer.

PR #1744, opened 2024-07-16
Update dependencies

chrmod:


extension-manifest-v3/package.json
```diff
@@ -28,8 +28,8 @@
   "devDependencies": {
     "shelljs": "^0.8.5",
     "sinon-chrome": "^3.0.1",
-    "vite": "^4.5.2",
-    "web-ext": "^7.11.0",
+    "vite": "4.5.2",
```
smalluban, line 31: Don't fix the package version, the lock file is for that. Here `^` should stay. 
Another thing, if you want to update `vite`, there is a major bump to `v5`.

smalluban (changes requested):

extension-manifest-v3/package.json
chrmod: we have to fix this package as we have a patch for it ;)

smalluban: Still the `patch-package` solves that very well, as if the version is bumped, it tries to apply the patch, and if fails, it throws an error, so if you fix the patch version (the `^` fixes the major), we have to manually update. 

I suppose that you want to avoid problems when using the general `npm update` command. The change is not needed, but if you like it that why, it's fine for me.

smalluban (approved):

extension-manifest-v3/package.json
chrmod: I've made a change as it broke on me with latest update

PR #1746, opened 2024-07-17
Fix exceptions DNR rules update

smalluban:
Trusting the website was removinfg the exception DNR rules, as there was not filter on the ids to be updated (before the exceptions it was just fine).

Reproduce bug path:

1. Open any page with a tracker from essential category and add global exception to block that tracker
2. Trust and untrust the page
3. Look at the tracker urls - they are not blocked (but they should)

The bug is about clearing rules, which blocked the tracker when exception was added.

The PR fixes the issue by:
1. Filter the ids for general DNR rulesets
2. Add `addUpdateListener` pattern to watch when TrackerDB changes, so then we must update exceptions (the filters might have changed)

The fix also will fix current broken state of the users by the 2nd point - as those users with less than hour (1 hour is a scheduled update of engines) the filters will be reloaded.

### DIFF
--- a/extension-manifest-v3/src/background/dnr.js
+++ b/extension-manifest-v3/src/background/dnr.js
@@ -12,26 +12,26 @@
 import { observe, ENGINES, isPaused } from '/store/options.js';
 import { TRACKERDB_ENGINE } from '/utils/engines.js';
 
-const PAUSE_RULE_PRIORITY = 10000000;
-
-const ALL_RESOURCE_TYPES = [
-  'main_frame',
-  'sub_frame',
-  'stylesheet',
-  'script',
-  'image',
-  'font',
-  'object',
-  'xmlhttprequest',
-  'ping',
-  'media',
-  'websocket',
-  'webtransport',
-  'webbundle',
-  'other',
-];
-
 if (__PLATFORM__ !== 'firefox') {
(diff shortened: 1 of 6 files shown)


extension-manifest-v3/src/background/dnr.js
```diff
@@ -74,21 +75,30 @@ if (__PLATFORM__ !== 'firefox') {
...
+    // The background process starts and runs for each tab, so we can assume
+    // that this function is called before the user can change the paused state
+    // in the panel or the settings page.
     if (!prevPaused) return;
 
-    const dynamicRules = await chrome.declarativeNetRequest.getDynamicRules();
     const hostnames = Object.keys(paused);
 
     if (hostnames.length) {
+      const removeRuleIds = await getPauseRulesIds();
```
chrmod, line 100: can we move it above the if?

chrmod (approved):

extension-manifest-v3/src/background/dnr.js
smalluban: It was here when I though that we must call it always (not check `prevPaused`). However, it is called for both cases, so for now it's pointless. Good catch.

chrmod (approved):

PR #1754, opened 2024-07-18
Fix several exceptions bugs (adblocker, trackerdb, exceptions)

### LINKED ISSUE #1748: URLs are blocked even if the tracker is trusted
**Description**
After blocking a trusted tracker and setting it back as trusted, URLs are still blocked.

**Steps to Reproduce**
1. Open any page, for example https://www.fandom.com/
2. Look for any tracker from an essential category, for example Google Tag Manager
3. Add a global exception to block that tracker
4. Reload the website and trust that tracker
5. Look at the tracker URLs - they are blocked (but they should not)

image

**Expected results**
When the tracker is trusted, URLs must be observed.
When the tracker is blocked, URLs must be blocked.

**Versions**
```
Browser: Chrome 126.0.6478.127
OS: Windows 10
GBE 10.3.14

Browser: Firefox 128
OS: macOS 14.5
GBE 10.3.14

Browser: Safari 17.5
OS: macOS 14.5
GBE 10.3.14

Browser: Firefox 127.0.2
OS: Android 14
GBE 10.3.14

Browser: Opera 112.0.5197.25
OS: Windows 10
GBE 10.3.14

Browser: Safari
OS: iOS 17.5.1
GBE 10.3.14
```

smalluban:
Fixes #1748

Fixes several exceptions bugs:
* Clear `trackerdb.js` logic to be independent from other modules. The exceptions part (isTrusted) is moved to `adblocker.js` code
* Unify access to exceptions in background process to one synchronius storage - this is temporary (as I want to move out the exceptions storage to options in long term solution), but required to avoid loop of updates with using TrackerException store model eariler
* Request metadata cannot be memoized, as in Firefox they are accessed before the request is marked as modified or blocked . I checked and in the blocking process the metadata is accessed only once - onBeforeRequest uses it for not main frame request, and onHeadersReceived once for main frame, so memoizing is useless after all. The second place which uses it is stats.js , but this is async process, which does not influence the performance of the requests (just for statistics in the panel).
* Fixed adjusted flag in Trackers page to display trackers, which have added domains, but the global state is default


philipp-classen (approved):

extension-manifest-v3/src/utils/trackerdb.js
```diff
@@ -36,84 +34,48 @@ const categoryOrder = [
...
-    ...tracker,
-    isFilterMatched,
-    isTrusted: store.ready(exception)
-      ? isTrusted(request.sourceHostname, tracker.category, exception)
-      : !isCategoryBlockedByDefault(tracker.category),
-  };
-
-  return metadata;
+  // Get tracker info from the first match of the TrackerDB
+  return getTrackers().get(matches[0].pattern.key);
```
chrmod, line 78: why not `getTracker(matches[0].pattern.key)` ?

chrmod (approved):

chrmod (approved):

extension-manifest-v3/src/utils/trackerdb.js
smalluban: It must be sync, as `getTracker` is async function

GRadziejewski:
@smalluban LGTM

PR #1758, opened 2024-07-22
Disable tracker wheel

chrmod:
Reverts https://github.com/ghostery/ghostery-extension/pull/1733

In all browsers, but Firefox, the Tracker Wheel should be disabled by default.


GRadziejewski:
After a new installation, the default state on Chrome, Opera and Edge:
image
On Firefox, it is the opposite - it's turned ON.
Checked. 

LGTM

GRadziejewski (approved):

Issue #1773, opened 2024-08-01
You must double-trust for all sites to change the protection status to 'Trusted at all sites'


GRadziejewski:
**Description**
You need to double trusted for all websites, to change protection status to "Trusted on all websites".

**Steps to Reproduce**
1. Open https://www.aarp.org/
2. Open Ghostery panel on the Detailed view
3. Open details for any tracker, for example "DoubleClick" from the Advertising category.
4. Click on the "Blocked on all websites" button
![image](https://github.com/user-attachments/assets/e5b53837-b3e4-4cae-bd47-fb0cf95adc39)

5. Click on "Trust on all websites" -> Protection status changed to "Trusted on all websites"
6. Click on "Block on this website" -> Protection status changed to "Blocked on this website"
7. Click on "Trust on all websites" -> Protection status changed to "Blocked on all websites"
8. Click on "Trust on all websites" -> Protection status do not change to "Trusted on all websites"
9. Click again on "Trust on all websites" -> Protection status changed to "Trusted on all websites"

https://github.com/user-attachments/assets/c7433ace-b214-4d96-b6ec-5dead05a4064

**Expected results**
When you click "Trust on all websites" the protection status should change to "Trusted on all websites".

**Versions**
```
Browser: Chrome 127.0.6533.89
OS: Windows 10
GBE 10.3.16

Browser: Opera 112.0.5197.39
OS: Windows 10
GBE 10.3.16

Browser: Safari 17.6 (19618.3.11.11.5)
OS: macOS 14.6 (23G8Q)
GBE 10.3.16

Browser: Firefox 128.0.3
OS: Windows 11
GBE 10.3.16
```

PR #1774, opened 2024-08-01
Avoid inconsistent "double-trust" state in the detailed protection view

philipp-classen:
Background: see https://github.com/ghostery/ghostery-extension/issues/1773, where the UI can get in an inconsistent state that requires another action get out of again.

For each tracker and each domain, there are two states:
* the tracker is blocked/trusted on the domain ("this website")
* the tracker is globally blocked/trusted ("all websites")

where "blocked/trust" is being defined as being the opposite of the default that would otherwise come into effect. This change simplifies the state transition by updating them individually. Interweaving them opens the question how the transition in the 4 state graph should be and at the same time avoiding inconsistent states like before (and also being able to reach all states).
    
Keeping them separated, which also avoid the problems, looks like the most intuitive approach.

fixes https://github.com/ghostery/ghostery-extension/issues/1773

### DIFF
--- a/extension-manifest-v3/src/pages/panel/views/protection-status.js
+++ b/extension-manifest-v3/src/pages/panel/views/protection-status.js
@@ -5,7 +5,7 @@ import * as labels from '@ghostery/ui/labels';
 import TabStats from '/store/tab-stats.js';
 import { toggleExceptionDomain } from '/store/tracker-exception.js';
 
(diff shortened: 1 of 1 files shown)


smalluban:
Your change does not include removal of the exception, when it backs to the default state - #1776

philipp-classen:
Closed in favor of https://github.com/ghostery/ghostery-extension/pull/1776

PR #1786, opened 2024-08-06
Support trusted types

chrmod:
Fix for the Chrome permission on pages that use Trusted Types:
```
VM16:6 This document requires 'TrustedScript' assignment. This script element was modified without use of TrustedScript assignment.
VM16:6 This document requires 'TrustedScript' assignment. An HTMLScriptElement was directly modified and will not be executed.
VM16:14 This document requires 'TrustedScript' assignment.
VM16:14 
 Uncaught 
TypeError: Failed to set the 'textContent' property on 'Node': This document requires 'TrustedScript' assignment.
    at codeRunningInContentScript (<anonymous>:14:26)
    at <anonymous>:16:5
```

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -245,30 +245,23 @@ ${scripts.join('\n\n')}}
   // wrapper to break the "isolated world" so that the patching operates
   // on the website, not on the content script's isolated environment.
   function codeRunningInContentScript(code, nonce) {
-    let script;
-    try {
-      script = document.createElement('script');
-      let content = decodeURIComponent(code);
-      if (nonce) {
-        const trustedTypePolicy = window.trustedTypes.createPolicy(
-          `ghostery-${Math.round(Math.random() * 1000000)}`,
-          {
-            createScript: (string) =>
-              !string || string.slice(13, 13 + nonce.length) === nonce
-                ? string
-                : null,
-          },
-        );
-        content = trustedTypePolicy.createScript(content);
-      }
-      script.textContent = content;
-      (document.head || document.documentElement).appendChild(script);
-    } catch (ex) {
-      console.error('Failed to run script', ex);
+    const script = document.createElement('script');
+    let content = decodeURIComponent(code);
+    if (nonce) {
+      const trustedTypePolicy = window.trustedTypes.createPolicy(
+        `ghostery-${Math.round(Math.random() * 1000000)}`,
+        {
+          createScript: (string) =>
+            !string || string.slice(13, 13 + nonce.length) === nonce
+              ? string
+              : null,
+        },
+      );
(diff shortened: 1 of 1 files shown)


smalluban (dismissed):

philipp-classen (dismissed):

GRadziejewski:
Error:

```
Uncaught TypeError: Failed to execute 'createPolicy' on 'TrustedTypePolicyFactory': Policy with name "default" already exists.
    at codeRunningInContentScript (<anonymous>:3:27)
    at <anonymous>:22:5
```

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -223,15 +235,23 @@ async function executeScriptlets(tabId, scripts) {
...
 ${debugMarker('run scriptlets (executing in "page world")')}
 ${scripts.join('\n\n')}}
 )()`;
 
   // wrapper to break the "isolated world" so that the patching operates
   // on the website, not on the content script's isolated environment.
-  function codeRunningInContentScript(code) {
+  function codeRunningInContentScript(code, nonce, nonceSize) {
+    if (window.trustedTypes && window.trustedTypes.createPolicy) {
+      window.trustedTypes.createPolicy('default', {
```
smalluban: I think you must use returned policy to create a script, but here the returned policy is not set to any variable.

Look here https://developer.mozilla.org/en-US/docs/Web/API/TrustedScript

smalluban (changes requested):

smalluban (approved):

extension-manifest-v3/src/background/adblocker.js
smalluban: Ok, you made a fix in the same time I wrote this.

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -246,16 +246,22 @@ ${scripts.join('\n\n')}}
   // wrapper to break the "isolated world" so that the patching operates
   // on the website, not on the content script's isolated environment.
   function codeRunningInContentScript(code, nonce, nonceSize) {
+    let trustedTypePolicy;
     if (window.trustedTypes && window.trustedTypes.createPolicy) {
-      window.trustedTypes.createPolicy('default', {
+      trustedTypePolicy = window.trustedTypes.createPolicy(nonce, {
```
philipp-classen, line 251: I would avoid using the nonce here. If there is an API to list them, for instance. But I think you can use any name as long as it is likely to avoid collisions.

smalluban:
It looks that trustedTypes are supported from a along time now, I would avoid condition for each script injection. As this is only a Chrome thing, please use `__PLATFORM__` to remove condition in the Firefox and Safari builds.

philipp-classen (dismissed):

Issue #1800, opened 2024-08-09
Trusted/restricted websites are not migrated from v8 to v10


GRadziejewski:
**Description**
Websites added as trusted/restricted, after upgrade from v8 to v10, are not migrated

**Steps to Reproduce**
1. Install v8
2. Add a few sites as trusted and restricted
5. Update from v8 to v10
6. Open Ghostery Settings / Websites
7. Check if websites added at step 2 are visible

https://github.com/user-attachments/assets/f0379b61-7641-4991-918a-3ed087168328

**Expected results**
Trusted/proprietary websites must be migrated from v8 to v10.

Versions
```
GBE 10.3.17 https://github.com/ghostery/ghostery-extension/pull/1738#issuecomment-2275679665
```

GRadziejewski:
Fixed.
Checked here: https://github.com/ghostery/ghostery-extension/issues/1803

https://github.com/user-attachments/assets/b48fa857-79f0-4cc8-8252-b82fbd41791f

Issue #1799, opened 2024-08-09
After upgrading from v8 to v10 there is again Onboarding 


GRadziejewski:
**Description**
After upgrading from v8 to v10, I can see Onboarding again.

**Steps to Reproduce**
1. Install v8
2. Wait for the Onboarding
3. Enable Ghostery
4. Update from v8 to v10
5. Wait for Onboarding - there will be again

https://github.com/user-attachments/assets/b68b7940-f88d-48f2-a4c0-2c3afbde3f68

**Expected results**
If in v8 I did Onboarding, in v10 I shouldn't have to.

Versions
```
GBE 10.3.17 https://github.com/ghostery/ghostery-extension/pull/1738#issuecomment-2275679665
```

GRadziejewski:
Fixed.
Checked here: https://github.com/ghostery/ghostery-extension/issues/1803

https://github.com/user-attachments/assets/b48fa857-79f0-4cc8-8252-b82fbd41791f

PR #1801, opened 2024-08-12
Fix options migration from v8

### LINKED ISSUE #1797: Errors during migration from v8 to v10
**Description**
During updating from v8 to v10, there were plenty of errors.

**Steps to Reproduce**
1. Install v8
2. Open DevTools console for the extension
3. Update from v8 to v10
4. Check the logs
5. A lot of error:

> Adblocker: Error while processing network request - adblocker not ready yet
> TypeError: options.paused.reduce is not a function
> Error while observing options: TypeError: options.paused.reduce is not a function
> TypeError: options.paused is undefined

image
image

https://github.com/user-attachments/assets/2435ced6-eb46-4e23-ae66-7d9ea74d18ba

**Expected results**
There should be no errors when migrating from v8 to v10.

Versions
```
GBE 10.3.17 https://github.com/ghostery/ghostery-extension/pull/1738#issuecomment-2275679665
```

smalluban:
Fixes #1797 #1799 #1800

The core issue was to try to migrate options (in v10), which come from v8 (by not setting `optionsVersion`). As the options from v8 have the latest structure (the function was updated when the options structure has been changing), it should skip that process. 

@GRadziejewski Keep in mind, that when the "update" process is triggered by hand (not by the browser, you can still see errors like `Adblocker not ready yet` - as after updating the extension, the browser trigger immediately `onBeforeRequest` API, which cannot run "that fast", as for the first time, the engines are taken from the disk.

### DIFF
--- a/extension-manifest-v3/src/store/options.js
+++ b/extension-manifest-v3/src/store/options.js
@@ -84,20 +84,23 @@ const Options = {
 
   [store.connect]: {
     async get() {
-      let { options = {}, optionsVersion = 0 } = await chrome.storage.local.get(
-        ['options', 'optionsVersion'],
-      );
+      let { options, optionsVersion } = await chrome.storage.local.get([
+        'options',
+        'optionsVersion',
+      ]);
 
-      // Migrate options
-      if (optionsVersion < OPTIONS_VERSION) {
+      // Try to migrate options from v8 if options
+      // are not set (the initial get) for supported platforms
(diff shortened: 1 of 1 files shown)


chrmod (approved):

GRadziejewski (commented):
Tests results: https://github.com/ghostery/ghostery-extension/issues/1803#issuecomment-2283357952

PR #1809, opened 2024-08-13
Fix copy redirect resources only for supported platforms

smalluban:


chrmod:
@smalluban additional commit from @seia-soto is actually important to release to all platforms.

chrmod (approved):

PR #1813, opened 2024-08-13
Refactor custom filters

smalluban:
WIP

### DIFF
--- a/extension-manifest-v3/src/background/custom-filters.js
+++ b/extension-manifest-v3/src/background/custom-filters.js
@@ -8,9 +8,101 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
-import { parseFilters } from '@cliqz/adblocker';
+import { store } from 'hybrids';
+import {
+  parseFilters,
+  detectFilterType,
+  FilterType,
+  CosmeticFilter,
+} from '@cliqz/adblocker';
 
+import {
+  createDocumentConverter,
+  createOffscreenConverter,
+} from '/utils/dnr-converter.js';
 import * as engines from '/utils/engines.js';
+
+import Options, { observe } from '/store/options.js';
+import CustomFilters from '/store/custom-filters.js';
+
+const convert =
+  __PLATFORM__ !== 'safari' && __PLATFORM__ !== 'firefox'
+    ? createOffscreenConverter()
+    : createDocumentConverter();
+
+class TrustedScriptletError extends Error {}
+
+// returns a scriptlet with encoded arguments
+// returns undefined if not a scriptlet
+// throws if scriptlet cannot be trusted
+function fixScriptlet(filter, trustedScriptlets) {
+  const cosmeticFilter = CosmeticFilter.parse(filter);
+
+  if (
+    !cosmeticFilter ||
+    !cosmeticFilter.isScriptInject() ||
+    !cosmeticFilter.selector
+  ) {
+    return null;
+  }
+
+  const parsedScript = cosmeticFilter.parseScript();
+
+  if (!parsedScript || !parsedScript.name) {
+    return null;
+  }
+
+  if (
+    !trustedScriptlets &&
+    (parsedScript.name === 'rpnt' ||
+      parsedScript.name === 'replace-node-text' ||
+      parsedScript.name.startsWith('trusted-'))
+  ) {
+    throw new TrustedScriptletError();
+  }
+
+  const [front] = filter.split(`#+js(${parsedScript.name}`);
+  const args = parsedScript.args.map((arg) => encodeURIComponent(arg));
+  return `${front}#+js(${[parsedScript.name, ...args].join(', ')})`;
+}
+
+function parseInput(text = '', { trustedScriptlets }) {
+  return text
+    .split('\n')
+    .map((f) => f.trim())
+    .filter(Boolean)
+    .reduce(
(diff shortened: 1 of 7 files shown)


extension-manifest-v3/src/background/custom-filters.js
```diff
@@ -31,51 +123,88 @@ async function updateDNRRules(dnrRules) {
...
+  return {
+    networkFilters: networkFilters.length,
+    cosmeticFilters: cosmeticFilters.length,
+    errors: notSupportedFilters.map(
+      (msg) => `Filter not supported: '${msg.filter}'`,
+    ),
+  };
+}
+async function update(text, { trustedScriptlets }) {
+  const parseResult = parseInput(text, { trustedScriptlets });
```
smalluban, line 158: I tried to move the logic without changes, but I am not sure if this is still required - as for the engine, the input of the filters is parsed again, this time by the `parseFilters` function from the adblocker.

I left it here for now, as the internal parser uses the `trustedScriptlets` flag - but there was a bug before, which did not pass "fixed" scriptlets to the engine, but it generated them again.

extension-manifest-v3/src/store/options.js
```diff
@@ -303,19 +318,36 @@ async function migrateFromV8() {
   }
 }
 
+function isEqual(a, b) {
```
philipp-classen, line 321: Note that it is not symmetric:
```
> isEqual({}, { foo: 'bar' })
true
> isEqual({ foo: 'bar' }, {})
false
```

extension-manifest-v3/src/background/custom-filters.js
```diff
@@ -8,10 +8,102 @@
...
+  ) {
+    throw new TrustedScriptletError();
+  }
+
+  const [front] = filter.split(`#+js(${parsedScript.name}`);
+  const args = parsedScript.args.map((arg) => encodeURIComponent(arg));
+  return `${front}#+js(${[parsedScript.name, ...args].join(', ')})`;
+}
+
+function parseInput(text = '', { trustedScriptlets }) {
```
philipp-classen, line 69: IMHO, the old name "parseFilters" was more descriptive than "parseInput".

extension-manifest-v3/src/utils/dnr-converter.js
```diff
@@ -18,12 +18,16 @@ export function createDocumentConverter() {
 
     window.addEventListener('message', (event) => {
       const requestId = event.data.rules.shift().condition.urlFilter;
+      let { rules, errors } = event.data;
 
-      if (__PLATFORM__ === 'safari') {
-        event.data.rules = event.data.rules.map(getCompatRule).filter(Boolean);
-      }
+      requests.get(requestId)({
```
philipp-classen, line 23: Here I struggled first to understand what the effect is (same with the original version). Took me a bit to realize that "requests" is a mapping to callbacks, which should be triggered when the results of a conversion are available.

Maybe it would help the reader to name it to something like "requestResolvers"?

extension-manifest-v3/src/store/options.js
smalluban: Good catch, but it is safe here - as the structure of both compared objects is always identical - they are both two versions of options. I think I'll rename the function to `isEqualOptions` to avoid confusion.

extension-manifest-v3/src/background/custom-filters.js
smalluban: The function is not required anymore - the only part with fixing scriptlets will be moved, after discussion with @chrmod

extension-manifest-v3/src/utils/dnr-converter.js
smalluban: I'll look into it, but this part was at least touched by me in this PR ;)

smalluban:
@philipp-classen Updated and fixed what I could change. The `parseInput` must stay, but I renamed it to `normalizeFilters`, as it better describes what actually the function does - it does not produce filters, just split the input into buckets, and fixes scriptlets if needed.

extension-manifest-v3/src/background/custom-filters.js
```diff
@@ -6,7 +6,7 @@
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0
+w * file, You can obtain one at http://mozilla.org/MPL/2.0
```
philipp-classen, line 9: A "w" slipped in by accident.

philipp-classen (approved):

smalluban:
@GRadziejewski Testing highlights:
* The feature logic has not changed
* Added on/off toggle for the whole feature, so we can disable custom filters without removing the input
* Fixed issue with updated adblocker engine - to proof the bug update 3.18 -> 4.0, to proof the fix 3.18 -> 4.0 -> this PR

GRadziejewski:
Test results: https://github.com/ghostery/ghostery-extension/issues/1822#issuecomment-2288995414

Issue: https://github.com/ghostery/ghostery-extension/issues/1823

smalluban:
After testing the feature, I added a small change to not confuse users - the "Update" (actually "Save" now - it's more what it is, and it's already translated phrase) is disabled only when filters update.
Also, it fixes the state when the global pause is turned on.

GRadziejewski:
Test results: https://github.com/ghostery/ghostery-extension/issues/1822

LGTM.

GRadziejewski (approved):

PR #1815, opened 2024-08-13
Fix regional filters update condition

smalluban:
There is a race condition in the setup of the regional filters engine, specifically when the extension is updated (not the very first time loaded).

In that case (because we have updated the adblocker), the engines mismatch version, and must be fetched. Somehow, in that case, the regional filters engine returns `undefined`, which satisfies the condition to skip the update of the regional filters.

As a result, the regional filters engine (cosmetics) is not merged until the user touches the functionality. 

I think this bug does not affect all of the users, as this is a race condition (if the engine is set up before calling the `observe` method, then it is fine).

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -28,12 +28,15 @@ import { getException } from './exceptions.js';
 let enabledEngines = [];
 let options = {};
 
+const regionalFiltersEngine = engines.init(engines.REGIONAL_ENGINE);
+
 const setup = asyncSetup([
   // Init engines
   engines.init(engines.CUSTOM_ENGINE),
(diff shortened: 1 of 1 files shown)


extension-manifest-v3/src/background/adblocker.js
```diff
@@ -84,11 +84,9 @@ const setup = asyncSetup([
 
     // Clean previous regional engines
     if (lastValue) {
-      await Promise.all(
-        lastValue.regions
-          .filter((id) => !regions.includes(id))
-          .map((id) => engines.clean(`lang-${id}`)),
-      );
+      lastValue.regions
```
smalluban, line 87: It's leftover, as the `engines.clean()` is not async function (it was initially).

chrmod (approved):

PR #1824, opened 2024-08-20
feat: dark mode

### LINKED ISSUE #907: Ghostery panel dark mode

smalluban:
Fixes #907 

Please test it out, any suggestions are welcome.


smalluban:
I have found an issue with the preview trackers - did not check it before. Working on it.

GRadziejewski:
1. The difference and contrast between 
a) the red background of the notification and the grey background of the panel does not match in my opinion.
Screenshot_3
b) The same between the Trusted's yellow background and the panel's grey background.
Screenshot_4

2. The transition between the grey background and the white circle and hands is too sharp.
image

GRadziejewski:
**Description**
When switched to the landscape view and when the details of the tracker are visible, you can scroll the background list of trackers.

**Steps to Reproduce**
1. Open any website, for example https://www.aarp.org/
2. Open Ghostery panel and select detailed view
3. Rotate the smartphone to horizontal view
4. Click on any tracker name to open details
5. Try scrolling to the bottom - you can still scroll the background when you reach the bottom of the details


https://github.com/user-attachments/assets/2b1216cb-b660-4fd6-a6a4-8fc7b216ec93


https://github.com/user-attachments/assets/99a08ff4-64fd-4cf6-81ab-41607aa05609



**Expected results**
When the details of the tracker are visible, it should not be possible to scroll through the background in the landscape view.

Versions
```
GBE 10.4.2
```

GRadziejewski:
Fixed https://github.com/ghostery/ghostery-extension/issues/1828

smalluban:
It looks like Safari has broken support for dark mode detection in the extension window popup. On iOS it works fine, but on the macOS Safari 17.6, I have a light mode panel even though I should have dark.
You can check the media by hand using JS, so... This is a popup window:
Zrzut ekranu 2024-08-23 o 13 44 03
This is the Settings page:
Zrzut ekranu 2024-08-23 o 13 44 35

As a result, the dark mode is "broken" on Safari (macOS)... I am not sure what we should do with that. The only reasonable action is to report the bug, but they just don't do anything with that...

smalluban:
@GRadziejewski The scroll "issue" is an effect of fixing landscape mode - we had to introduce the scroll on the body level to make it work, which is very tricky to block from inner scrollable elements. Not much we can do with it. We still don't "support" that way of accessing to panel (it makes no sense), but it's better to give users access to the whole panel.

smalluban:
@GRadziejewski I slightly updated reds and yellows, you can check it out.

GRadziejewski:
@smalluban there is one more pop-up with a white background:
image

But aside from that, much better, LGTM.

smalluban:
@GRadziejewski The notification should be dark now + some minor fixes.

GRadziejewski (approved):
LGTM

chrmod:
Screenshot 2024-08-28 at 10 24 47
Screenshot 2024-08-28 at 10 19 53
For this element we could keep the white font in both versions.

chrmod:
Screenshot 2024-08-28 at 10 28 16
Screenshot 2024-08-28 at 10 20 51
Trackers Preview in light mode has a drop shadow that helps it to stand out from page content a bit more. In dark mode that shadow does not help as it is not visible. Maybe we should make a border bit brighter instead?

chrmod:
Screenshot 2024-08-28 at 10 24 36
The notifications bar that is fixed to the bottom is covering the panel content. Somehow in the dark mode, it is not obvious that panel content is scrollable.

smalluban:
@chrmod I made some minor fixes:

Zrzut ekranu 2024-08-28 o 11 00 52
Zrzut ekranu 2024-08-28 o 11 00 45

chrmod (approved):

PR #1840, opened 2024-08-26
Error reporting

chrmod:


extension-manifest-v3/src/background/healthcheck.js
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { captureException } from '/utils/errors.js';
+
+async function healthcheck() {
```
philipp-classen, line 14: Can we delay this a bit? Ideally, we should avoid to push non-essential IO not on the startup, because it makes it harder for other modules to initialize in time.

extension-manifest-v3/src/background/healthcheck.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    await indexedDB.databases();
+    await chrome.storage.local.get(['test']);
+  } catch (error) {
+    console.error(error);
+    captureException(error);
+  }
+}
+
+healthcheck();
```
philipp-classen, line 24: Can we delay this a bit? Ideally, we should avoid to push non-essential IO not on the startup, because it makes it harder for other modules to initialize in time.

philipp-classen (approved):

philipp-classen (approved):

extension-manifest-v3/src/background/healthcheck.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    await indexedDB.databases();
+    await chrome.storage.local.get(['test']);
+  } catch (error) {
+    console.error(error);
+    captureException(error);
+  }
+}
+
+setTimeout(healthcheck, 10000);
```
smalluban, line 24: I don't like calling it for 99% of users, which does not have that problem. We should add a `captureException` at least only in `engines.js`, where `loadFromStorage` or `saveToStarge` fails, meaning that `indexedDB` is broken.

smalluban (changes requested):

extension-manifest-v3/src/utils/errors.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+Sentry.init(config);
+
+getBrowserInfo().then(({ token }) => {
+  Sentry.setTag('ua', token);
+});
+
+export function captureException(error) {
+  console.error(error);
```
smalluban, line 44: It's a duplication of console.error()

smalluban (changes requested):

philipp-classen (dismissed):

smalluban (approved):

PR #1852, opened 2024-08-30
Enable extension testing on Firefox

seia-soto:
Firefox browser on Playwright doesn't allow us to start the browser with extension installed. This PR tries to fix it by installing the extension via remote control using RDP.

---

### TODO

**noSuchActor in getRoot request**

From my setup, I'm unable to get `addonsActor` property. I need to figure out if this is no longer possible on playwright version of Firefox browsers.

https://github.com/mozilla/web-ext/blob/master/src/firefox/remote.js#L73

```js
{
  from: 'root',
  error: 'noSuchActor',
  message: 'No such actor for ID: undefined'
}
```

### DIFF
--- a/extension-manifest-v3/tests/src/fixtures.js
+++ b/extension-manifest-v3/tests/src/fixtures.js
@@ -1,5 +1,7 @@
-import { test as base, chromium, firefox } from '@playwright/test';
+import { test as base } from '@playwright/test';
 import { downloadAddon } from '../src/helpers';
+import { loadChromiumBrowserWithExtension } from './loaders/chromium';
+import { loadFirefoxBrowserWithExtension } from './loaders/firefox';
 
 const extensionUrls = {
   firefox:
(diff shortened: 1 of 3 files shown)


chrmod:
Closing for now as it is not clear if we want to use playwright at all.

Besides, given that we use `web-ext`, the addon installation in firefox can be simplified with it's remote module, for example (example is puppeteer but works the same with playwright):

```js
import puppeteer from "puppeteer";
import getPort from "get-port";
import { connect } from "../node_modules/web-ext/lib/firefox/remote.js";

import getExtensionPath from "./extension.js";

const pathToExtension = getExtensionPath();

export default async function () {
  const rppPort = await getPort();
  const browser = await puppeteer.launch({
    headless: false,
    browser: "firefox",
    protocol: "cdp",
    args: [`--start-debugger-server=${rppPort}`],
    extraPrefsFirefox: {
      "devtools.chrome.enabled": true,
      "devtools.debugger.prompt-connection": false,
      "devtools.debugger.remote-enabled": true,
    },
  });

  const rdp = await connect(rppPort);
  await rdp.installTemporaryAddon(pathToExtension);
  return browser;
}
```

PR #1856, opened 2024-09-02
fix: merge adblocker filter lists to one main engine

### LINKED ISSUE #1846: Merge adblocker engines
The current, multi-engine architecture comes with number of complexities. Most notably, exceptions have to be shared between all engines (and currently it does not work as intended for the fixes and custom-filters engines). 
To simplify the runtime logic, we can use `FiltersEngine.merge` function to unite all engines into a single one. That way, when performing the blocking, we need to consider one engine only.

smalluban:
Fixes #1846

* Instead of loading multiple adblocker filter engines, we merge them into a single one
* Update process fixed to be sequential (avoid Firefox IO race conditions)
* Refactored and simplified code around the `/utils/engines.js` and `/background/adblocker.js`
* Added the `filtersUpdatedAt` option and `Last update: {date}` footer in the Settings page. It's clickable and forces the update the filter lists


smalluban:
I have found some edge cases when testing the update. Please wait with the review.

smalluban:
I solved the issues. It was related to creating the custom filters engine before the main was initialized - as there was a race. I added `setup.pending` to the custom filters.

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -24,102 +25,116 @@ import { getMetadata } from '/utils/trackerdb.js';
...
     }
 
-    // Schedule merge when one of the regional engines is updated
-    regions.forEach((id) => {
-      engines.addUpdateListener(`lang-${id}`, mergeRegionalEngines);
-    });
+    if (
+      // Experimental filters enabled
+      (options.experimentalFilters &&
+        lastValue?.experimentalFilters === false) ||
```
philipp-classen, line 114: If lastValue does not exist, this check will always fail. Is that expected?

extension-manifest-v3/src/utils/engines.js
```diff
@@ -406,107 +378,71 @@ async function update(name) {
...
-    }
-  });
-
   return (
-    get(name) || (await loadFromStorage(name)) || (await loadFromDisk(name))
+    get(name) ||
+    (await loadFromStorage(name)) ||
+    (name !== MAIN_ENGINE &&
+      name !== CUSTOM_ENGINE &&
+      (await loadFromDisk(name)))
```
philipp-classen, line 407: Perhaps we should rename "loadFromDisk" as loadBundledEngine (or something like that) to make it clear that it is the engine that comes as part of the release.

I find disk and storage confusing, because our storage is also on the local disk.

extension-manifest-v3/src/utils/indexeddb.js
```diff
@@ -22,15 +22,20 @@ export async function deleteDatabases() {
...
+
+      return new Promise((resolve, reject) => {
+        deleteDB(name, {
+          blocked() {
+            console.error(
+              `[utils|indexeddb] Failed to delete database ${name} because it is blocked`,
+            );
+            reject();
+          },
+        }).then(resolve);
```
philipp-classen, line 38:
```suggestion
        }).then(resolve).catch(reject);
```

Otherwise, the promise might hang.

extension-manifest-v3/src/utils/engines.js
```diff
@@ -217,27 +179,37 @@ function check(response) {
...
-    name === CUSTOM_ENGINE ||
-    name === REGIONAL_ENGINE ||
-    // If the IndexedDB is corrupted, and there is no way to load the engine
-    // from storage, we should not try to update it
-    (await loadFromStorage(name)) === null
-  ) {
+export async function update(name) {
+  // If the IndexedDB is corrupted, and there is no way to load the engine
+  // from the storage, we should skip the update.
+  if ((await loadFromStorage(name)) === null) {
```
philipp-classen, line 212: Could be worth logging a warning here that the update is skipped, because storage is not available.

extension-manifest-v3/src/background/adblocker.js
smalluban: Yes, as this should only pass through is `lastValue` is set.

In simple words the condition is: "When a user actively changed the setting to `true`"

extension-manifest-v3/src/utils/engines.js
smalluban: I am not sure (but I am an author)... ;) What do you think about `loadFromFilesystem`?

philipp-classen: loadFromFile would also be clearer IMHO, yes.

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -24,102 +25,116 @@ import { getMetadata } from '/utils/trackerdb.js';
...
+      // Filters outdated (1 hour)
+      options.filtersUpdatedAt < Date.now() - HOUR_IN_MS
+    ) {
+      const updateEngines = getEnabledEngines(options)
+        // Remove the custom engine, as it is created client-side only
+        .filter((name) => name !== engines.CUSTOM_ENGINE)
+        // Add engines outside of the main engine
+        .concat(engines.TRACKERDB_ENGINE);
```
chrmod, line 122:
```suggestion
      const updateEngines = new Set([
        ...getEnabledEngines(options),
        engines.TRACKERDB_ENGINE,
      ]);
      updateEngines.delete(engines.CUSTOM_ENGINE);
```

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -24,102 +25,116 @@ import { getMetadata } from '/utils/trackerdb.js';
...
+
+      (async () => {
+        for (const id of updateEngines) {
+          await engines.update(id).catch(() => null);
```
chrmod, line 126: This will make multiple `merge` operations.

How about we pass an option to not notify listeners and merge all engines after the for loop explicitly.

chrmod (changes requested):

extension-manifest-v3/src/background/adblocker.js
smalluban: It won't, only after all of the updates `store.set(Options ...)` (L129) triggers the merge.

extension-manifest-v3/src/background/adblocker.js
smalluban: Apple vs oranges ;) You create an array on the fly, trigger the spread operator (so iteration over all of the enabled engines), then also remove one of them.. Looks less readable, and I doubt that it's more preformant. This action is taken at most once an hour (or when user forces it). I recommend not to make premature optimizations.

extension-manifest-v3/src/background/adblocker.js
chrmod: ah, perfect

chrmod (dismissed):

smalluban:
@GRadziejewski I forgot about your part of the review ;) 

I have tested how the engines work with the PR on Chrome and Firefox, so I'll highlight the need for more careful smoke tests on that in the upcoming release.

PR #1859, opened 2024-09-03
fix:  improve error handing for IndexedDB access

smalluban:


extension-manifest-v3/src/utils/engines.js
```diff
@@ -143,12 +143,9 @@ async function loadFromStorage(name) {
       return engine;
     }
   } catch (e) {
-    captureException(e);
-    // If there is an error loading the engine from storage, the DB must be corrupted.
-    // In this case, we should delete it, as it will be reloaded on the next run.
-    await IDB.deleteDB('engines').catch((e2) =>
-      console.error('Failed to cleanup corrupted engine db', e2),
-    );
+    if (!e.message?.includes('serialized engine version mismatch')) {
```
chrmod, line 146: please ignore InvalidStateError as well so we wont flood Sentry with those errors

chrmod (changes requested):

extension-manifest-v3/src/utils/engines.js
smalluban: I used a part of the error message, I couldn't find a solution to use `InvalidStateError`

PR #1863, opened 2024-09-05
Fix missing telemetry fields

chrmod:
Those data points were always a part of the minimalistic telemetry signals that were used in Ghostery 8.

### DIFF
--- a/extension-manifest-v3/src/background/telemetry/index.js
+++ b/extension-manifest-v3/src/background/telemetry/index.js
@@ -54,6 +54,10 @@ const loadStorage = async () => {
 const getConf = async (storage) => {
   const options = await store.resolve(Options);
 
+  // Historically install_data was stored in Options.
+  // As it is used by telemetry only, it is here migrated
+  // to telemetry storage.
+  // TODO: cleanup Options after September 2024
   if (!storage.installDate) {
     saveStorage(storage, {
       installDate:
--- a/extension-manifest-v3/src/background/telemetry/index.js
+++ b/extension-manifest-v3/src/background/telemetry/index.js
@@ -54,11 +54,15 @@ const loadStorage = async () => {
 const getConf = async (storage) => {
   const options = await store.resolve(Options);
 
-  if (!storage.installDate) {
-    storage.installDate =
-      options.installDate || new Date().toISOString().split('T')[0];
-    storage.installRandomNumber = Math.floor(Math.random() * 100) + 1;
-    await saveStorage(storage, {});
+  if (!storage.installDate || !storage.installRandom) {
+    saveStorage(storage, {
+      installDate:
+        storage.installDate ||
+        options.installDate ||
+        new Date().toISOString().split('T')[0],
+      installRandom:
+        storage.installRandom || Math.floor(Math.random() * 100) + 1,
+    });
   }
 
   return {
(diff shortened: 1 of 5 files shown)


philipp-classen (approved):

extension-manifest-v3/src/store/options.js
```diff
@@ -83,6 +83,7 @@ const Options = {
     ...(__PLATFORM__ === 'opera' ? { serpShownAt: 0, serpShown: 0 } : {}),
   },
   installDate: '',
+  installRandomNumber: 0,
```
smalluban, line 86: Please use `installRandomId` or `installRandomID` - it's obvious that it's a number... ;) We don't use `installDateString` ;)

extension-manifest-v3/src/store/options.js
```diff
@@ -107,6 +108,14 @@ const Options = {
         options = __PLATFORM__ !== 'safari' ? await migrateFromV8() : {};
       }
 
+      if (!options.installDate) {
```
smalluban, line 111: I would move that configuration bootstrap into separate method, like `checkInstallMetrics(options)`

smalluban (changes requested):

extension-manifest-v3/src/background/telemetry/index.js
```diff
@@ -51,15 +51,23 @@ const loadStorage = async () => {
   return storage;
 };
 
-const getConf = async () => {
+const getConf = async (storage) => {
   const options = await store.resolve(Options);
 
+  if (!storage.installDate) {
+    storage.installDate =
+      options.installDate || new Date().toISOString().split('T')[0];
```
philipp-classen, line 59: Here (if it is not present in storage), it prefers it from options. Should the same be done with installRandomNumber? Or why is the options fallback needed? Also, how can the options be out-of-sync with the storage?

extension-manifest-v3/src/background/telemetry/index.js
```diff
@@ -51,15 +51,24 @@ const loadStorage = async () => {
...
 
-const getConf = async () => {
+const getConf = async (storage) => {
   const options = await store.resolve(Options);
 
+  if (!storage.installDate) {
+    saveStorage(storage, {
+      installDate:
+        options.installDate || new Date().toISOString().split('T')[0],
+      installRandom: Math.floor(Math.random() * 100) + 1,
```
chrmod, line 61: those are now part of the telemetry storage instead of options

extension-manifest-v3/src/background/telemetry/index.js
```diff
@@ -51,15 +51,24 @@ const loadStorage = async () => {
...
 };
 
-const getConf = async () => {
+const getConf = async (storage) => {
   const options = await store.resolve(Options);
 
+  if (!storage.installDate) {
+    saveStorage(storage, {
+      installDate:
+        options.installDate || new Date().toISOString().split('T')[0],
```
philipp-classen, line 60: That part still confuses me: why there is a fallback to options for installDate, but not for the random number. It is not obvious to me, why we need the fallback here? If storage.installDate is not present, how can options have it?

philipp-classen: If I understand correctly, options.installDate is migration code? And installRandom did not exist before.

smalluban (approved):

extension-manifest-v3/src/background/telemetry/index.js
chrmod: Comment added

PR #1864, opened 2024-09-06
Disable autoconsent cosmetic injection

seia-soto:
refs https://github.com/ghostery/broken-page-reports/issues/826

Autoconsent now enables cosmetic rules by default and it seems like this change can cause conflicts when creating exceptions. This is because their injections are done by style tag creation. This PR toggles off the style injections coming from autoconsent.

The breakage on `docs.spring.io` shows how this actually can be a problem because we may not override the style tag contents.

```html
<style id="autoconsent-css-rules">#modal-1 div[data-micromodal-close] { display: none !important; z-index: -1 !important; pointer-events: none !important; } </style>
```

By toggling the switch off, we solves
1. cosmetic injection conflicts (as we're already shipping EL cookies)
2. cosmetic exceptions/overrides cannot be applied due to style tag contents

### DIFF
--- a/extension-manifest-v3/src/background/autoconsent.js
+++ b/extension-manifest-v3/src/background/autoconsent.js
@@ -31,7 +31,9 @@ async function initialize(msg, tab, frameId) {
           action: 'autoconsent',
           type: 'initResp',
           rules,
-          config: {},
+          config: {
+            enableCosmeticRules: false,
+          },
         },
         {
           frameId,


smalluban:
Don't we expect them to work currently? Don't we remove the cosmetics for the CMPs from our filters?

seia-soto:
Yes, so we're doing the same job twice in actual:

- First happens in our annoyances/cookies lists
- Second happens in autoconsent side

However, those styles created by autoconsent adds extra complexity to debug what's happening in the website and fix the broken page by extra rules. I don't know if it's true that style tags have priority over user agent styles, but those were not able to disable.

seia-soto:
Additional context for

> Don't we remove the cosmetics for the CMPs from our filters?

We do, but it's only when autoconsent cannot work properly without the exceptions. So the baseline would be the application of cookie lists.

smalluban (approved):
Thanks for the explanation, it's fine to me.

chrmod (approved):

PR #1865, opened 2024-09-06
fix: fallback to storage.local when IndexedDB is corrupted

smalluban:
Relates to https://github.com/ghostery/ghostery-extension/issues/1837

Fixes the issue for engines with corrupted IndexedDB in Firefox when history is globally disabled or private browsing is turned on globally.

It adds a fallback for the eninges to be stored in `chrome.storage.local`. As Firefox supports saving UnitArrays, the implementation is stratgh forward. 

I tried few different approaches, but I think only two may apply - hardcoded fallback for Firefox for all users, or use "in-time" detection for each engine.

The first solution makes Firefox again a different, which I would like to avoid - we should keep the code the same as much as possible. Also, only a small group of users are affected by the Firefox bug.

The detection outside of the function and using a flag is hazardous. As the IndexedDB error message comes in asynchronius way, we cannot avoid a race condition, so the `loadFromStorage` function might be called before we set the flag. 

The last argument is about how the `getDB()` already works. It saves the opened session to the database in `getDB.current` - as this is done only once, for the each access to the database, after it's once rejected, it rejects immeditelly.

### DIFF
--- a/extension-manifest-v3/src/utils/engines.js
+++ b/extension-manifest-v3/src/utils/engines.js
@@ -143,6 +143,11 @@ async function saveToStorage(name) {
       await table.delete(name);
     }
 
+    if (__PLATFORM__ === 'firefox') {
+      // Clear out the fallback local storage if the engine is saved to the IDB
+      chrome.storage.local.set({ [`engines:${name}`]: null });
+    }
+
     await tx.done;
   } catch (e) {
     if (__PLATFORM__ === 'firefox') {
--- a/extension-manifest-v3/src/utils/engines.js
+++ b/extension-manifest-v3/src/utils/engines.js
@@ -94,7 +94,9 @@ async function loadFromStorage(name) {
       .then((db) => {
         const tx = db.transaction('engines');
         const table = tx.objectStore('engines');
-        return table.get(name);
+        return table.get(name).then((result) => {
+          return tx.done.then(() => result);
+        });
       })
       .catch((e) => {
         // Suppress the private browsing mode in Firefox
@@ -138,6 +140,8 @@ async function saveToStorage(name) {
     } else {
       await table.delete(name);
     }
+
+    await tx.done;
   } catch (e) {
     if (__PLATFORM__ === 'firefox') {
       const key = `engines:${name}`;
--- a/extension-manifest-v3/src/utils/engines.js
+++ b/extension-manifest-v3/src/utils/engines.js
@@ -90,12 +90,25 @@ async function getDB() {
 
 async function loadFromStorage(name) {
   try {
-    const db = await getDB();
+    const engineBytes = await getDB()
+      .then((db) => {
+        const tx = db.transaction('engines');
+        const table = tx.objectStore('engines');
+        return table.get(name);
+      })
+      .catch((e) => {
+        // Suppress the private browsing mode in Firefox
+        if (!e.message?.includes('database that did not allow mutations')) {
+          captureException(e);
+        }
 
-    const tx = db.transaction('engines');
-    const table = tx.objectStore('engines');
-
-    const engineBytes = await table.get(name);
+        if (__PLATFORM__ === 'firefox') {
+          const key = `engines:${name}`;
+          return chrome.storage.local.get([key]).then((data) => data[key]);
+        } else {
+          throw e;
+        }
+      });
 
     if (engineBytes) {
       const engine = deserializeEngine(engineBytes);
@@ -104,17 +117,6 @@ async function loadFromStorage(name) {
       return engine;
     }
   } catch (e) {
-    const msg = e.message || '';
-
-    if (
-      // Adblocker core updates the engine version
-      !msg.includes('serialized engine version mismatch') &&
-      // Private Browsing mode in Firefox
-      !msg.includes('database that did not allow mutations')
-    ) {
-      captureException(e);
-    }
-
     console.error(`[engines] Failed to load engine "${name}" from storage`, e);
   }
 
@@ -123,15 +125,28 @@ async function loadFromStorage(name) {
 
 async function saveToStorage(name) {
   const engine = loadFromMemory(name);
-  const db = await getDB();
+  const serialized = engine?.serialize();
 
-  const tx = db.transaction('engines', 'readwrite');
-  const table = tx.objectStore('engines');
+  try {
+    const db = await getDB();
 
-  if (engine) {
-    await table.put(engine.serialize(), name);
-  } else {
-    await table.delete(name);
+    const tx = db.transaction('engines', 'readwrite');
+    const table = tx.objectStore('engines');
+
+    if (engine) {
+      await table.put(serialized, name);
+    } else {
+      await table.delete(name);
+    }
+  } catch (e) {
+    if (__PLATFORM__ === 'firefox') {
+      const key = `engines:${name}`;
+      const data = engine ? { [key]: serialized } : { [key]: null };
+
+      return chrome.storage.local.set(data);
+    }
+
+    throw e;
   }
 }
 
--- a/extension-manifest-v3/src/manifest.firefox.json
+++ b/extension-manifest-v3/src/manifest.firefox.json
@@ -14,6 +14,7 @@
     "webNavigation",
     "webRequest",
     "webRequestBlocking",
+    "unlimitedStorage",
     "http://*/*",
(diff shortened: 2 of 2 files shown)


extension-manifest-v3/src/utils/engines.js
```diff
@@ -90,12 +90,25 @@ async function getDB() {
...
+        return table.get(name);
+      })
+      .catch((e) => {
+        // Suppress the private browsing mode in Firefox
+        if (!e.message?.includes('database that did not allow mutations')) {
+          captureException(e);
+        }
+
+        if (__PLATFORM__ === 'firefox') {
+          const key = `engines:${name}`;
```
philipp-classen, line 106: Would be worth logging a warning if the fallback is used.

extension-manifest-v3/src/utils/engines.js
```diff
@@ -123,15 +125,28 @@ async function loadFromStorage(name) {
...
+    const table = tx.objectStore('engines');
+
+    if (engine) {
+      await table.put(serialized, name);
+    } else {
+      await table.delete(name);
+    }
+  } catch (e) {
+    if (__PLATFORM__ === 'firefox') {
+      const key = `engines:${name}`;
```
philipp-classen, line 143: Would be worth logging a warning if the fallback is used.

extension-manifest-v3/src/utils/engines.js
```diff
@@ -123,15 +125,28 @@ async function loadFromStorage(name) {
...
-  } else {
-    await table.delete(name);
+    const tx = db.transaction('engines', 'readwrite');
+    const table = tx.objectStore('engines');
+
+    if (engine) {
+      await table.put(serialized, name);
+    } else {
+      await table.delete(name);
+    }
```
philipp-classen, line 140: At this point, we have successfully updated IndexedDB. It is the source of truth, and I assume it is likely that it could be read on the startup.

So, I wonder if we should do a best-effort (fire-and-forget operation) on Firefox here to purge the fallback in chrome.storage.local?

extension-manifest-v3/src/utils/engines.js
```diff
@@ -123,15 +125,28 @@ async function loadFromStorage(name) {
...
-  const tx = db.transaction('engines', 'readwrite');
-  const table = tx.objectStore('engines');
+  try {
+    const db = await getDB();
 
-  if (engine) {
-    await table.put(engine.serialize(), name);
-  } else {
-    await table.delete(name);
+    const tx = db.transaction('engines', 'readwrite');
```
chrmod, line 133: btw. should't we do `await tx.done` after the put/delete?

extension-manifest-v3/src/utils/engines.js
```diff
@@ -90,12 +90,25 @@ async function getDB() {
...
-
-    const tx = db.transaction('engines');
-    const table = tx.objectStore('engines');
-
-    const engineBytes = await table.get(name);
+    const engineBytes = await getDB()
+      .then((db) => {
+        const tx = db.transaction('engines');
+        const table = tx.objectStore('engines');
+        return table.get(name);
```
chrmod, line 97: please add `await tx.done`

extension-manifest-v3/src/utils/engines.js
smalluban: Not sure, but there is `tx.done` promise, so I'll add it.

extension-manifest-v3/src/utils/engines.js
smalluban: I am not sure about it, as it will flood the console, as this is called each time the BG starts.

extension-manifest-v3/src/utils/engines.js
smalluban: This is valid question, but we would need to read storage for 99% users, which don't need it. You cannot know if the "bug" state was, it is not here, and we should clean up the storage.

smalluban:
@GRadziejewski Testing highlights after technical review:

Chromium-based browsers and Safari should not be affected by any of the changes - it should work exactly as before - you can check if engines are loaded correctly (cosmetics).

On Firefox it should be exactly the same as above for default browser settings. However, if user disables history or creates a global private mode, without the change engines are loaded from disk each time background process starts. We can test both cases, but testing the latter is very difficult - loading unpacked extension in "no-history" mode fails (it is some internal Firefox bug) - you must use Firefox Nightly, set `xpinstall.signatures.required` in `about:config` to false, then use `.zip` to install the extension.

extension-manifest-v3/src/utils/engines.js
```diff
@@ -92,12 +92,27 @@ async function getDB() {
 
 async function loadFromStorage(name) {
   try {
-    const db = await getDB();
-
-    const tx = db.transaction('engines');
-    const table = tx.objectStore('engines');
-
-    const engineBytes = await table.get(name);
+    const engineBytes = await getDB()
```
philipp-classen, line 95: Perhaps using async-syntax would make it slightly easier to read than promise chaining? Just an opinion:

```
  try {
    const db = await getDB();
    const tx = db.transaction('engines');
    const table = tx.objectStore('engines');
    const result = await table.get(name);
    await tx.done;
    return result;
  } catch (e) {
    // Suppress the private browsing mode in Firefox
    if (!e.message?.includes('database that did not allow mutations')) {
      captureException(e);
    }

    if (__PLATFORM__ === 'firefox') {
      const key = `engines:${name}`;
      const data = await chrome.storage.local.get([key]);
      return data[key];
    } else {
      throw e;
    }
  }
  ```

File: extension-manifest-v3/src/utils/engines.js (hunk 3 above)
Comment (philipp-classen) [reply extension-manifest-v3/src/utils/engines.js:144]: That is true. Might not be the best place to do it here. I'm fine with leaving that out of the PR here.

Leaving old engines around is probably not harmful. They have some space overhead, but I don't think it results in inconsistencies. For problems like that, we could think of having a different place in the extension (e.g. during an extension update when migrations happen), where we could put cleanup. Keeping timestamps is generally useful, because then it possible to detect data that has expired.

Comment (philipp-classen) [reply extension-manifest-v3/src/utils/engines.js:144]: As said, I'm fine with leaving it out. Though when I think about it, the overhead of a delete to storage should not be a issue, since it happens in a path where we update the engine. This is an expensive IO write operation already. In that path, having an extra delete of a value that will almost certainly not exist, is avoidable overhead for most Firefox users, but probably not noticeable.

File: extension-manifest-v3/src/utils/engines.js (hunk 2 above)
Comment (smalluban) [reply extension-manifest-v3/src/utils/engines.js:149]: It was a general comment to both places  - We have already logs if the function successfully saves the engine - regardless of the target.

File: extension-manifest-v3/src/utils/engines.js (hunk 3 above)
Comment (smalluban) [reply extension-manifest-v3/src/utils/engines.js:144]: I thought that you meant the `loadFromStorage`. You're right, as the saveToStorage is not that often, I can add it there.

File: extension-manifest-v3/src/utils/engines.js (hunk 6 above)
Comment (smalluban) [reply extension-manifest-v3/src/utils/engines.js:95]: We still needs a `try/catch` around the deserialization - for example, it throws for adblocker version mismatch - I am not sure about nested try/catch...

Comment (philipp-classen) [reply extension-manifest-v3/src/utils/engines.js:95]: Alright. I have no strong opinion either.

**Review by chrmod — APPROVED** [review]

**Comment by smalluban** [comment]
@GRadziejewski There is technical approval. If possible, please test out the PR. There is a tip on how to test it faster, the extension can be installed form source, but you must enable it in private windows to make it work (then restart the extension).

**Comment by GRadziejewski** [comment]
## Smoke tests Iteration no 1:
Tests of: https://github.com/ghostery/ghostery-extension/pull/1865#issuecomment-2337690735

---
Checked features:

- Enable/Disable Ad-blocking 
- Enable/Disable Anti-tracking
- Enable/Disable Never Consent 
- Pause Ghostery
- Regional filters
- Check YT

```
Browser: Chrome 128.0.6613.120
OS: Windows 10
GBE 10.4.4

Browser: Firefox 130.0
OS: Windows 10
GBE 10.4.4

Browser: Firefox Nightly 132.0a1
OS: Windows 10
GBE 10.4.4
```
---


## Issues:
- 

## Fixed:
- 

## Cannot reproduce:
- 

## Will not be fixed:
-

GRadziejewski (approved):

PR #1867, opened 2024-09-06
Update vite

chrmod:
The builds starts just fine so maybe we don't need to patch vite anymore.


philipp-classen (approved):

smalluban (changes requested):
I doubt that in current architecture we can avoid patching Vite, as we have "external" packages, and they must be loaded before static imports (the reason behind the first patch). Another is watching `node_modules`, which is required until we remove "mono-repo" architecture completely. 

Here you have an example of difference between this PR and main result of imports:

```js
// This PR
import __vite_glob_0_0 from './components/alert.js';
import __vite_glob_0_1 from './components/button.js';
import __vite_glob_0_2 from './components/card.js';
import __vite_glob_0_3 from './components/container.js';
import __vite_glob_0_4 from './components/copy.js';
import __vite_glob_0_5 from './components/dialog.js';
import __vite_glob_0_6 from './components/feedback.js';
import __vite_glob_0_7 from './components/menu-item.js';
import __vite_glob_0_8 from './components/navigation-card.js';
import __vite_glob_0_9 from './components/notification.js';
import __vite_glob_0_10 from './components/options-item.js';
import __vite_glob_0_11 from './components/options.js';
import __vite_glob_0_12 from './components/pause.js';
import Home from './views/home.js';
import __vite_glob_0_14 from './views/navigation.js';
import __vite_glob_0_15 from './views/protection-status.js';
import __vite_glob_0_16 from './views/tracker-details.js';
import '../../npm/@ghostery/ui/src/modules/panel/index.js';
import define from '../../npm/hybrids/src/define.js';
import mount from '../../npm/hybrids/src/mount.js';
import router from '../../npm/hybrids/src/router.js';
import { html } from '../../npm/hybrids/src/template/index.js';

// main
import '../../npm/@ghostery/ui/src/modules/panel/index.js';
import Home from './views/home.js';
import __vite_glob_0_0 from './components/alert.js';
import __vite_glob_0_1 from './components/button.js';
import __vite_glob_0_2 from './components/card.js';
import __vite_glob_0_3 from './components/container.js';
import __vite_glob_0_4 from './components/copy.js';
import __vite_glob_0_5 from './components/dialog.js';
import __vite_glob_0_6 from './components/feedback.js';
import __vite_glob_0_7 from './components/menu-item.js';
import __vite_glob_0_8 from './components/navigation-card.js';
import __vite_glob_0_9 from './components/notification.js';
import __vite_glob_0_10 from './components/options-item.js';
import __vite_glob_0_11 from './components/options.js';
import __vite_glob_0_12 from './components/pause.js';
import __vite_glob_0_14 from './views/navigation.js';
import __vite_glob_0_15 from './views/protection-status.js';
import __vite_glob_0_16 from './views/tracker-details.js';
import define from '../../npm/hybrids/src/define.js';
import mount from '../../npm/hybrids/src/mount.js';
import router from '../../npm/hybrids/src/router.js';
import { html } from '../../npm/hybrids/src/template/index.js';
```

On main all `__vite_glob_*` are imported after panel node module...

smalluban:
Normally, it would not be a problem, but we have "external" configuration of the `i18n` - which must be called before anything else.. This is why the panel package must be loaded before components are generated by static imports.

chrmod:
@smalluban got it. I've brought back the patch. 
This is however a very esoteric knowledge. Do you think we can remove this special patching once the "monorepo" is gone?

smalluban:
I think we will, the thing is very special, as our needs are special. As the code was shared between packages/extensions.. it must have been like that.

smalluban (commented):
I suggest to close this PR, and back with the update after moving packages into the source. Then it should be possible to avoid patching. I would do that at once. 

Currently, updating Vite doesn't give us much, as we mainly use roll-up behind it (it's still convenient to use Vite, as it gives much configuration by default).

chrmod:
I'm fine to do it later, but not that old vite trigger security warning from npm.

PR #1868, opened 2024-09-06
Telemetry: send active faster and remove upgrade signals

chrmod:
From what I can tell, the `active` signal was only send on subsequent service worker startup. In practice, it should not make much difference, but in principle it should be send as soon as possible. 
Additionally, `upgrade` signal is no longer is use, so the was removed from the code.

### DIFF
--- a/extension-manifest-v3/src/background/telemetry/index.js
+++ b/extension-manifest-v3/src/background/telemetry/index.js
@@ -113,6 +113,9 @@ chrome.runtime.onMessage.addListener((msg) => {
     log('Telemetry recordUTMs() error', error);
   }
 
-  telemetry.ping(JUST_INSTALLED ? 'install' : 'active');
+  if (JUST_INSTALLED) {
+    telemetry.ping('install');
+  }
+  telemetry.ping('active');
   telemetry.setUninstallUrl();
 })();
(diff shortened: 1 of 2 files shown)


extension-manifest-v3/src/background/telemetry/index.js
```diff
@@ -113,6 +113,9 @@ chrome.runtime.onMessage.addListener((msg) => {
     log('Telemetry recordUTMs() error', error);
   }
 
-  telemetry.ping(JUST_INSTALLED ? 'install' : 'active');
+  if (JUST_INSTALLED) {
+    telemetry.ping('install');
+  }
+  telemetry.ping('active');
```
philipp-classen, line 119: Since service workers, there should be not much difference compared to the old version. So, I'm indifferent about the change here.

However, I wonder sending the `active` signal as soon as possible is actually worse than intentionally delaying it. The main question is if an extension that has been installed and not being active for at least a while should be considered active? (Perhaps one hour is a reasonable threshold.)

By delaying, the data could be less noisy. The loss is that clients that installed it, but instantly stop using it, will not contribute to the active population. It is like what counts as a guest in a restaurant? Anyway, who entered the building, or someone who spend some time?

smalluban (approved):

extension-manifest-v3/src/background/telemetry/index.js
chrmod: I would also prefer to change the semantics of `active` to something indicating actual activity, no just presence. But maybe instead of changing it would be better to introduce a new signal and compare the two?

This PR is meant to bring parity with Ghostery 8, so we compare apples to apples.

PR #1871, opened 2024-09-06
TrackerDB updates

chrmod:


extension-manifest-v3/src/utils/engines.js
```diff
@@ -271,7 +271,11 @@ export async function update(name) {
     }
 
     // Make a full update if we need to remove some lists
```
smalluban, line 273: Please update the comment to reflect the condition with TrackerDB

smalluban (changes requested):

extension-manifest-v3/src/utils/engines.js
```diff
@@ -271,7 +271,11 @@ export async function update(name) {
...
-    if (foundListsToRemove) {
+    if (
+      foundListsToRemove ||
+      (name === 'trackerdb' &&
+        data.lists[0].checksum !== engine.lists[0].checksum)
```
smalluban, line 277: Currently, the trackerdb does not have any list - the code assumes that it exists, and will throw if there are no `lists[0]`. We must be sure that the "new" TrackerDB will be included in the release, or add

```suggestion
        data.lists[0] && data.lists[0].checksum !== engine.lists[0].checksum)
```

smalluban: Ok, it's even worse.. the `engine.lists` is a Map...

smalluban (changes requested):
I checked, and the trackerdb lists don't provide diffs, so the L264 makes `foundListsToRemove` be `true` without your additional condition.

smalluban (approved):

PR #1875, opened 2024-09-10
Adblocker: inject scriptlets to a proper frame

chrmod:
fixes https://github.com/ghostery/broken-page-reports/issues/760


smalluban (approved):

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -391,21 +391,21 @@ async function injectScriptlets(tabId, url) {
...
-    executeScriptlets(tabId, scriptlets);
+    executeScriptlets(tabId, frameId, scriptlets);
   }
 }
 
 if (__PLATFORM__ === 'safari') {
   chrome.runtime.onMessage.addListener((msg, sender) => {
     if (sender.url && msg.action === 'injectScriptlets') {
-      injectScriptlets(sender.tab.id, sender.url);
+      injectScriptlets(sender.tab.id, sender.frameId, sender.url);
```
seia-soto, line 401: I really don't expect the case that Safari doesn't give `frameId` because we're always accessing to tab information. However, it would be nice if we can measure how it'll be important to fallback with zero value (means main_frame in Safari).

seia-soto (approved):

smalluban (changes requested):
The linked broken page report lists sites, where I cannot reproduce the problem, or the fix does not help:

Cannot reproduce on `main`: https://thesignatureedge.com/, http://www.badgertrust.org.uk/, https://www.bassistik-live.com/

With Ghostery the site hangs (like infinite loop in JS) - https://www.athom.tech/ - regardless of the fix (`main`/ this PR).

seia-soto:
@smalluban I think their was a scriptlet conflict after the fix. Please, add the following filters to the custom filters and try to click navigation items on https://www.thesignatureedge.com/gallery .

```adb
athom.tech,bassistik-live.com,badgertrust.org.uk,thesignatureedge.com,blackmountainhoney.co.uk#@#+js(set, Sentry, {})
```

After that to solve, add the following filters to disable bad filters which will inject scriptlets into invalid frames.

```adb
ecom.wixapps.net#@#+js(set, Sentry, {})
ecom.wixapps.net#@#+js(set, Sentry.init, noopFunc)
```

smalluban (approved):
The test case works, with the custom filters on this branch, and the navigation is back to normal.

PR #1878, opened 2024-09-10
fix: unify adblocker engines update process

smalluban:
A few minor fixes to the adblocker engines maintenance and update process:

* Avoids reloading the main engine when no engines were updated
* Simplifies the logic in `observe()` pattern, to make clear what when happens
* Protects `updateEngines` from calling the function before the update finishes, as there are multiple factors, which trigger the update

The update process was simplified and moved to one place. It is only triggered by the `updateEngines()` function. Also, the update is done only, if the user passes the onboarding, and then it saves the timestamp.

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -72,11 +72,11 @@ async function reloadMainEngine() {
     );
 
     console.info(
-      `[adblocker] engine reloaded with: ${enabledEngines.join(', ')}`,
+      `[adblocker] Main engine reloaded with: ${enabledEngines.join(', ')}`,
     );
   } else {
     engines.create(engines.MAIN_ENGINE);
-    console.info('[adblocker] Engine reloaded with no filters');
+    console.info('[adblocker] Main engine reloaded with no filters');
   }
 }
 
@@ -90,7 +90,6 @@ async function updateEngines() {
 
   try {
     updating = true;
-
     const enabledEngines = getEnabledEngines(options);
 
     if (enabledEngines.length) {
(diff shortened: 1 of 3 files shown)


extension-manifest-v3/src/background/adblocker.js
```diff
@@ -19,6 +19,7 @@ import { parse } from 'tldts-experimental';
 import Options, { observe, ENGINES, isPaused } from '/store/options.js';
 
 import * as engines from '/utils/engines.js';
+import * as trackerdb from '/utils/trackerdb.js';
```
chrmod, line 22: trackerdb is already imported - please merge import statements

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -70,18 +72,48 @@ async function reloadEngine(enabledEngines) {
...
+        if (id === engines.CUSTOM_ENGINE) continue;
+        updated = (await engines.update(id).catch(() => false)) || updated;
+      }
+
+      // Reload the main engine after all engines are updated
+      if (updated) await reloadMainEngine();
 
+      // Update TrackerDB engine
+      trackerdb.setup.pending && (await trackerdb.setup.pending);
+      await engines.update(engines.TRACKERDB_ENGINE).catch(() => null);
```
chrmod, line 106: shouldn't trackerdb update even if no other engine is enabled?

smalluban: I made it on purpose - I think after all, it's best to all only onboarded users to update the trackerdb more often than a release - those not onboarded will have an update with a new release. It also simplifies the case by setting the timestamp only when engines are enabled. Without that, it must be set not to update trackerdb repeatedly if the user doesn't want to onboard.

chrmod: okay, wanted to be sure that's the case

chrmod (approved):

PR #1881, opened 2024-09-13
feat: use scripting.ExecutionWorld enum for executeScript world

seia-soto:
This fixes the autoconsent and scriptlet breakages on Firefox version greater than 127 where the scripting.ExecutionWorld.MAIN was implemented. See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/ExecutionWorld for more information. From Firefox version 128, the value of undefined given to world property of executeScript's first argument will fallback to ISOLATED.

The use of scripting.ExecutionWorld enum for world argument is better than defining in literal since we don't need to test if browser supports the execution world of MAIN. Unless browser supports MAIN world execution, the value will fallback to undefined which is safe.

---

fixes https://github.com/ghostery/broken-page-reports/issues/834
fixes https://github.com/ghostery/broken-page-reports/issues/837
fixes https://github.com/ghostery/broken-page-reports/issues/838
fixes https://github.com/ghostery/broken-page-reports/issues/839
fixes https://github.com/ghostery/broken-page-reports/issues/846
fixes https://github.com/ghostery/broken-page-reports/issues/845

### DIFF
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -349,7 +349,7 @@ ${scripts.join('\n\n')}}
   chrome.scripting.executeScript(
     {
       injectImmediately: true,
-      world: __PLATFORM__ === 'firefox' ? undefined : 'MAIN',
+      world: chrome.scripting.ExecutionWorld.MAIN,
       target: {
         tabId,
         allFrames: true,
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -349,7 +349,10 @@ ${scripts.join('\n\n')}}
   chrome.scripting.executeScript(
     {
       injectImmediately: true,
-      world: chrome.scripting.ExecutionWorld?.MAIN || 'MAIN',
+      world:
+        chrome.scripting.ExecutionWorld?.MAIN ?? __PLATFORM__ === 'firefox'
+          ? undefined
+          : 'MAIN',
       target: {
         tabId,
         allFrames: true,
--- a/extension-manifest-v3/src/background/adblocker.js
+++ b/extension-manifest-v3/src/background/adblocker.js
@@ -350,9 +350,8 @@ ${scripts.join('\n\n')}}
     {
       injectImmediately: true,
       world:
-        chrome.scripting.ExecutionWorld?.MAIN ?? __PLATFORM__ === 'firefox'
-          ? undefined
-          : 'MAIN',
+        chrome.scripting.ExecutionWorld?.MAIN ??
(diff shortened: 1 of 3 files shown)


seia-soto:
refs https://github.com/ghostery/broken-page-reports/issues/834
refs https://github.com/ghostery/broken-page-reports/issues/837
refs https://github.com/ghostery/broken-page-reports/issues/838
refs https://github.com/ghostery/broken-page-reports/issues/839

chrmod (dismissed):

smalluban (dismissed):

philipp-classen (dismissed):

GRadziejewski:
Is there anything I could test at the moment? From what I understand, I should wait for the build to be compiled to test it properly. I'm unassigning myself from the list, so if @seia-soto, you think I can sit down and do some testing, go ahead and assign me back.

seia-soto:
> Is there anything I could test at the moment? From what I understand, I should wait for the build to be compiled to test it properly. I'm unassigning myself from the list, so if @seia-soto, you think I can sit down and do some testing, go ahead and assign me back.

Oops, sorry. I don't think so. I thought the packages will be automatically generated but it's actually not. I think you can test this change with next upcoming release, as this doesn't look like including critical changes. Thanks!

chrmod:
Have tested this and had to make small updates to make it work on Safari. It is now ready to merge.

smalluban (dismissed):

philipp-classen (dismissed):

extension-manifest-v3/src/background/adblocker.js
```diff
@@ -349,7 +349,10 @@ ${scripts.join('\n\n')}}
   chrome.scripting.executeScript(
     {
       injectImmediately: true,
-      world: __PLATFORM__ === 'firefox' ? undefined : 'MAIN',
+      world:
+        chrome.scripting.ExecutionWorld?.MAIN ?? __PLATFORM__ === 'firefox'
```
smalluban, line 353: Are you sure about the priority of `??` against the `?` From the code styling, it looks like when L353 is truth then the result is `undefined`. If this is not your intention, I would add brackets from `__PLATFORM__ === 'firefox`...

I checked, and definitely it's not the same:

```js
 chrome.scripting.ExecutionWorld?.MAIN ??
        (__PLATFORM__ === 'firefox' ? undefined : 'MAIN'),
```

smalluban (changes requested):

extension-manifest-v3/src/background/adblocker.js
chrmod: good catch

chrmod: I actually didn't know about the difference in priority of `||` and `??`

smalluban (approved):

PR #1884, opened 2024-09-16
fix: youtube & iframe util from packages/ui to main source

smalluban:
* Moves out `youtube` from `packages/ui` to the main source 
* Moves and cleans up the `iframe` feature
* The notification flows go through the background process and single content script (avoids duplication of the mounting mechanism in Youtube content script)
* Adds `pages/notifications` to group all notifications in one path
* Cleans up the `opera-serp` util
* Adds `debugMode` for easier testing and `stagingMode` for allowing engines to update from staging CDN

I tested the Youtube Wall notification (Chrome, Firefox, Safari) and Opera Serp (Opera).


extension-manifest-v3/src/utils/opera.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+import { setCookie } from '/utils/api.js';
+
+const TEST_COOKIE_NAME = `ghostery:opera:cookie:test:${Date.now()}`;
+
+let isSupported = undefined;
```
smalluban, line 15: After enabling SERP pages, the browser reloads the extension. We can use that fact to make a check only once.

extension-manifest-v3/src/background/notifications.js
```diff
@@ -0,0 +1,50 @@
...
+
+import * as notifications from '/utils/notifications.js';
+
+export function openNotification(tabId, id, params) {
+  const url = chrome.runtime.getURL(
+    `/pages/notifications/${id}.html` +
+      (params
+        ? `?${Object.entries(params)
+            .map(([key, value]) => `${key}=${encodeURIComponent(value)}`)
+            .join('')}`
```
philipp-classen, line 20:
```suggestion
            .join('&')}`
```

extension-manifest-v3/src/utils/opera.js
```diff
@@ -0,0 +1,48 @@
...
+export async function isSerpSupported() {
```
philipp-classen, line 16: From what I see, this function may be called concurrently. If so, could setting the cookies then conflict?

extension-manifest-v3/scripts/build.js
```diff
@@ -57,13 +57,14 @@ if (argv.target.startsWith('safari')) {
   argv.target = 'safari';
 }
 
+// Add flags to manifest
+if (argv.debug) manifest.debug = true;
+if (argv.staging) manifest.staging = true;
```
chrmod, line 62: was wondering why you "invent" the stating mode, but from the other changes it looks like it was implicitly there for some subsystems.

We have to decide how those modes should work. 

Historically:
* debug - was setting all backends to staging and was enabling some additional logging
* staging - was recently introduced to download adblocker engines from the staging environment

It looks like we have 3 separate aspects to deal with:
* debugging of the addon (logs, some settings like opera serp, scritplets debugging)
* adblocker engines environment
* backends (staging vs production)

extension-manifest-v3/src/background/onboarding.js
```diff
@@ -26,9 +31,28 @@ observe('onboarding', (onboarding) => {
 });
 
 if (__PLATFORM__ === 'chromium' && isOpera()) {
-  chrome.webNavigation.onCompleted.addListener((details) => {
+  const NOTIFICATION_DELAY = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds
+  const NOTIFICATION_SHOW_LIMIT = 4;
+
+  chrome.webNavigation.onCompleted.addListener(async (details) => {
```
chrmod, line 37: can we please move this code outside of the onboarding?

extension-manifest-v3/src/utils/engines.js
```diff
@@ -221,7 +221,7 @@ function notifyListeners(name) {
   });
 }
 
-const CDN_HOSTNAME = chrome.runtime.getManifest().debug
+const CDN_HOSTNAME = stagingMode
```
chrmod, line 224: in the metrics we use debugMode instead - I think we should do it here as well.

chrmod (changes requested):
The cleanup looks good. Only few nits to cleanup.

extension-manifest-v3/scripts/build.js
smalluban: The staging mode usually does not work. They it yourself. The engines are not updated there (I suppose we have to trigger it manually). 

Firstly, there was only a debug mode, but with that issue I had to add separate flag for loading engines from staging.

extension-manifest-v3/src/background/onboarding.js
smalluban: Ok, I'll create an opera.js file for that.

extension-manifest-v3/src/utils/engines.js
smalluban: Sure. I think I did that looking for URLs to the CDN. 

As we discuss that, I am considering if we should add then the debug mode flag into the main `npm start` script instead of another command. When the staging engines was combined with debug mode, it wasn't a good idea. 

Now, I think, that we could do that for access to those hard to test features by default, and avoid sending pings/WTM signals to the production environment when developing.

extension-manifest-v3/scripts/build.js
chrmod: ah, so you mean even if we downloaded staging engine pre-build, they would update to production engine in the runtime anyways?

I'm fine to keep the staging mode. Shall we then update all endpoints configuration to use stagingMode instead of debugMode?

extension-manifest-v3/src/utils/engines.js
chrmod: reading this and other comments convinces me that we should have `stagingMode` here and in other places the configure endpoints

extension-manifest-v3/scripts/build.js
smalluban: `npm start -- --staging` fails... as the staging CDN does not have the proper version of the engines..

For me `--staging` should only set the engines' endpoints - as this must be used only when we know that we pushed something to the staging CDN for engines and only engines.

chrmod: the problem occurs as we don't create staging engines on every adblocker release - perhaps we should. When this happen please reach out so we can schedule a CI build.

PR #1886, opened 2024-09-17
fix: disable essential telemetry until onboarded

chrmod:


extension-manifest-v3/src/background/telemetry/index.js
```diff
@@ -113,9 +114,16 @@ chrome.runtime.onMessage.addListener((msg) => {
     log('Telemetry recordUTMs() error', error);
   }
 
-  if (JUST_INSTALLED) {
-    telemetry.ping('install');
-  }
-  telemetry.ping('active');
+  observe('terms', async (terms) => {
+    telemetryEnabled = terms;
```
smalluban, line 118: This function runs at least once every background run. You don't need a flag, just check terms and call pings.

smalluban (changes requested):

extension-manifest-v3/src/background/telemetry/metrics.js
```diff
@@ -390,7 +379,7 @@ class Metrics {
         this.storage[`${type}_${frequency}`] = timeNow;
         this.saveStorage(this.storage);
```
philipp-classen, line 380: Here, we are missing an `await`. IMHO, better to send too few messages than too many.

extension-manifest-v3/src/background/telemetry/index.js
chrmod: telemetry is always there, as it is used to set the uninstall url

smalluban (approved):
Ok,we can leave it as it is until we change the uninstall URL setup. 

I strongly suggest to not initialize telemetry just for setting the uninstall URL, which before the onboarding should be clean, so don't need to go through the telemetry at all.

philipp-classen (approved):

chrmod:
@smalluban I completely agree. In next iteration I will re-write telemetry code from the ground-up.

PR #1892, opened 2024-09-20
fix: add TrackerDB `Utilities` category

### LINKED ISSUE #1885: Add translations for new "utlity" category
Background: The "essential" category in TrackerDB as been [renamed as "utility"](https://github.com/ghostery/trackerdb/pull/423/files).

Two places where we need to provide translations:

1. Ghostery settings (first entries that used to be "Essentials" has no label)
Screenshot 2024-09-17 at 12-12-30 Ghostery settings

2. Panel (first entries that used to be "Essentials" has no label)
panel

smalluban:
Fixes #1885

* Adds `utilities` category, which replaces the `essential`
* Keeps the `essential`, as the TrackerDB must be updated when all of the platforms update to the version with the change

@philipp-classen We should clean up the `essential` category once we update. Please create an issue, when you will update the TrackerDB.

Also, it would be great to push the "changed" TrackerDB to the staging CDN, so we can test the change.


packages/ui/src/utils/categories.js
```diff
@@ -40,6 +41,7 @@ export const order = [
   'site_analytics',
   'consent',
   'essential',
```
chrmod, line 43: should we remove `essential`

philipp-classen:
> Fixes #1885
> 
>     * Adds `utilities` category, which replaces the `essential`
> 
>     * Keeps the `essential`, as the TrackerDB must be updated when all of the platforms update to the version with the change
> 
> 
> @philipp-classen We should clean up the `essential` category once we update. Please create an issue, when you will update the TrackerDB.
> 
> Also, it would be great to push the "changed" TrackerDB to the staging CDN, so we can test the change.

Unfortunately, is not possible with our current infrastructure. Once TrackerDB is modified, it will propagate to production. That means, first releasing the update here to all platforms and then updating TrackerDB is what I would suggest.

philipp-classen (approved):

PR #1897, opened 2024-09-23
fix: move packages/ui into the main source

smalluban:
Moves all of the modules of `packages/ui` into the main source:
* Module `global` into the `src/ui`, other modules into the specific pages, where they were used
* Unifies the panel/onboarding styling into a single `ui/styles.css` loaded automatically with the `ui` module
* Unifies component naming prefixes - global component with `ui-`, page components with page name prefix, like `onboarding-` or `settings-`
* Minor fixes to the global components - missing hover states, cleared font sizing, etc

I did all of the smoke tests on Firefox and Chrome. However, the PR contains mainly visual changes, so please check out the package too when reviewing.


chrmod (approved):
I've scan the changes and all looks good. The PR could have been smaller if you would decide tho keep the previous naming scheme with `gh-` prefix.
We have to test the changes manually to be sure all if fine.

philipp-classen (approved):

PR #1901, opened 2024-09-26
fix: update panel enable button to use danger type

### LINKED ISSUE #1900: change button color to red for "Enable Ghostery" when user skips onboarding
same shade as "Permissions required" 
Image

smalluban:
Fixes #1900

Zrzut ekranu 2024-09-26 o 15 01 15


mdsandu (approved):
danger type on

PR #1906, opened 2024-09-27
Eslint update

chrmod:


philipp-classen (approved):

eslint.config.js
```diff
@@ -0,0 +1,34 @@
+import globals from 'globals';
```
smalluban, line 1: Is this required to have JS configuration, not static `.eslintrc`? It's much less readable, the globals can be hardcoded. I don't see much value in that.

src/pages/panel/views/tracker-details.js
```diff
@@ -168,12 +168,11 @@ export default {
               <div layout="column gap:2">
                 <div layout="column gap">
                   ${tracker.requestsBlocked.map(
-                    ({ url }) =>
-                      html`
-                        <panel-copy oncopy="${showCopyNotification}">
-                          ${url}
-                        </panel-copy>
-                      `,
+                    ({ url }) => html`
```
smalluban, line 171: Finally, they fixed it.

src/ui/wheel.js
```diff
@@ -94,7 +94,7 @@ export function getOffscreenImageData(size, categories, options) {
   let canvas;
   try {
     canvas = new OffscreenCanvas(size, size);
-  } catch (e) {
+  } catch {
```
smalluban, line 97: I thought it was a recent feature, but it was introduced in ES2019, so removing unused catch parameters is quite safe.

PR #1916, opened 2024-10-01
chore(tests): add e2e tests with WebdriverIO

smalluban:
* Setup e2e with [wdio](https://webdriver.io/)
* Adds where required `data-qa` IDs in UI for reliable selectors
* Adds support for opening the panel in a separate window
* Fixes few minor bugs found when creating tests

There are known issues in WebDriver BiDi implementation in Firefox, such as https://github.com/w3c/webdriver-bidi/issues/753. Because of that, we have to monitor the situation with tests in Firefox, as they might fail without an issue on our side. I added 5 retries to the specs, and usually after the first retry the spec passes.


tests/e2e/wdio.conf.js
```diff
@@ -0,0 +1,27 @@
...
+      browserName: 'chrome',
+      'goog:chromeOptions': {
+        args: (DEBUG ? [] : ['headless', 'disable-gpu']).concat([
+          `--load-extension=${path.join(__dirname, '..', '..', 'dist')}`,
+          '--disable-search-engine-choice-screen',
+        ]),
+      },
+    },
+  ],
+  injectGlobals: false,
```
chrmod, line 21: what does this do?

tests/e2e/specs/onboarding.js
```diff
@@ -0,0 +1,25 @@
+import { browser, expect } from '@wdio/globals';
+
```
chrmod, line 2: how about a different folder structure?
```
tests/e2e/onboarding.spec.js
tests/unit/adblocker.spec.js
```

tests/e2e/utils.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export async function navigateToPage(page, file = 'index.html') {
+  await browser.url(
+    `chrome-extension://${await getExtensionId()}/pages/${page}/${file}`,
+  );
+}
+
+export function getQAElement(qa) {
+  return $(`[data-qa="${qa}"]`);
```
chrmod, line 22: I think the common pattern is to call this attribute `data-testid`

tests/e2e/utils.js
```diff
@@ -0,0 +1,23 @@
...
+  return extensionId;
+}
+
+export async function navigateToPage(page, file = 'index.html') {
+  await browser.url(
+    `chrome-extension://${await getExtensionId()}/pages/${page}/${file}`,
+  );
+}
+
+export function getQAElement(qa) {
```
chrmod, line 21: maybe `getByTestId`?

tests/e2e/utils.js
```diff
@@ -0,0 +1,23 @@
+import { browser, $ } from '@wdio/globals';
+
+let extensionId = '';
+export async function getExtensionId() {
+  if (!extensionId) {
+    await browser.url('chrome://extensions');
```
chrmod, line 6: this will influence the context of the test. 

maybe instead of checking in the runtime, we can include a `key` in the manifest to have a fixed extension id? 
like it is done here https://github.com/ghostery/webextension-event-recorder/blob/ae9005bc60d4385312730d8ea067f53230144fbb/extension/manifest.json#L6
(key will be there in debug builds only)

tests/e2e/wdio.conf.js
smalluban: Turns off injecting the `browser`, `$$`, and `$` into the global contexts of specs. I prefer to import them explicitly.

tests/e2e/utils.js
smalluban: The tests should run on the package closest to the one we publish. This is the recommended way to get the extension id from the WDIO docs. What do you mean by that it will influence the context? How can opening the settings page impact tests?

tests/e2e/specs/onboarding.js
smalluban: There are some configuration files, which will share the same folder with specs then. I am not sure. I would keep the `specs` subpath.

tests/e2e/utils.js
smalluban: The WDIO has the `getElement` API, i think we will never try to query our own DOM structure differently, so I even consider renaming it to `getElement`.

tests/e2e/utils.js
chrmod: This function will load `extensions` page and keep it open (maybe you should just close the tab after the id is acquired). And it will happen first time the extensionid is accessed, which may not be expected behavior when you filter tests.


Including `key` in the manifest should be safe as  `key` is does not influence the extension behavior.


Also, this exact method wont work on Firefox.

tests/e2e/specs/onboarding.js
chrmod: this is also my concern, I would prefer to be able to tell if a file is a test or a util based on its name, hence `.spec.` infix.

tests/e2e/utils.js
chrmod: I'm good with `getElement`, but maybe we can be more explicit by calling it `getExtensionElement` and `navigateToExtensionPage` ?

tests/e2e/utils.js
chrmod: to me QA (Quality Assurance, is quite bit more than tests) but I also don't have a strong opinion. `qa` is short, that's definitely a benefit. `data-testid` is a convention I saw in other projects, I think it comes from https://testing-library.com/docs/queries/bytestid/

how about we simply use regular `id`s ?

tests/e2e/wdio.conf.js
```diff
@@ -0,0 +1,27 @@
+import path from 'node:path';
```
chrmod, line 1: why don't we put this file in the project root?

package.json
```diff
@@ -12,8 +12,10 @@
     "start:update": "./scripts/update.sh",
     "debug": "npm run build -- --watch --debug",
     "licenses": "license-report --config=../.license-report-config.json > dist/licenses.html",
-    "lint": "eslint src/",
-    "test": "node --test && npm run lint",
+    "lint": "eslint",
+    "test": "npm run build && npm run wdio",
+    "wdio": "wdio tests/e2e/wdio.conf.js",
+    "wdio:debug": "DEBUG=true wdio tests/e2e/wdio.conf.js",
```
chrmod, line 18: `DEBUG=true npm run wdio`

package.json
```diff
@@ -12,8 +12,10 @@
     "start:update": "./scripts/update.sh",
     "debug": "npm run build -- --watch --debug",
     "licenses": "license-report --config=../.license-report-config.json > dist/licenses.html",
-    "lint": "eslint src/",
-    "test": "node --test && npm run lint",
+    "lint": "eslint",
+    "test": "npm run build && npm run wdio",
```
chrmod, line 16: not sure if `build` should be a part of the `test`. cannot we make user to build it by themselves before running tests?

tests/e2e/wdio.conf.js
chrmod: very good, that's also my preference. I was confused as you still have global `describe`.

smalluban: This is how mocha works, I am not sure if we can import `describe`, `it` and others directly.

tests/e2e/specs/onboarding.js
smalluban: We can go with the suffix, no problem. I'll try flattening the structure, it might be not that bad.

tests/e2e/utils.js
smalluban: That was my first idea. However, sometimes we need that for another purpose (usual CSS/JS stuff), and the `id` can be only one.

tests/e2e/utils.js
smalluban: The WDIO uses the same tab to navigate. You can try out the `npm run build && npm run wdio:debug` to see how it works.

tests/e2e/wdio.conf.js
smalluban: The debug version is so different, that it is clearer to have two saparate configurations. Because of that we have two files, and this is only required for running end-to-end tests, so I thought that keeping it nested is clearer.

tests/wdio.conf.js
```diff
@@ -0,0 +1,65 @@
...
+        },
+      ],
+      injectGlobals: false,
+      reporters: ['spec'],
+      logLevel: 'silent',
+      mochaOpts: { timeout: 24 * 60 * 60 * 1000 },
+      before: async (capabilities, specs, browser) => {
+        if (capabilities.browserName === 'firefox') {
+          const extension = await fs.readFile(
+            path.join(extPath, 'ghostery-firefox.zip'),
```
chrmod, line 60: both browsers are able to run unpacked addons. maybe webdriverio does have a first class support for that, but it is not hard to achieve.

chrmod (changes requested):

tests/wdio.conf.js
smalluban: If you mean loading not exported internal part of the `web-ext` library for Firefox, I would keep it as it is:
https://github.com/ghostery/webextension-event-recorder/blob/main/src/firefox.js#L3

I would even prefer to load zips in both browsers, but Chrome does not support it (as far as I have checked all the arguments and documentation). 

And after all, what's wrong with loading the zip? The WDIO driver supports that out of the box, we want to test the production version, and we have to make both builds parallel - so still I would have to copy the `dist/` somewhere.

GRadziejewski: Unpacking the extension increases the duration of the test, which is not some big problem, as well as generating a few lines of extra code to maintain, which is also not a problem. Nevertheless, in my opinion, it is always safer to unzip first - it can be considered as a kind of safeguard that the extension is downloaded at all.

chrmod: Chrome loads unpacked extensions just fine https://github.com/ghostery/webextension-event-recorder/blob/e41cd33260f48128ef5e195b35f2b607316b62e6/src/chrome.js#L12

smalluban: I think @GRadziejewski misunderstood that - it is just the opposite, but the extension is loaded only once per session, for all tests... In Chrome we load the unpacked version (there is no way to load zip), and in Firefox we load zip (as in my opinion loading the unpacked version does not give any advantage, and adds a compilation to the implementation). After all, it has no impact on running tests.

chrmod: when it comes to `web-ext` we can make a PR to project to export the `remote`

tests/e2e/privacy.spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+async function updatePrivacySettings(name, value) {
+  await browser.url(await getExtensionPageURL('settings'));
+
+  const toggle = await getExtensionElement(`toggle:${name}`);
+  if ((await toggle.getProperty('value')) !== value) {
+    await toggle.click();
+  }
+
+  await browser.pause(1000);
```
chrmod, line 27: this may not be long enough for the DNR to update. but I understand that currently we have no visible side effect to wait for

scripts/e2e.sh
```diff
@@ -0,0 +1,21 @@
+rm -rf ./web-ext-artifacts
+mkdir ./web-ext-artifacts
+
+# Build and copy Chrome extension
+npm run build -- --silent
+if [ $? -ne 0 ]; then
```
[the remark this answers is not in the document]
smalluban: Where? Istead of `if` or add `-e` to `npm run` ... ?

tests/e2e/privacy.spec.js
```diff
@@ -0,0 +1,113 @@
...
+}
+
+describe('Main features', () => {
+  before(enableExtension);
+
+  describe('Never-consent', () => {
+    beforeEach(() => updatePrivacySettings('never-consent', false));
+
+    const WEBSITE_URL = 'https://stackoverflow.com/';
+    const SELECTOR = '#onetrust-consent-sdk';
```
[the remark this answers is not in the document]
smalluban: This is a pattern to make it easier to switch to another website, if the test become flaky. The `SELECTOR` is used also in the next test, so it's shared.

tests/e2e/privacy.spec.js
```diff
@@ -39,7 +39,7 @@ describe('Main features', () => {
 
     it('displays consent popup', async () => {
       await browser.url(WEBSITE_URL);
-      await browser.pause(5000);
+      await browser.pause(2000);
```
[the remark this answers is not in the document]
smalluban: Yes, we do in the next line. I think this is a leftover.

tests/wdio.conf.js
```diff
@@ -0,0 +1,71 @@
...
+
+import path from 'node:path';
+import url from 'node:url';
+import { promises as fs } from 'node:fs';
+
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url));
+const extPath = path.join(__dirname, '..', 'web-ext-artifacts');
+
+export let config = {
+  specs: ['onboarding.spec.js', 'privacy.spec.js'],
```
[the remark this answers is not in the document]
smalluban: For the debug version the specs are merged into a single session, so the order of execution is important - the `onboarding` must be the first. We can add a glob here, but to ensure `onboarding` is always the first one, we need to add a prefix to the name. I think listing files explicitly is good enough, as we will not have too many specs.

scripts/e2e.sh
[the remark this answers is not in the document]
chrmod: start the file with
```sh
#!/bin/bash
set -e
```

tests/e2e/privacy.spec.js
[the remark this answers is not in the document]
chrmod: I get that. IMO this is not making the test better as when I have to debug a flaky test I would have to scroll the whole file to find which url is being tested.

package.json
```diff
@@ -12,8 +12,10 @@
     "start:update": "./scripts/update.sh",
     "debug": "npm run build -- --watch --debug",
     "licenses": "license-report --config=../.license-report-config.json > dist/licenses.html",
-    "lint": "eslint src/",
-    "test": "node --test && npm run lint",
+    "lint": "eslint",
+    "test": "./scripts/e2e.sh",
```
chrmod, line 16: this wont work on windows

scripts/build.js
```diff
@@ -51,6 +51,7 @@ const argv = process.argv.slice(2).reduce(
 );
 
 const pkg = JSON.parse(readFileSync(resolve(pwd, 'package.json'), 'utf8'));
+const silent = argv.silent;
```
chrmod, line 54: why this is needed?

tests/wdio.conf.js
```diff
@@ -0,0 +1,71 @@
...
+        },
+      },
+    ],
+    logLevel: 'error',
+    mochaOpts: { timeout: 24 * 60 * 60 * 1000 },
+  };
+} else {
+  config = {
+    ...config,
+    specFileRetries: 5,
```
chrmod, line 44: would prefer to avoid that. if the tests are so flaky that we have many false negatives, then this setting may generate false positives

tests/wdio.conf.js
```diff
@@ -0,0 +1,71 @@
...
+          ],
+        },
+      },
+      {
+        browserName: 'firefox',
+        'moz:firefoxOptions': { args: ['-headless'] },
+      },
+    ],
+    before: async (capabilities, specs, browser) => {
+      if (capabilities.browserName === 'firefox') {
```
chrmod, line 63: we could execute the build here

chrmod (approved):

tests/wdio.conf.js
smalluban: Read the description carefully - this is for buggy web driver implementation in Firefox, this is not the issue of the tests itself.

tests/wdio.conf.js
smalluban: No, as I want to have a way to run tests locally without building them every time. I can run `npm run wdio` when I change test, but I don't need to re-build the extension.

PR #1923, opened 2024-10-04
chore(build): download rule resources from CDN

seia-soto:
This PR deprecates the way of redirect resource extraction from engine as we added proper redirect resources distribution via adblocker-filters CDN.


chrmod (approved):

chrmod:
@seia-soto I don't really see a difference between the two approaches :)

PR #1927, opened 2024-10-07
feat(tests): Add e2e tests for main privacy features

smalluban:
* Adds e2e tests for all main features listed in `tests/README.md`
* Updates `tests/README.md` with an icon 🤖 to indicate that test is covered with the e2e suite
* Fixes a rare issue (from time to time I saw it when testing), when `www` subdomain was added to paused domain (thanks to e2e tests)


tests/e2e/privacy.spec.js
```diff
@@ -56,50 +63,143 @@ describe('Privacy', () => {
     });
   });
 
-  describe('Ad-Blocking', () => {
+  describe.only('Ad-Blocking', () => {
```
chrmod, line 66: remove only

tests/e2e/privacy.spec.js
```diff
@@ -25,6 +28,10 @@ async function updatePrivacySettings(name, value) {
   }
 
   await expect(toggle).toHaveElementProperty('value', value);
+
+  // Allow background process to update the settings
+  // E.g. reload engines / DNR rules
+  await browser.pause(2000);
```
chrmod, line 34: would be best to create a github issue to provide visual feedback to the settings page for which those tests could wait to confirm the action was successful (or not successful)

tests/e2e/privacy.spec.js
```diff
@@ -56,50 +63,143 @@ describe('Privacy', () => {
...
 
     it('blocks ads on a page', async () => {
       await updatePrivacySettings('ad-blocking', true);
 
+      await browser.url(PAGE_URL);
+      await expect($(SELECTOR)).not.toBeDisplayed();
+    });
+  });
+
+  describe('Anti-Tracking', () => {
```
chrmod, line 84: only noticed now, but if this is mocha then please use function statements instead of arrow functions

tests/e2e/privacy.spec.js
```diff
@@ -56,50 +63,143 @@ describe('Privacy', () => {
...
+    beforeEach(() => updatePrivacySettings('anti-tracking', false));
+
+    const TRACKER_IDS = ['facebook_connect', 'pinterest_conversion_tracker'];
+
+    it('does not block tracker requests on the page', async () => {
+      await disableCache();
+      await browser.url(PAGE_URL);
+
+      await switchToPanel(async () => {
```
chrmod, line 93: how about a different syntax:
```js
const goBack = await switchToPanel();

// regular tests

await goBack();
```

tests/e2e/privacy.spec.js
```diff
@@ -56,50 +63,143 @@ describe('Privacy', () => {
...
+    it("pauses the website's privacy settings", async () => {
+      await browser.url(PAGE_URL);
+
+      await switchToPanel(async () => {
+        const pauseComponent = await getExtensionElement('component:pause');
+        const pauseButton = await getExtensionElement('button:pause');
+
+        if (await pauseComponent.getProperty('paused')) {
+          await pauseButton.click();
+          await browser.pause(1000);
```
chrmod, line 174: can we await some ui change here, like a class name added?

tests/e2e/utils.js
```diff
@@ -35,17 +43,12 @@ export async function getExtensionId() {
...
-    : 'moz-extension';
-
 export async function getExtensionPageURL(page, file = 'index.html') {
-  return `${SCHEME}://${await getExtensionId()}/pages/${page}/${file}`;
+  return `${browser.isChromium ? 'chrome-extension' : 'moz-extension'}://${await getExtensionId()}/pages/${page}/${file}`;
 }
 
 export function getExtensionElement(id) {
-  return $(`[data-qa="${id}"]`);
+  return $(`>>>[data-qa="${id}"]`);
```
chrmod, line 51: what does `>>>` do?

tests/e2e/utils.js
```diff
@@ -54,6 +57,9 @@ export async function enableExtension() {
 
   await getExtensionElement('button:enable').click();
   await expect(getExtensionElement('view:success')).toBeDisplayed();
+
+  // Give the extension some time to initialize (updating the engines in the background)
+  await browser.pause(5000);
```
chrmod, line 62: lets create a github issue to have some visual feedback once the extension is fully loaded

tests/wdio.conf.js
```diff
@@ -11,61 +11,131 @@
...
+        }
+      }
+    } catch (e) {
+      console.error('Error while building the extension');
+      console.error(e);
+
+      process.exit(1);
+    }
+
+    const file = await fs.readFile(
```
chrmod, line 93: lets use `readFileSync`

tests/e2e/utils.js
```diff
@@ -11,8 +11,16 @@
 
 import { browser, expect, $ } from '@wdio/globals';
 
+export async function disableCache() {
+  if (browser.isChromium) {
+    await browser.sendCommand('Network.setCacheDisabled', {
```
chrmod, line 16: isn't there a way to disable network cache with Chrome command line arguments?

chrmod (changes requested):

src/store/tab-stats.js
```diff
@@ -84,7 +84,7 @@ const Stats = {
         return tabStats;
       }
 
-      const { hostname } = parse(tab.url);
+      const hostname = parse(tab.url).hostname?.replace(/^www\./, '');
```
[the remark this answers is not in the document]
smalluban: It was a bug, as I wrote in the description - I am fixing it here. Wherever we get the hostname, we clear the `www` subdomain. Here we didn't.

tests/e2e/privacy.spec.js
smalluban: I will be hard, as those are listeners to the options, which also triggers in other situations. For example merging engines happens for many reasons, but after switching some of the toggles, it happens.

tests/e2e/privacy.spec.js
smalluban: Sure, good catch.

tests/e2e/privacy.spec.js
smalluban: Why?

tests/e2e/privacy.spec.js
smalluban: Switching can throw errors, so you can hang in some state, where the focus could still be on another window. 

This pattern ensures, that whatever happens, after `await switchToPanel(...)` you are back to the main context. Your pattern does not guarantee that.

tests/e2e/utils.js
smalluban: No, there was in 2025, but they removed the flag.

tests/e2e/utils.js
smalluban: Ensure that the selector goes into the Shadow DOM

tests/wdio.conf.js
```diff
@@ -11,61 +11,131 @@
...
+    );
+    const server = createServer((req, res) => {
+      res.writeHead(200, { 'Content-Type': 'text/html' });
+      res.end(file);
+    });
+
+    // starts a simple http server locally on port 6789
+    server.listen(PAGE_PORT, '127.0.0.1', () => {
+      console.log('Testing page server listening on', PAGE_URL);
+    });
```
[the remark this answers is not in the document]
smalluban: I would be careful with that, as there are multiple sessions running the same test file. If the server would start there, you will have a problem with setting dynamically the port, etc.. I would keep it as simple as possible until it can't fulfill our requirements.

tests/e2e/privacy.spec.js
```diff
@@ -56,50 +63,143 @@ describe('Privacy', () => {
...
+    it('shows the ads on the page', async () => {
       await browser.url(WEBSITE_URL);
+      await expect($(SELECTOR)).toBeDisplayed();
+    });
+
+    it('hides the ads on the page', async () => {
+      await updatePrivacySettings('regional-filters', true);
+      const checkbox = await getExtensionElement(
+        'checkbox:regional-filters:pl',
+      );
```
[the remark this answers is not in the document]
smalluban: This is the opposite - whatever the language of the browser is, it ensures that the `pl` region will be turned off/turned on.

tests/e2e/privacy.spec.js
smalluban: I checked, and we wait for the result on the panel itself, so the pause here is not needed.

tests/e2e/privacy.spec.js
chrmod: https://mochajs.org/#arrow-functions

tests/e2e/privacy.spec.js
chrmod: if it hangs then it should timeout, if something else goes wrong, it should throw.

not sure if I see the benefit of nested syntax

tests/wdio.conf.js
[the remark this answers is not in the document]
chrmod: ok for now

tests/e2e/privacy.spec.js
chrmod: Today we use wdio, tomorrow it may be using puppeteer, but most likely we will always use mocha :)

tests/e2e/privacy.spec.js
chrmod: gotcha, lets discuss if we don't want to change this architecture as having some sorts of visual feedback may be useful for some users, not only for testing. I would definitely want to get confirmation if DNR ruleset got applied successfully for example.

tests/e2e/privacy.spec.js
smalluban: You haven't convinced me, but I have changed it, as this a detail - if you think that's better, let it be ;)

tests/e2e/privacy.spec.js
smalluban: I would have to show you an example, but working with real browser is not that easy. In your pattern, there is no explicit clean up, if something goes wrong between switch and go back. This is required, as the browser may stay focused on the incorrect window. Because of that, the whole code must be in try/catch call, and a clean up of the switch must be made if something throws an error.

chrmod (approved):

PR #1941, opened 2024-10-09
Fix prettier configuration

chrmod:


smalluban:
The release with the latest commit fixes the problem in `engines.js` - but you can do it here too.

smalluban (approved):

PR #1943, opened 2024-10-10
fix(dnr): add support for `excludedResourceTypes` in Safari converter

smalluban:
Filters like

>@@||*^$~image,~stylesheet,~script,~object,~font,~media,~subdocument,~ping,~xmlhttprequest,~websocket,~other

were converted to an extensive DNR rule, which allows everything, as the exceptions were not used.

PR #1948, opened 2024-10-11
Ad-blocking: better DNR rulesets for Safari

chrmod:
surpasses https://github.com/ghostery/ghostery-extension/pull/1943

Removed compatibility rules are no longer required as Safari ignores incompatibilities silently so the support for those features is now correct.

### DIFF
--- a/src/manifest.safari-ios.json
+++ b/src/manifest.safari-ios.json
@@ -46,12 +46,17 @@
       {
         "id": "ads",
         "enabled": false,
-        "path": "rule_resources/dnr-safari.json"
+        "path": "rule_resources/dnr-ads.json"
+      },
+      {
+        "id": "annoyances",
+        "enabled": false,
+        "path": "rule_resources/dnr-annoyances.json"
       },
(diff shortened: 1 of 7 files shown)


smalluban (approved):
It looks promising. My only concern is which version `Safari ignores incompatibilities silently `. Should we bump the minimum version of Safari with that change?

src/manifest.safari-ios.json
```diff
@@ -46,12 +46,17 @@
       {
         "id": "ads",
         "enabled": false,
-        "path": "rule_resources/dnr-safari.json"
+        "path": "rule_resources/dnr-ads.json"
+      },
+      {
+        "id": "annoyances",
+        "enabled": false,
+        "path": "rule_resources/dnr-annoyances.json"
```
smalluban, line 54: What about `dnr-tracking` ?

chrmod: it does not fit the 75k limit :(

PR #1950, opened 2024-10-11
fix(tests): Make runner more stable & add update process testing

smalluban:
* Clears out the self-opening onboarding page after extension installation - it made some test flaky
* Moves config files into the `e2e` path and specs to `spec` subdirectory - as the file number grows, it is more readable when everything is in the `tests/e2e` scope
* Renames specs, so we can use glob pattern for future use without changing the config
* Adds `npm run wdio:update` command to test suite with updated extension from the latest release


tests/e2e/spec/_onboarding.spec.js
```diff
@@ -19,9 +23,11 @@ describe('Onboarding', function () {
     await getExtensionElement('button:skip').click();
     await expect(getExtensionElement('view:skip')).toBeDisplayed();
 
-    await browser.url(await getExtensionPageURL('panel'));
+    await switchToPanel(async function () {
+      await expect(getExtensionElement('button:enable')).toBeDisplayed();
+    });
 
-    await expect(getExtensionElement('button:enable')).toBeDisplayed();
+    await browser.url('about:blank');
```
chrmod, line 30: maybe put it in `afterEach`?

tests/e2e/utils.js
```diff
@@ -24,18 +24,20 @@ async function getExtensionId() {
   if (!extensionId) {
     switch (browser.capabilities.browserName) {
       case 'chrome': {
-        await browser.url('chrome://extensions');
+        const url = 'chrome://extensions';
+        await browser.url(url);
```
chrmod, line 28: why we need a variable ?

chrmod (commented):
First time I see extension update tests. Really cool stuff @smalluban !

tests/e2e/utils.js
smalluban: It must be a leftover.. ;)

tests/e2e/wdio.conf.js
```diff
@@ -0,0 +1,169 @@
...
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url));
+export function setupTestPage() {
+  const file = readFileSync(path.join(__dirname, 'page.html'), 'utf8');
+  const server = createServer((req, res) => {
+    res.writeHead(200, { 'Content-Type': 'text/html' });
+    res.end(file);
+  });
+
+  // starts a simple http server locally on port 6789
+  server.listen(PAGE_PORT, '127.0.0.1', () => {
```
chrmod, line 61: we should close the server after the test

tests/e2e/wdio.update.conf.js
```diff
@@ -0,0 +1,143 @@
...
+  existsSync,
+} from 'node:fs';
+import { execSync } from 'node:child_process';
+import { $, expect } from '@wdio/globals';
+
+import { getExtensionPageURL } from './utils.js';
+
+import * as wdio from './wdio.conf.js';
+
+export const config = {
```
chrmod, line 27: maybe add a comment on what this config is doing

tests/e2e/wdio.conf.js
smalluban: This server is supposed to run during the whole testing process. It is set up only once and cleared out by the Nodejs itself. Look a the nodejs.org example from the home page - it does not closes itself too ;)

chrmod (approved):

PR #1951, opened 2024-10-11
Release v10.4.10

smalluban:


AdamGhst (changes requested):
CHROME
Smoke tests 1-14: PASSED
YouTibe: PASSED

FIREFOX
Smoke tests 1-14: PASSED
YouTibe: PASSED

SAFARI:
Smoke tests 1-9 And 11-14 PASSED.
Smoke test 10 FAILED: exeption tool for trackers does not work. urls are still blocked, not observed. 
YouTube: FAILED. Ghostery block ads only with first page/video load. After reloading the Youtube, video ads are not visible, but for the time of their lenghts we can see loading window. after that, end card of ad is visible. Realoding Youtube restarts the problem. Only first page/video load prevents ads to start. 
image

smalluban:
@AdamGhst The fix for Safari is in build 41. Please re-test exceptions. However, I couldn't reproduce the issue on YT. If the exceptions are fine, I think we should go with it.

Issue #1952, opened 2024-10-11
[QA] Tests of 10.4.9


GRadziejewski:
Issue nr 1.

Description
The Software Licenses page is broken. Its formatting is missing.

Steps to Reproduce

1. Open Ghostery panel
2. Open Ghostery Settings
3. Scroll down and click on "Software Licenses"
4. New tab is opened.
![image](https://github.com/user-attachments/assets/a06f7436-c163-4d39-80a3-ba94738095ce)

Expected results
The formatting of the licence page should not be broken.

```
Browser: Opera 113.0.5230.142
OS: Windows 10 Pro
GBE: 10.4.9

Browser: Chrome 129.0.6668.100
OS: Windows 10 Pro
GBE: 10.4.9
```

PR #1953, opened 2024-10-11
Fix: licenses page

chrmod:
fixes https://github.com/ghostery/ghostery-extension/issues/1952


smalluban (approved):

smalluban:
What about adding a e2e test for that page, to be sure that the fix works?

smalluban:
To not delay the fix, I'll think about adding the test later.

PR #1956, opened 2024-10-14
fix(onboarding): update the condition to support users updated from v8

smalluban:
The function uses only the `onboarding` scope to avoid too many executions when options change. 

I've checked and `setup_shown` is set to `1` and not cleared in `v8`. 

The update process does not open the onboarding anymore.


smalluban (commented):
As the updated users have `options.onboarding.shownAt === 0`, I think it might be worth to clear out that option at all. It was cleared out in all v8 versions, @chrmod?

chrmod (approved):

PR #1958, opened 2024-10-14
fix(dnr): Filter unsupported `regexFilter` when converting for Safari

smalluban:
This PR brings back the filtering rules with `regexFilter` syntax support. There is a `chrome.declarativeNetRequest.isRegexSupported` API, but it's not available in offscreen documents, and it's asynchronous. The code change would affect Chromium browsers as well. I prefer to do it in a separate PR to avoid blocking the `v10.4.10` release.


smalluban:
Closing in favor of #1959

PR #1964, opened 2024-10-15
chore(tests): Add  e2e tests for custom filters

smalluban:
This relates to #1960. As we have a change in custom filters, I've created e2e tests for the feature.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -95,25 +95,27 @@ async function updateEngines() {
 
       // Update engines from the list of enabled engines
       await Promise.all(
-        enabledEngines.map((id) =>
-          engines.update(id).then(
-            (v) => {
-              updated = updated || v;
-            },
-            () => {},
+        enabledEngines
+          .filter((id) => id !== engines.CUSTOM_ENGINE)
+          .map((id) =>
+            engines.update(id).then(
+              (v) => {
+                updated = updated || v;
+              },
+              () => {},
+            ),
           ),
-        ),
       );
 
       // Reload the main engine after all engines are updated
       if (updated) await reloadMainEngine();
 
-      // Update timestamp after the engines are updated
-      await store.set(Options, { filtersUpdatedAt: Date.now() });
-
       // Update TrackerDB engine
       trackerdb.setup.pending && (await trackerdb.setup.pending);
       await engines.update(engines.TRACKERDB_ENGINE).catch(() => null);
+
+      // Update timestamp after the engines are updated
+      await store.set(Options, { filtersUpdatedAt: Date.now() });
     }
   } finally {
     updating = false;
@@ -153,7 +155,9 @@ export const setup = asyncSetup([
     engines.setEnv('env_experimental', value);
 
     // Experimental filters changed to enabled
-    if (lastValue !== undefined && value) updateEngines();
+    if (lastValue !== undefined && value) {
+      await updateEngines();
+    }
   }),
 ]);
 
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -92,21 +92,28 @@ async function updateEngines() {
 
     if (enabledEngines.length) {
       let updated = false;
+
       // Update engines from the list of enabled engines
-      for (const id of enabledEngines) {
-        if (id === engines.CUSTOM_ENGINE) continue;
-        updated = (await engines.update(id).catch(() => false)) || updated;
-      }
+      await Promise.all(
+        enabledEngines.map((id) =>
+          engines.update(id).then(
+            (v) => {
+              updated = updated || v;
+            },
+            () => {},
+          ),
+        ),
+      );
 
       // Reload the main engine after all engines are updated
       if (updated) await reloadMainEngine();
 
+      // Update timestamp after the engines are updated
+      await store.set(Options, { filtersUpdatedAt: Date.now() });
+
       // Update TrackerDB engine
       trackerdb.setup.pending && (await trackerdb.setup.pending);
       await engines.update(engines.TRACKERDB_ENGINE).catch(() => null);
-
-      // Update timestamp after the engines are updated
-      await store.set(Options, { filtersUpdatedAt: Date.now() });
     }
   } finally {
     updating = false;
@@ -139,7 +146,7 @@ export const setup = asyncSetup([
     }
 
     if (options.filtersUpdatedAt < Date.now() - HOUR_IN_MS) {
-      updateEngines();
+      await updateEngines();
     }
   }),
   observe('experimentalFilters', async (value, lastValue) => {
--- a/src/background/helpers.js
+++ b/src/background/helpers.js
@@ -9,7 +9,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
-import { resolveObservers } from '/store/options.js';
+import { getIdleOptionsObservers } from '/store/options.js';
 
 chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
   switch (msg.action) {
@@ -33,19 +33,11 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
     case 'openPrivateWindowWithUrl':
       chrome.windows.create({ url: msg.url, incognito: true });
       break;
-    case 'idle': {
-      // The idle detection is a two-step process:
-      // Schedule delayed check to ensure that the background process has started
-      // running the tasks. Then resolve all options observers and send the response
-      // in the next delayed timeout.
-      setTimeout(() =>
-        resolveObservers().then(() => {
-          setTimeout(() => {
-            sendResponse();
-            console.info('[helpers] idle response...');
-          });
-        }),
-      );
+    case 'idleOptionsObservers': {
+      getIdleOptionsObservers().then(() => {
+        sendResponse();
+        console.info('[helpers] "idleOptionsObservers" response...');
+      });
 
       return true;
     }
--- a/src/background/helpers.js
+++ b/src/background/helpers.js
@@ -35,7 +35,7 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
       break;
     case 'idleOptionsObservers': {
       getIdleOptionsObservers().then(() => {
-        sendResponse();
+        sendResponse('done');
         console.info('[helpers] "idleOptionsObservers" response...');
       });
 
--- a/src/background/helpers.js
+++ b/src/background/helpers.js
@@ -9,6 +9,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
+import { resolveObservers } from '/store/options.js';
+
 chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
   switch (msg.action) {
     case 'getCurrentTab':
@@ -31,6 +33,22 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
     case 'openPrivateWindowWithUrl':
       chrome.windows.create({ url: msg.url, incognito: true });
       break;
+    case 'idle': {
+      // The idle detection is a two-step process:
+      // Schedule delayed check to ensure that the background process has started
+      // running the tasks. Then resolve all options observers and send the response
+      // in the next delayed timeout.
+      setTimeout(() =>
+        resolveObservers().then(() => {
+          setTimeout(() => {
+            sendResponse();
+            console.info('[helpers] idle response...');
+          });
+        }),
+      );
+
+      return true;
+    }
   }
 
   return false;
--- a/src/background/helpers.js
+++ b/src/background/helpers.js
@@ -9,7 +9,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
-import { getIdleOptionsObservers } from '/store/options.js';
+import { idleOptionsObservers } from '/store/options.js';
 
 chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
   switch (msg.action) {
@@ -33,8 +33,9 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
     case 'openPrivateWindowWithUrl':
       chrome.windows.create({ url: msg.url, incognito: true });
       break;
+    // This is used only by the e2e tests to detect idle state
     case 'idleOptionsObservers': {
-      getIdleOptionsObservers().then(() => {
+      idleOptionsObservers.then(() => {
         sendResponse('done');
         console.info('[helpers] "idleOptionsObservers" response...');
       });
--- a/src/pages/settings/components/checkbox.js
+++ b/src/pages/settings/components/checkbox.js
@@ -22,7 +22,7 @@ export default {
   disabled: { value: false, reflect: true },
   input: (host) => host.querySelector('input'),
   checked: {
-    value: (host, value) => value ?? host.input?.checked,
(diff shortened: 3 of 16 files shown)


tests/e2e/spec/advanced.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+describe('Advanced Features', function () {
+  before(enableExtension);
+
+  describe('Custom Filters', function () {
+    before(async function () {
+      await setCustomFilters(
+        `@@connect.facebook.net^` +
+          `${PAGE_DOMAIN}###custom-filter` +
+          `${PAGE_DOMAIN}##+js(rpnt, h1, Test Page, "Hello world")`,
```
chrmod, line 57: shouldn't those be separated by `\n`?

smalluban: Yes :)

tests/e2e/spec/advanced.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+  describe('Custom Filters', function () {
+    before(async function () {
+      await setCustomFilters([
+        `@@connect.facebook.net^`,
+        `${PAGE_DOMAIN}###custom-filter`,
+        `${PAGE_DOMAIN}##+js(rpnt, h1, Test Page, "Hello world")`,
+      ]);
+
+      await disableCache();
```
chrmod, line 60: Can we hide this away? Cannot we set it one time before all tests?

tests/e2e/wdio.conf.js
```diff
@@ -85,9 +86,9 @@ export function buildForChrome() {
 }
 
 export const config = {
-  specs: argv.debug ? [['**/*.spec.js']] : ['**/*.spec.js'],
+  specs: [['**/*.spec.js']],
   specFileRetries: 2,
```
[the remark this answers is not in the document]
smalluban: There are still bugs in Firefox implementation, and sometimes we have navigation errors (which are coused by the Firefox itself). I recommend to keep it. However, we can test it out, how it will go without it, and if it is required, we can bring it back.

tests/e2e/spec/advanced.spec.js
smalluban: I thought about it too, there is the`browser.overwriteCommand` API - we can add disabling cache on Chrome before we call the original `browser.url`.

tests/e2e/wdio.conf.js
```diff
@@ -149,6 +147,17 @@ export const config = {
       await browser.installAddOn(extension.toString('base64'), true);
     }
 
+    // Disable cache for Chrome to avoid caching issues for browser.url()
+    if (capabilities.browserName === 'chrome') {
+      browser.overwriteCommand('url', async function (orig, url, options) {
+        await browser.sendCommand('Network.setCacheDisabled', {
```
chrmod, line 153: that this have to be called for every navigation? I was hoping it can be done once for the browsing session

smalluban: Sadly, it works only for the "next" navigation event. I tested it when I found the solution, as I thought the same as you.

smalluban:
The last commit highlights: 

* Added `idle` action to detect when the background process ended all of the tasks
* Minor fixes to the engine update - we can start updating each engine parallel, just the merge must be done when all of them are updated - it should make the process much faster

src/background/adblocker.js
```diff
@@ -92,21 +92,28 @@ async function updateEngines() {
 
     if (enabledEngines.length) {
       let updated = false;
+
       // Update engines from the list of enabled engines
-      for (const id of enabledEngines) {
-        if (id === engines.CUSTOM_ENGINE) continue;
```
chrmod, line 97: Is this no longer needed?

src/background/adblocker.js
```diff
@@ -92,21 +92,28 @@ async function updateEngines() {
...
-        updated = (await engines.update(id).catch(() => false)) || updated;
-      }
+      await Promise.all(
```
chrmod, line 97: What's the reason to do it in parallel ? Wasn't it fine to do it one after other?

src/background/adblocker.js
```diff
@@ -92,21 +92,28 @@ async function updateEngines() {
...
 
+      // Update timestamp after the engines are updated
+      await store.set(Options, { filtersUpdatedAt: Date.now() });
+
       // Update TrackerDB engine
       trackerdb.setup.pending && (await trackerdb.setup.pending);
       await engines.update(engines.TRACKERDB_ENGINE).catch(() => null);
-
-      // Update timestamp after the engines are updated
-      await store.set(Options, { filtersUpdatedAt: Date.now() });
```
chrmod, line 109: With this we no longer know if trackerdb was updated

src/background/helpers.js
```diff
@@ -31,6 +33,22 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
...
       chrome.windows.create({ url: msg.url, incognito: true });
       break;
+    case 'idle': {
+      // The idle detection is a two-step process:
+      // Schedule delayed check to ensure that the background process has started
+      // running the tasks. Then resolve all options observers and send the response
+      // in the next delayed timeout.
+      setTimeout(() =>
+        resolveObservers().then(() => {
+          setTimeout(() => {
```
chrmod, line 43: setImmidiate?

src/background/helpers.js
```diff
@@ -31,6 +33,22 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
     case 'openPrivateWindowWithUrl':
       chrome.windows.create({ url: msg.url, incognito: true });
       break;
+    case 'idle': {
```
chrmod, line 36: please don't call it `idle` or background idle as this reflects state of options observers only and only from one instance of time. The name have should have something about `options` in it.

src/pages/settings/components/checkbox.js
```diff
@@ -11,26 +11,36 @@
...
+  input: (host) => host.querySelector('input'),
+  checked: {
+    value: (host) => host.input?.checked ?? false,
+    connect: (host, key, invalidate) => {
+      host.input?.addEventListener('change', invalidate);
+      return () => {
+        host.input?.removeEventListener('change', invalidate);
+      };
+    },
+  },
```
chrmod, line 32: why this is not a `onchange` property on the input? because the slot? It took me a while to understand what it does.

src/background/helpers.js
smalluban: https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate 
> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards...

The `requestIdleCallback` is not available in Service Workers - I tried it...

src/store/options.js
```diff
@@ -178,13 +179,15 @@ const Options = {
       // Sync if the current memory context get options for the first time
       if (!prevOptions) sync(options);
 
-      observers.forEach(async (fn) => {
-        try {
-          await fn(options, prevOptions);
-        } catch (e) {
-          console.error(`[options] Error while observing options: `, e);
+      resolvedObservers = (async () => {
```
chrmod, line 182: this looks like it may be super random. if there are multiple sources of option mutation a customer have to be lucky to call `resolveObservers` at the right time

src/store/options.js
```diff
@@ -392,6 +395,10 @@ export async function observe(...args) {
   };
 }
 
+export function resolveObservers() {
```
chrmod, line 398: can we use a different name for variable and for the function?

src/pages/settings/components/checkbox.js
smalluban: The component is a UI wrapper for the `<input>` element, and it is in the `<slot>` element.

src/store/options.js
smalluban: No, it's not. The background process has only one instance of options, and this method is called only once, after the options are updated.

src/background/adblocker.js
smalluban: I did not think about it in this way. You're right.

src/pages/settings/components/checkbox.js
chrmod: ok, I like this version more than `const target = host.render().querySelector('slot').assignedElements()[0];`

tests/e2e/spec/advanced.spec.js
```diff
@@ -0,0 +1,101 @@
...
+  describe('Custom Filters', function () {
+    before(async function () {
+      await setCustomFilters([
+        `@@connect.facebook.net^`,
+        `${PAGE_DOMAIN}###custom-filter`,
+        `${PAGE_DOMAIN}##+js(rpnt, h1, Test Page, "Hello world")`,
+      ]);
+    });
+
+    after(() => setPrivacyToggle('custom-filters', false));
```
chrmod, line 55: probably best to update custom filters with empty input before disabling. just in case this state wont come up and mess other tests

tests/e2e/spec/main.spec.js
```diff
@@ -64,106 +51,100 @@ describe('Main Features', function () {
...
+          await expect(
+            getExtensionElement(`icon:tracker:${trackerId}:modified`),
+          ).not.toBeDisplayed();
         }
       });
     });
 
     it('blocks tracker requests on the page', async function () {
-      await updatePrivacySettings('anti-tracking', true);
+      await setPrivacyToggle('anti-tracking', true);
```
chrmod, line 96: maybe disable all languages here?

tests/e2e/utils.js
```diff
@@ -52,6 +44,23 @@ export function getExtensionElement(id) {
   return $(`>>>[data-qa="${id}"]`);
 }
 
+export async function waitForIdleBackgroundTasks() {
+  const protocol = browser.isChromium ? 'chrome-extension' : 'moz-extension';
```
chrmod, line 48: wasn't protocol defined somewhere else already? in config?
Maybe it would be possible to attach some data to the browser object?

tests/e2e/utils.js
```diff
@@ -52,6 +44,23 @@ export function getExtensionElement(id) {
...
+
+  if (!(await browser.getUrl()).startsWith(protocol)) {
```
chrmod, line 50: maybe instead of throwing we ensure that this is executes in the right context. How about we open new browser window with extension page and close it after we get the response?

(trick with the window I think is needed only for Firefox)

tests/e2e/utils.js
```diff
@@ -52,6 +44,23 @@ export function getExtensionElement(id) {
...
+    throw new Error(
+      'Background idle state must be checked from the extension context',
+    );
+  }
+
+  // Wait for the 'idleOptionsObservers' response
+  await browser.execute(
+    browser.isChromium
+      ? () => chrome.runtime.sendMessage({ action: 'idleOptionsObservers' })
```
chrmod, line 59: I think in chrome you should be able to evaluate a code in the background. but maybe this is better for now as we will have same mechanism for both platforms (so hopefully will have less surprises)

tests/e2e/utils.js
```diff
@@ -52,6 +44,23 @@ export function getExtensionElement(id) {
...
+      : () => browser.runtime.sendMessage({ action: 'idleOptionsObservers' }),
```
chrmod, line 60: why cannot you call `chrome` on Firefox? is it still callback driven?

tests/e2e/wdio.conf.js
chrmod: I'm pretty sure it works for entire session. Maybe we do something wrong elsewhere?

tests/e2e/wdio.update.conf.js
```diff
@@ -138,10 +137,16 @@ export const config = {
           ),
         );
 
+        // Ensure extension reloaded the source, so we can open the settings page
+        await browser.pause(5000);
```
chrmod, line 141: maybe this can be replaced with some `browser.waitUntil` and some message passing? we can await response from background to be sure it is ready. Most of the time this should be faster than 5 seconds and in case it will be slower the tests will always pass. (but you may want to extend a default mocha timeout for this step)

tests/e2e/spec/main.spec.js
smalluban: I don't understand this one; it is a test for anti-tracking. There is no relation to languages here. Did you mean regional filters?

tests/e2e/utils.js
smalluban: Yes, but above, not in the config.

tests/e2e/utils.js
smalluban: This code might be called before onboarding, in the panel contexts, and how it can open url/window is related to that context - to simply not break tests (as jumping on windows/url makes test flaky). This is the reason why I did not add that into the function, but there is a check to be sure, if someone would use it on wrong context.

tests/e2e/utils.js
smalluban: If we would like to call some function in the background it must be available in global context - I think this pattern is clearer, as requires less changes on the background.

tests/e2e/utils.js
smalluban: There are cases where this code was executed even before the javascript code of the page is executed (as this code is ES module (deffer)). We can use `chrome`, but than is lotery, if you have "fixed" or "old" version

tests/e2e/spec/main.spec.js
smalluban: The config set the default language to `en-GB`, so the regional filters never have predefined regions (we don't have `en` region).

src/store/options.js
```diff
@@ -24,6 +24,11 @@ export const GLOBAL_PAUSE_ID = '<all_urls>';
 
 const observers = new Set();
 
+let idleOptionsObservers = Promise.resolve();
+export function getIdleOptionsObservers() {
+  return idleOptionsObservers;
+}
+
```
chrmod, line 31: please add a comment that is meant to be used by tests only

tests/e2e/spec/_onboarding.spec.js
```diff
@@ -14,32 +14,32 @@ import {
...
   it('keeps ghostery disabled', async function () {
+    await browser.url(await getExtensionPageURL('onboarding'));
     await getExtensionElement('button:skip').click();
     await expect(getExtensionElement('view:skip')).toBeDisplayed();
 
     await switchToPanel(async function () {
       await expect(getExtensionElement('button:enable')).toBeDisplayed();
     });
+
+    await browser.url('about:blank');
```
chrmod, line 30: why not in afterEach? is that helping with Firefox? if so, maybe add some readme for the tests to share that knowledge

tests/e2e/spec/main.spec.js
chrmod: Yes, I've meant the regional filters, I've put the comment on the wrong line - sorry.

smalluban:
With the latest commit tests passed 3 times in a row. Unfortunately, there must be a `browser.pause()` for Firefox in the update process, as otherwise we still can have failures in setup. I tried my best to avoid it, but the reload of the extension must influence the web driver, and one of two tries fails on the setup step, not the tests.

tests/e2e/spec/_onboarding.spec.js
smalluban: Yes, but not only for Firefox. The tests are set for 2 retries, but the before/after hooks do not retry. It is best to avoid them.
We can create README with tips, but I would do that after the test configuration and setup settle down.

chrmod (approved):

PR #1970, opened 2024-10-17
Ad-blocking: use new resources json format

chrmod:
Surpasses https://github.com/ghostery/ghostery-extension/pull/1968


chrmod (commented):
I've tested it locally and all seem to work well. Automated test for custom filters should ensure it is correct.

seia-soto (approved):

PR #1972, opened 2024-10-18
fix(adblocker): Clean up scriptlets and cosmetics execution

smalluban:
Highlights in the changes:
* `active` is always true, as I checked and this relies on the engine configuration - and we always will have scriptlets and cosmetics turned on - waste to deserialize argument and check it every time...
* I have no clue why we encode and decode string, which we passed by the API * it is not added to any URL. Do you know if there is a reason this is required?

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -254,7 +254,7 @@ async function injectCosmetics(msg, sender) {
   // ids and hrefs observed in the DOM. MutationObserver is also used to
   // make sure we can react to changes.
   {
-    const { active, styles } = engine.getCosmeticsFilters({
+    const { styles } = engine.getCosmeticsFilters({
       domain,
       hostname,
       url,
@@ -272,10 +272,6 @@ async function injectCosmetics(msg, sender) {
       // This will be done every time we get information about DOM mutation
       getRulesFromDOM: msg.lifecycle === 'dom-update',
     });
-
-    if (active === false) {
-      return;
-    }
 
     specificStyles.push(styles);
     specificFrameId = frameId;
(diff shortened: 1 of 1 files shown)


philipp-classen:
The reason for the URI encoding is because the scriptlet code is passed twice ("breaking the isolated word"):
1. extension to the content script
2. content script to the page

You need to be aware of encoding problems in both steps. URL encoding helps, because you then have a string that will not be running into quoting problems and can be passed across such boundaries.

smalluban:
I still wonder if encoding of the page can have an impact on defined string variables. The character might be displayed differently, but the char codes must be the same, aren't they?

smalluban:
I tried to find resources on that subject, but I couldn't. Can you provide any documentation, on which points that set string literal characters might change because of the page encoding?

philipp-classen:
So, you are asking basically if quoting is really needed? I think it is, but is difficult to come up with an example out of my head that would break otherwise.

The code that we use now is derived from this PR: https://github.com/ghostery/adblocker/pull/2750. The topic has been discussed also here, but in a slightly different context: https://github.com/ghostery/adblocker/pull/2750/files#diff-8b9d83cd2b834713818f1290efe43d08cdea910aae5968963cfa9545cc95da33R655

For instance, code with regular expressions like this here caused problems in the old implementation:
```
/[.*+?^${}()|[\]\\]/
```

In the new code base, it is slightly different; but if there are any modifications being done by the <script> element and how argument are passed through, it is quite likely that it is possible to find something that would then break.

A simpler example where it could make a different would something like that:
```
console.log('\\\\'.length);
```

Not sure if this would print the same results when passed through.

smalluban:
Ok, thanks for clarification. What about this?

```js
String.raw`Hi\u000A!`;
// 'Hi\\u000A!', same here, this time we will get the
// \, u, 0, 0, 0, A, 6 characters.
// All kinds of escape characters will be ineffective
// and backslashes will be present in the output string.
// You can confirm this by checking the .length property
// of the string.
```

smalluban:
Closing in favor of #1977

PR #1977, opened 2024-10-21
Ad-blocking: improve scriptlet injection

chrmod:
Surpasses: #1972 

Specifically at Safari < 18, scriptlets were not injected fast enough to block ads on some pages. This PR ensures that scritplet injection is sync. Additionally, on Safari we use two different triggering methods (content script message and webNavigation) to trigger the injection, but we ensure that scriptlets are injected only once. 

Related code was cleaned up from obsolete comments and dead paths.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -205,11 +205,9 @@ function adblockerInjectStylesWebExtension(
 }
 
 // copied from https://github.com/cliqz-oss/adblocker/blob/0bdff8559f1c19effe278b8982fb8b6c33c9c0ab/packages/adblocker-webextension/adblocker.ts#L297
-async function injectCosmetics(msg, sender) {
-  try {
-    setup.pending && (await setup.pending);
-  } catch (e) {
-    console.error(`[adblocker] Error while setup cosmetic filters: ${e}`);
+function injectCosmetics(msg, sender) {
+  if (setup.pending) {
+    console.warn(`[adblocker] not ready for cosmetic injection`);
     return;
   }
 
@@ -385,6 +383,11 @@ async function executeScriptlets(tabId, frameId, scripts) {
 }
 
 function injectScriptlets(tabId, frameId, url) {
+  if (setup.pending) {
+    console.warn('[adblocker] not ready for scriptlet injection');
+    return;
+  }
+
   const { hostname, domain } = parse(url);
   if (!hostname || isPaused(options, hostname)) {
     return;
@@ -467,9 +470,7 @@ if (__PLATFORM__ === 'firefox') {
       if (details.tabId < 0 || details.type === 'main_frame') return;
 
       if (setup.pending) {
-        console.error(
-          '[adblocker] Error while processing network request - adblocker not ready yet',
-        );
+        console.warn('[adblocker] not ready for network blocking');
         return;
       }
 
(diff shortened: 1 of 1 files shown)


src/background/adblocker.js
```diff
@@ -418,12 +410,12 @@ if (__PLATFORM__ === 'safari') {
 
     return false;
   });
-} else {
-  chrome.webNavigation.onCommitted.addListener(async (details) => {
-    injectScriptlets(details.tabId, details.frameId, details.url);
-  });
 }
 
+chrome.webNavigation.onCommitted.addListener((details) => {
```
chrmod, line 415: async removed

src/background/adblocker.js
```diff
@@ -367,14 +373,7 @@ ${scripts.join('\n\n')}}
   );
 }
 
-async function injectScriptlets(tabId, frameId, url) {
```
chrmod, line 370: async removed 

@smalluban wonder if we should not early exit if setup is still pending.

src/background/adblocker.js
```diff
@@ -310,26 +310,32 @@ chrome.runtime.onMessage.addListener((msg, sender) => {
...
-  const codeRunningInPage = `(function(){
-${debugMarker('run scriptlets (executing in "page world")')}
-${scripts.join('\n\n')}}
-)()`;
+  const scriptlets = `(function(){ ${scripts.join('\n\n')}} )();`;
 
-  // wrapper to break the "isolated world" so that the patching operates
-  // on the website, not on the content script's isolated environment.
-  function codeRunningInContentScript(code) {
+  function scritpletInjector(code, uuid) {
```
philipp-classen, line 327:
```suggestion
  function scriptletInjector(code, uuid) {
```

smalluban (changes requested):
The only logic change here is to add `onCommited` callback to Safari. From my testing on Safari 17.6, it did not help in any way - there was no difference between content script message listener and `webnavigation.onComitted`.

src/background/adblocker.js
```diff
@@ -367,14 +373,7 @@ ${scripts.join('\n\n')}}
   );
 }
 
-async function injectScriptlets(tabId, frameId, url) {
-  try {
-    setup.pending && (await setup.pending);
```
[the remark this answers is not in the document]
chrmod: Not super sure how the runloop will handle that across different engines. 
But to be sure we are as fast as possible, ad-blocking path should always be sync. I've changed this place and all others to print a warning instead of an error as it is expected that from time to time adblocker wont startup in time.

smalluban (approved):
If adding  an `async` listener for `webRequest` in Firefox is fine, the code looks good to me.

smalluban (approved):

src/background/adblocker.js
```diff
@@ -311,36 +303,38 @@ async function injectCosmetics(msg, sender) {
...
   if (msg.action === 'getCosmeticsFilters') {
-    injectCosmetics(msg, sender).catch((e) =>
-      console.error(
-        `[adblocker] Error while processing cosmetics filters: ${e}`,
-      ),
-    );
+    injectCosmetics(msg, sender);
   }
 
   return false;
```
philipp-classen, line 309: Only a detail since it has no changed. But in principle, we could also remove the "return false" here, couldn't we? The default of returning undefined, should have the same effect.

Issue #1985, opened 2024-10-22
Use Mozilla's addon-linter in the CI


chrmod:
https://github.com/mozilla/addons-linter

This tool is used by the AMO reviewer to help them detect potential problems with the extension. Recently they have noticed a potentially unsafe use of innerHTML which is not fixed by https://github.com/ghostery/ghostery-extension/commit/53ba99566ea7f0d5b7b6a9e8362d3063fdb63a4d

Hopefully we can also use addon-linter to learn and fix such issues ahead of time.

PR #1992, opened 2024-10-25
CI: Integrate addons-linter

chrmod:
fix https://github.com/ghostery/ghostery-extension/issues/1985


.github/workflows/test.yml
```diff
@@ -23,6 +23,11 @@ jobs:
       - name: Lint
         run: npm run lint
 
+      - name: Addons Linter
+        run: |
+          npm run build -- firefox
```
smalluban, line 28: My only concern is to run `npm run build` multiple times during this action, but as I propose to add the build step to the `addons-linter` command, it must be like this.

package.json
```diff
@@ -13,6 +13,7 @@
     "debug": "npm run build -- --watch --debug",
     "licenses": "license-report --config=./.license-report-config.json > dist/licenses.html",
     "lint": "eslint",
+    "addons-linter": "addons-linter dist",
```
smalluban, line 16: If the dist does not exist or it is not a Firefox build, the command will not work as expected. What do you think about the change, and also rename the command to follow `lint`:
```suggestion
    "lint:build": "npm run build firefox && addons-linter dist",
```

package.json
```diff
@@ -13,6 +13,7 @@
     "debug": "npm run build -- --watch --debug",
     "licenses": "license-report --config=./.license-report-config.json > dist/licenses.html",
     "lint": "eslint",
```
smalluban, line 15: We can change that to be `lint:src` and create a new `lint` command, which does both - so the test GitHub Action would not change at all.

smalluban (changes requested):

package.json
```diff
@@ -31,7 +33,9 @@
     "@wdio/globals": "^9.1.2",
     "@wdio/mocha-framework": "^9.1.2",
     "@wdio/spec-reporter": "^9.1.3",
+    "addons-linter": "^7.3.0",
     "eslint": "^9.13.0",
+    "eslint-config-prettier": "^9.1.0",
```
chrmod, line 38: while resolving merge conflicts had to remove package-lock, which exposed that `eslint-config-prettier` is a peer dependency of `eslint-plugin-prettier` and has to be explicitly listed

smalluban (approved):

PR #1994, opened 2024-10-25
fix: update platforms manifest configuration

smalluban:
* Merges Safari two manifests two one - they have been identical for a while now.
* Updates Safari manifest to use `manifest_version: 3` - the most important benefit is more precise `web_acessible_resources` syntax, background page based on scripts lists, and unifiaction of some properties.
* Moves out entry points from the `web_acessible_resources` if they should not be there. Instead, the content scripts are defined in one place, but those that are triggered dynamically by the background page are flagged with custom `run_at: 'background_execue_script'`, and removed from the manifest on the build step
* Adds `use_dynamic_urls` to the resources. It also adds the required change to the `getURL` calls, as the current Chrome version has a bug with that feature


scripts/build.js
```diff
@@ -207,8 +202,8 @@ if (manifest.declarative_net_request?.rule_resources) {
       resources: redirectResources.map((filename) =>
         join('rule_resources/redirects', filename),
       ),
-      all_frames: true,
```
smalluban, line 210: There is no `all_frames` key in the `web_accessible_resources` (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/web_accessible_resources)

smalluban:
@chrmod I tested, and `webRequest` is available in both macOS and iOS in Safari 18. It does not break the manifest in Safari 17 on iOS or macOS (it just doesn't add that API). This is important as separate manifests won't be needed because of that API.

However, the `webRequest` behaves differently in Safari than in Chrome - `onBeforeRequest` has fewer keys in details, `onBeforeRedirect` does not trigger for our redirects, the `onErrorOccurred` does not call at all, etc... I will investigate the possibility of using it, but it must be done in separate PR.

Please review the PR as it is now.

chrmod (approved):

Issue #1995, opened 2024-10-28
Safari returns "about:blank" as "runtime.MessageSender.url" on website preload


seia-soto:
Safari may start preloading website before an user decides a visit to the URL. This may lead to falsy sender information: `runtime.MessageSender` and we need to fallback a value of "sender.url" to "sender.tab.url"

refs https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/MessageSender
refs https://github.com/ghostery/broken-page-reports/issues/935
refs https://github.com/ghostery/adblocker/issues/4396

The following is a sampled `sender` object on preload:

```json
{
  "id": "local.beretto.ghostery-extension",
  "tab": {
    "id": 739,
    "index": 0,
    "active": true,
    "width": 1815,
    "audible": false,
    "url": "https://www.naver.com/",
    "mutedInfo": {
      "muted": false
    },
    "windowId": 663,
    "title": "NAVER",
    "incognito": false,
    "pinned": false,
    "height": 1174,
    "highlighted": true,
    "isArticle": false,
    "isInReaderMode": false,
    "selected": true,
    "status": "complete"
  },
  "frameId": 227,
  "url": "about:blank",
  "origin": "null"
}
```

You can reproduce this by adding the following to `injectCosmetics` in `src/background/adblocker.js` and visiting any website through start page of Safari.

```js
if (url === 'about:blank' || !url) {
  console.log(JSON.stringify(sender, null, 2));
  debugger;
}
```

```
diff --git a/src/background/adblocker.js b/src/background/adblocker.js
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -209,4 +209,9 @@
   const domain = parsed.domain || '';
 
+  if (url === 'about:blank' || !url) {
+    console.log(JSON.stringify(sender, null, 2));
+    debugger;
+  }
+
   if (!sender.tab || isPaused(options, hostname)) {
     return;
```

When this problem occurs, our adblock is not able to process cosmetic exceptions properly. See the related issue on ghostery/adblocker: https://github.com/ghostery/adblocker/issues/4396 for details.

seia-soto:
Otherwise, you may visit `koreanair.com` to check a real-world issue. Put `koreanair.com#@#div.cookie-banner` in custom filters and see if the site becomes scrollable.

PR #1996, opened 2024-10-28
fix: fallback `sender.url` on Safari

seia-soto:
fixes https://github.com/ghostery/ghostery-extension/issues/1995
refs https://github.com/ghostery/broken-page-reports/issues/935
refs https://github.com/ghostery/adblocker/issues/4396

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -214,7 +214,14 @@ async function injectCosmetics(msg, sender) {
   }
 
   // Extract hostname from sender's URL
-  const { url = '', frameId } = sender;
+  const { frameId } = sender;
+  let url = sender.url;
+
+  // Try `sender.tab.url` as a fallback when Safari preloads a website
+  if (!url || url === 'about:blank') {
+    url = sender.tab.url;
+  }
+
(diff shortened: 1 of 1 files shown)


src/background/adblocker.js
```diff
@@ -214,7 +214,14 @@ async function injectCosmetics(msg, sender) {
   }
 
   // Extract hostname from sender's URL
-  const { url = '', frameId } = sender;
+  const { frameId } = sender;
+  let url = sender.url;
+
+  // Try `sender.tab.url` as a fallback when Safari preloads a website
+  if (!url || url === 'about:blank') {
+    url = sender.tab.url;
```
smalluban, line 222: Can't we just always use `sender.tab.url`?

seia-soto: MDN docs specifies that `sender.tab` can be `undefined`, I don't know if there's a case without `sender.tab`.

seia-soto:
Now falling back to `sender.url` unless `sender.tab.url` is available.

smalluban (approved):

PR #2003, opened 2024-10-29
fix(options): decouple observe from options

smalluban:
I have found a few minor issues with the `observe` pattern:
* The `observe(fn)` function runs `fn` immediately after initialization and always resolves (the error is caught). For example, if the `adblocker` engines don't start correctly, the `asyncSetup` will still resolve
* As the `fn` is called immediately, the idle mechanism cannot work with an initial run of the background process, it only watches the next changes
* `errors.js` is attached to different contexts, so the observer can be changed to fetch options when needed
*  Telemetry uses `observe` in async call, so it is attached later

The solution is to decouple the `observe` mechanism and make the initial call of all observers in the same mechanism as an update of options. Also, the `asyncSetup` was exceeded to 10 seconds, but here it is just my assumption, that on some slow devices, the first merge of all engines after onboarding might be slow.


src/utils/errors.js
```diff
@@ -44,21 +45,20 @@ getBrowserInfo().then(
   () => {},
 );
 
-let terms = false;
-observe('terms', (value) => {
-  terms = value;
-});
+export async function captureException(error) {
+  const { terms } = await store.resolve(Options);
```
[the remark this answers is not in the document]
smalluban: When using observe pattern, it adds a function call to queue (or create a queue if used outside background), which is rerly used. 

If for any reason the previous pattern would fail in setting terms, we would not send error reports too. I think the situation is the same.

src/store/options.js
```diff
@@ -174,35 +169,34 @@ const Options = {
           // sendMessage may fail without potential target
         });
 
-      sync(options, keys).catch(() => null);
+      sync(options, keys);
```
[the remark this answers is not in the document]
smalluban: This was redundant. The `sync` has try/catch already with logging. In other places we do not add explicit catch() call.

src/background/custom-filters.js
```diff
@@ -209,29 +210,34 @@ async function update(text, { trustedScriptlets }) {
...
 
-observe('customFilters', async ({ enabled, trustedScriptlets }, lastValue) => {
-  // Background startup
-  if (!lastValue) {
-    // If custom filters are disabled, we don't care if engine was reloaded
-    // as custom filters should be empty
-    if (!enabled) return;
+observer.addListener(
+  'customFilters',
+  async ({ enabled, trustedScriptlets }, lastValue) => {
```
chrmod, line 215: maybe you can deconstruct the first argument in a function body - that way we may fit the line length limit and the rest of the function will require no indentation changes.

src/background/dnr.js
```diff
@@ -20,7 +21,7 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
   // Ensure that DNR rulesets are equal to those from options.
   // eg. when web extension updates, the rulesets are reset
   // to the value from the manifest.
-  observe(async (options) => {
+  observer.addListener(async function dnr(options) {
```
chrmod, line 24: why not the arrow function like in other places?

src/background/telemetry/index.js
```diff
@@ -76,52 +63,56 @@ const getConf = async (storage) => {
...
+observer.addListener('terms', async function telemetry(terms) {
+  enabled = terms;
 
-  try {
-    await recordUTMs(telemetry, JUST_INSTALLED);
-  } catch (error) {
-    log('Telemetry recordUTMs() error', error);
+  if (terms) {
+    setup.pending && (await setup.pending);
+    metrics.ping('install');
```
chrmod, line 106: should this have a condition ?

src/background/telemetry/index.js
```diff
@@ -76,52 +63,56 @@ const getConf = async (storage) => {
...
-  });
-
-  telemetry.setUninstallUrl();
-})();
+chrome.runtime.onMessage.addListener((msg) => {
+  if (enabled && msg.action === 'telemetry') {
+    (async () => {
+      setup.pending && (await setup.pending);
+      metrics.ping(msg.event);
+    })();
```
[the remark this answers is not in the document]
chrmod: imo it just more readable. runloop differences does not matter here afaict

src/store/options.js
```diff
@@ -174,35 +169,34 @@ const Options = {
...
+
 chrome.runtime.onMessage.addListener((msg) => {
   if (msg.action === UPDATE_OPTIONS_ACTION_NAME) {
     store.clear(Options, false);
     store.get(Options);
   }
 });
 
-export default Options;
+export function isPaused(options, domain = '') {
```
chrmod, line 194: Can we please do not move functions that are not affected by the PR? Keeping them in the same files makes reviewing easier and keeps git history meaningful.

src/utils/observer.js
```diff
@@ -0,0 +1,92 @@
+/**
```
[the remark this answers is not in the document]
chrmod: no strong opinion either as this is a first observer. but yes, calling it options observer would be easier to comprehend at the first glance

src/utils/observer.js
```diff
@@ -0,0 +1,92 @@
...
+    observers.add(wrapper);
+  });
+}
+
+let queue = null;
+export function run(options, prevOptions) {
+  if (observers.size === 0) return;
+
+  queue ||= (async () => {
+    console.group(`[observer] Run options observers...`);
```
chrmod, line 73: does it make sense to use console.group in async function? what if a console from a different scope will be call while this function is still running?

src/utils/observer.js
```diff
@@ -0,0 +1,92 @@
...
+
+    observers.add(wrapper);
+  });
+}
+
+let queue = null;
+export function run(options, prevOptions) {
+  if (observers.size === 0) return;
+
+  queue ||= (async () => {
```
chrmod, line 72: should you somehow "append" new "run" into the previous one if it hasn't finished yet? only then this is an actual queue

chrmod (changes requested):

src/background/dnr.js
smalluban: I added logs for each observer, and it uses the property name or function name to log, which observer is running.

src/background/telemetry/index.js
smalluban: No, I checked and the Metrics has the same: https://github.com/ghostery/ghostery-extension/blob/3ef12a5fb5dfd67c3de48f0f8f2750009ebe7af1/src/background/telemetry/metrics.js#L560-L561

src/store/options.js
smalluban: I completely reorganized the file, and the final structure should be like this. It is pointless to make another PR with moving lines around one file.

src/utils/observer.js
smalluban: This is the only possible drawback. However, I added logs for each observer, which runs (there is a change) - so there is much more logs now. It looks much cleaner when grouped. Please try it out. If you have better idea, you're welcome ;)

src/background/telemetry/index.js
chrmod: well, maybe we can move it outside of the metric, now it looks like we want to send install ping on every startup - which is not true

smalluban: It looks, like it looks, after all, if you don't change metrics, it will be checked twice. And I would keep the guard of internal data of Metrics inside of metrics...

src/utils/observer.js
smalluban: I added it. I tested, and "super fast" clicking on some long executing options appends changes, and idle is executed after the latest change.

src/background/adblocker.js
```diff
@@ -124,41 +125,46 @@ async function updateEngines() {
...
-    }
-  }),
+      if (options.filtersUpdatedAt < Date.now() - HOUR_IN_MS) {
+        await updateEngines();
+      }
+    },
+  ),
+  OptionsObserver.addListener(
+    'experimentalFilters',
+    async (value, lastValue) => {
```
chrmod, line 159: add a function name

src/background/telemetry/index.js
chrmod: previous version had a check, new one does not.

src/store/options.js
chrmod: it's just a function order, why it makes a difference? is it alphabetical now?

src/utils/options-observer.js
```diff
@@ -0,0 +1,93 @@
...
+    observers.add(wrapper);
+  });
+}
+
+let queues = new Set();
+export async function waitForIdle() {
+  for (const queue of queues) await queue;
+}
+
+export async function execute(options, prevOptions) {
```
chrmod, line 74: now that the name of the module is OptionObserver, maybe we can call this function `update` or `trigger`?

chrmod (changes requested):

chrmod (approved):

src/store/options.js
smalluban: The export is made just after the definition of the model. It keeps things more connected.

src/background/adblocker.js
smalluban: It would be the same as the property name, and the function name is only used when there is no property. This is only for logging, not to make more understandable code.

PR #2033, opened 2024-11-06
fix(adblocker): early execution of adblocker cosmetics

smalluban:
## Background of the changes

Currently the `dist/content_scripts/adblocker.js` has 736 lines of code, which over 90% is not used.

We do the following logic at the start-up of the content script (implicitly by calling `injectCosmetics()` from `'@ghostery/adblocker-webextension-cosmetics'`):

```js
getCosmeticsFilters({ lifecycle: 'start', ids: [], classes: [], hrefs: [] }).then((response) => 
handleResponseFromBackground(window, response));
```

Our code always responds with an empty object `{}` so the whole logic behind `handleResponseFromBackground` is not needed.

## New execution logic

Because the first initial call for cosmetics is made without any information from the page, we can call for them upfront using `onCommitted` listener. Also, we can merge calls for scriptlets and styles to only one `engine.getCosmeticFilters()`, and use both results (styles and scripts) for injection.

## Listen to DOM updates

The logic from `adblocker-webextension-cosmetics'` was rewritten as close as possible to the original logic. Still, as the `DOMMonitor` is for a general usage, there is a lot of logic, which can be simplified. After all, the idea is to catch all of the "new" classes, ids, and hrefs, that are added to the page.

## Safari `onCommitted` bug

I carefully checked if the bug still exists, and after testing on macOS (18.0.1), iOS - iPad 17.5 (simulator), iPhone 17.5 (physical device) - I can prove that opening page from bookmarks or pasting the URL in the new tab executes `onCommitted` callback correctly. I cannot reproduce a situation where the listener is not called...

If the above is right, we can remove from Safari the need for:
* Additional call to the BG from the content script
* A race condition for executing scriptlets

## Injecting styles

The injection of styles can be simplified to use only the `chrome.scripting.insertCSS` - all supported browsers have that API.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -167,158 +167,11 @@ export const setup = asyncSetup([
   ),
 ]);
 
-function adblockerInjectStylesWebExtension(
-  styles,
-  { tabId, frameId, allFrames = false },
-) {
-  // Abort if stylesheet is empty.
-  if (styles.length === 0) {
-    return;
-  }
+/*
+ * Cosmetics injection
+ */
 
-  if (chrome.scripting && chrome.scripting.insertCSS) {
-    const target = {
-      tabId,
-    };
-
-    if (frameId) {
-      target.frameIds = [frameId];
-    } else {
-      target.allFrames = allFrames;
-    }
-    chrome.scripting
-      .insertCSS({
-        css: styles,
-        origin: 'USER',
-        target,
-      })
-      .catch((e) => console.warn('[adblocker] failed to inject CSS', e));
-  } else {
-    const details = {
-      allFrames,
-      code: styles,
-      cssOrigin: 'user',
-      matchAboutBlank: true,
-      runAt: 'document_start',
-    };
-    if (frameId) {
-      details.frameId = frameId;
-    }
-    chrome.tabs
-      .insertCSS(tabId, details)
-      .catch((e) => console.warn('[adblocker] failed to inject CSS', e));
-  }
-}
-
-// copied from https://github.com/cliqz-oss/adblocker/blob/0bdff8559f1c19effe278b8982fb8b6c33c9c0ab/packages/adblocker-webextension/adblocker.ts#L297
-async function injectCosmetics(msg, sender) {
-  try {
-    setup.pending && (await setup.pending);
-  } catch (e) {
-    console.error('[adblocker] not ready for cosmetic injection', e);
-    return;
-  }
-
-  // Extract hostname from sender's URL
-  const { frameId } = sender;
-  const url = sender.tab?.url || sender.url;
-  const parsed = parse(url);
-  const hostname = parsed.hostname || '';
-  const domain = parsed.domain || '';
-
-  if (!sender.tab || isPaused(options, hostname)) {
-    return;
-  }
-
-  const genericStyles = [];
-  const specificStyles = [];
-  let specificFrameId = null;
-
-  const engine = engines.get(engines.MAIN_ENGINE);
-
-  // Once per tab/page load we inject base stylesheets. These are always
-  // the same for all frames of a given page because they do not depend on
-  // a particular domain and cannot be cancelled using unhide rules.
-  // Because of this, we specify `allFrames: true` when injecting them so
-  // that we do not need to perform this operation for sub-frames.
-  if (frameId === 0 && msg.lifecycle === 'start') {
-    const { styles } = engine.getCosmeticsFilters({
-      domain,
-      hostname,
-      url,
-
-      classes: msg.classes,
-      hrefs: msg.hrefs,
-      ids: msg.ids,
-
-      // This needs to be done only once per tab
-      getBaseRules: true,
-      getInjectionRules: false,
-      getExtendedRules: false,
-      getRulesFromDOM: false,
-      getRulesFromHostname: false,
-    });
-
-    genericStyles.push(styles);
-  }
-
-  // Separately, requests cosmetics which depend on the page it self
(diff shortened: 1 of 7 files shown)


src/content_scripts/adblocker.js
```diff
@@ -9,15 +9,124 @@
...
+  };
+
+  const run = () => {
+    clear();
+    fn();
+  };
 
-  return Promise.resolve({});
+  return () => {
+    if (maxWaitTimer === undefined) {
```
philipp-classen, line 29: Why change the original implementation (omitting the `maxWait > 0` check)?

https://github.com/ghostery/adblocker/blob/a7d149e36bb7dc748b3081eeb8fc14858210164e/packages/adblocker-content/src/index.ts#L62

tests/e2e/spec/main.spec.js
```diff
@@ -48,8 +48,9 @@ describe('Main Features', function () {
     });
   });
 
-  describe('Ad-Blocking', function () {
+  describe.only('Ad-Blocking', function () {
```
philipp-classen, line 51:
```suggestion
  describe('Ad-Blocking', function () {
```

smalluban: Oh.. good catch

src/content_scripts/adblocker.js
philipp-classen: IHMO, this change makes the function worse. Before, maxWait was an optional parameter, now it is is not (it will do setTimeout(..., undefined), which fires directly.

smalluban: The logic is exactly the same, the `maxWait` is always grether than 0...

```js
      if (maxWait > 0 && maxWaitTimer === undefined) {
```

src/content_scripts/adblocker.js
```diff
@@ -9,15 +9,124 @@
...
+        ids: Array.from(selectors.ids),
+        hrefs: Array.from(selectors.hrefs),
+      });
+
+      selectors.classes.clear();
+      selectors.ids.clear();
+      selectors.hrefs.clear();
+    }
+  },
+  { waitFor: 25, maxWait: 1000 },
```
smalluban, line 75: @philipp-classen Here you have maxWait set to 1000.

src/content_scripts/adblocker.js
philipp-classen: Note that "debounce" implementation normally has no mandatory maxWait, so we are deviating from the standard interface by making it mandatory. The existing implementation would also support calls like:

```
debounce(fn, { waitFor: 25  });
```

I'm aware that in our current heuristic, we do pass a `maxWait`. (When we experimented, we originally did not have it. I was in favor of adding such an enforcement, so this is where it came from. However, even for me, it is not clear if it is better or worse to enforce every one second if there were pending changes recently.)

Well, I'm still not convinced that optimizing here makes sense. The old implementation was fine in my opinion. Code also get reused and requirements change. But if you feel it is making it more readable by making it a required parameter, we can change the interface.

src/background/adblocker.js
```diff
@@ -379,56 +221,119 @@ async function executeScriptlets(tabId, frameId, scripts) {
...
+
+    injectStyles(styles, tabId);
+  }
 }
 
-chrome.webNavigation.onCommitted.addListener((details) => {
-  injectScriptlets(details.tabId, details.frameId, details.url);
+chrome.webNavigation.onCommitted.addListener(
+  (details) => {
+    injectCosmetics(details, { bootstrap: true });
```
seia-soto, line 315: This is completely optional suggestion but is there any reason to wrap this with an object? Can we just pass `true` instead of `{ bootstrap: true }`? I don't think we need to care about extensibility here.

seia-soto (changes requested):
Looks like clean and performant, just addressing edge cases and chores.

src/background/adblocker.js
smalluban: It takes more parameters, look at the https://github.com/ghostery/ghostery-extension/blob/3b70ef64c69ac77d23e6870472482ef06e9ffa7b/src/background/adblocker.js#L272

src/content_scripts/adblocker.js
```diff
@@ -9,15 +9,124 @@
...
+const observer = new MutationObserver((mutations) => {
+  const visited = new Set();
+
+  for (const mutation of mutations) {
+    switch (mutation.type) {
+      case 'attributes': {
+        if (mutation.attributeName === 'class') {
+          mutation.target.classList.forEach((c) => addSelector('classes', c));
+        } else if (mutation.attributeName === 'id') {
+          addSelector('ids', mutation.target.id);
```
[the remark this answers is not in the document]
smalluban: Thanks! I will change that.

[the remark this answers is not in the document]
smalluban: This is the current implementation on `main`:

```js
// Update ids
                const id = element.getAttribute('id');
                if (typeof id === 'string') {
                    ids.add(id);
                }
                // Update classes
                const classList = element.classList;
                for (const class entry of classList) {
                    classes.add(class entry);
                }
                // Update href
                const href = element.getAttribute('here);
                if (type of href === 'string') {
                    refs.add(href);
                }
```

So, I will keep the classes as they are. The href and id will come from the attribute.

src/background/adblocker.js
seia-soto: Oh, I didn't notice those are coming from `config`. Then all good, thanks for the explanation 👍 .

seia-soto (approved):

PR #2034, opened 2024-11-06
fix(adblocker): engines update process clean up

smalluban:
It must be somehow a race condition, as rarely I found a state, where engines were not correctly written to the `IndexedDB`. Then, when engines are updated, the following message is logged:

> [engines] Skipping update for engine "${name}" as the engine is not available

I could fix the issue (by keeping the running extension with the problem) by adding `engines.init()` to the update process. It ensures that the engine is correctly initialized before we try to update it.

However, I can't reproduce the issue - it is kind of random.

Also, I added a delay to reloading main engine to avoid UI freezing and making to much changes to it.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -54,31 +54,42 @@ function getEnabledEngines(config) {
   return [];
 }
 
-async function reloadMainEngine() {
-  const enabledEngines = getEnabledEngines(options);
+function reloadMainEngine() {
+  clearTimeout(reloadMainEngine.timeout);
 
-  if (enabledEngines.length) {
-    engines.replace(
-      engines.MAIN_ENGINE,
-      (
-        await Promise.all(
-          enabledEngines.map((id) =>
-            engines.init(id).catch(() => {
-              console.error(`[adblocker] failed to load engine: ${id}`);
-              return null;
-            }),
-          ),
-        )
-      ).filter((engine) => engine),
-    );
+  // Debounce the reloading of the main engine to avoid multiple reloads
+  // when the user changes multiple options in a short period of time
+  reloadMainEngine.timeout = setTimeout(
+    async () => {
+      const enabledEngines = getEnabledEngines(options);
 
-    console.info(
-      `[adblocker] Main engine reloaded with: ${enabledEngines.join(', ')}`,
-    );
-  } else {
-    engines.create(engines.MAIN_ENGINE);
-    console.info('[adblocker] Main engine reloaded with no filters');
-  }
+      if (enabledEngines.length) {
+        engines.replace(
+          engines.MAIN_ENGINE,
+          (
+            await Promise.all(
+              enabledEngines.map((id) =>
+                engines.init(id).catch(() => {
+                  console.error(`[adblocker] failed to load engine: ${id}`);
+                  return null;
+                }),
+              ),
+            )
+          ).filter((engine) => engine),
+        );
+
+        console.info(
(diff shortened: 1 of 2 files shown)


src/background/adblocker.js
```diff
@@ -54,31 +54,42 @@ function getEnabledEngines(config) {
...
-          ),
-        )
-      ).filter((engine) => engine),
-    );
+function reloadMainEngine() {
+  clearTimeout(reloadMainEngine.timeout);
+
+  // Debounce the reloading of the main engine to avoid multiple reloads
+  // when the user changes multiple options in a short period of time
+  reloadMainEngine.timeout = setTimeout(
```
philipp-classen, line 62: Perhaps this could be a use case for requestIdleCallback (unfortunately not ready for Safari, but you can still fallback to setTimeout):

https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback

You can specify a timeout like the two seconds here, so you still have an upper bound. It will not batch multiple updates together, by default. So, it is not a drop-in replacement. But it could help to let the browser figure out an idle period to run the CPU bound computation.

smalluban: Service workers don't support it. Also, there is no problem with idle on the background side, it is on the web page context, which in Chrome is separated. 

The other thing is to denounce changes, as a user can click fast enough on option toggles, so main engine must be reloaded each time.

src/background/adblocker.js
```diff
@@ -98,25 +105,26 @@ async function updateEngines() {
...
-
       // Update TrackerDB engine
       trackerdb.setup.pending && (await trackerdb.setup.pending);
       await engines.update(engines.TRACKERDB_ENGINE).catch(() => null);
 
       // Update timestamp after the engines are updated
       await store.set(Options, { filtersUpdatedAt: Date.now() });
+
+      // Reload the main engine after all engines are updated
+      if (updated) await reloadMainEngine();
```
smalluban, line 127: The reload can happen for many reasons, but the timing should only be connected to the update process. Because of that, I moved the `reloadMainEngine()` call after setting a new date. More importantly, if for some reason the `reloadMainEngine` would throw, we will not end with an endless loop of updates.

src/background/adblocker.js
```diff
@@ -98,25 +105,26 @@ async function updateEngines() {
...
-              () => {},
-            ),
-          ),
+          .map(async (id) => {
+            try {
+              await engines.init(id);
+              const updatedEngine = await engines.update(id);
+              updated ||= updatedEngine;
+            } catch {
+              // Ignore errors
```
philipp-classen, line 114: Could we log the error? Looks relevant if it fails.

smalluban: It would be redundant, as the update function already logs errors, but also rethrows, so we have to catch it here. 

The logic here did not change, I only added engine.init(), and it was more convenient to change structure to async. Before we had a promise with an empty catch function.

seia-soto (approved):

src/background/adblocker.js
```diff
@@ -140,17 +142,13 @@ export const setup = asyncSetup([
           (enabledEngines.length !== prevEnabledEngines.length ||
             enabledEngines.some((id, i) => id !== prevEnabledEngines[i])))
       ) {
-        // The regional filters engine is no longer used, so we must remove it
```
smalluban, line 143: This code has been here since `10.4.5`, and we have not removed it yet because of the Firefox delay. As the `10.4.10` is this code will be on production in Firefox after at least 2-3 weeks - it is the best time to clear it up.

PR #2048, opened 2024-11-12
Update wtm/reporting

chrmod:


smalluban (approved):
LGTM. As the `terms` are set to `true` only one-way, unloading can be safely removed from the code.

philipp-classen (approved):

PR #2052, opened 2024-11-13
fix(options): clean up options migration logic

smalluban:
This PR prepares options for the sync feature refactor (move to the background process).

The current stats show, that all of the users migrated to v10. The last month's DAU of v8 is less than 5 (five users). 

Also, the options migration logic can be moved to a separate function, still avoiding another `await`. I figurated out that we can keep the condition in the `get()` method, but the migration body moves to a separate function.

### DIFF
--- a/src/store/options.js
+++ b/src/store/options.js
@@ -10,7 +10,6 @@
  */
 
 import { store } from 'hybrids';
-import { deleteDB } from 'idb';
 
 import { getUserOptions, setUserOptions } from '/utils/api.js';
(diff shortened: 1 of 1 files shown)


src/store/options.js
```diff
@@ -196,8 +160,44 @@ chrome.runtime.onMessage.addListener((msg) => {
   }
 });
 
+async function migrate(options, optionsVersion) {
+  const keys = [];
```
smalluban, line 164: The logic is untouched - copy/paste from the previous L118 to L150.

chrmod (approved):

PR #2057, opened 2024-11-14
chore(deps): set weekly update schedule

smalluban:
We are flooded with compensating updates, like `tldsts`, which updates very often, but to production we deliver only the latest change before the release. I think there is no point to update deps that often. 

If there are internal deps, we can always update them by hand (like adblocker or reporting).

The weekly update should fit well with our current weekly release process.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,7 @@ updates:
       - dependency-name: 'tldts-experimental'
     schedule:
       interval: 'daily'
+      time: '07:00'
     reviewers:
       - 'smalluban'
       - 'chrmod'
@@ -25,6 +26,7 @@ updates:
       - dependency-name: 'tldts-experimental'
     schedule:
       interval: 'weekly'
+      time: '07:00'
     reviewers:
       - 'smalluban'
     groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,23 +1,23 @@
 version: 2
 updates:
   - package-ecosystem: npm
-    directory: "/"
+    directory: '/'
     schedule:
-      interval: daily
-      time: "04:00"
+      interval: weekly
```
chrmod, line 6: please keep it daily for dependencies and maybe move to weekly for dev dependencies

chrmod (changes requested):

.github/dependabot.yml
```diff
@@ -1,23 +1,46 @@
...
+    directory: '/'
+    allow:
+      - dependency-type: 'production'
+    ignore:
+      - dependency-name: 'tldts-experimental'
     schedule:
-      interval: daily
-      time: "04:00"
+      interval: 'daily'
+      time: '10:00'
```
chrmod, line 11: please run it earlier so it is done when we start to work

chrmod (approved):

PR #2062, opened 2024-11-14
fix(sync): move options sync to background process

smalluban:
This PR moves out the sync of options into the background process.

The trigger of the sync process programmatically stays as it was - every time the background starts up the sync is triggered. However, if any other context wants to force the sync, the message `syncOptions` must be sent (added to the settings page and panel).

The `background/session.js` can be cleaned up from alarms. As the sync is triggered with every background process start-up, the alarms don't give much, as any request made by the browser brings the BG back to life. For the same reason, it is safe to trigger sync then, as from my testing, BG is not killed that often (many sites make some background calls, or the user moves the mouse over the page or something similar).

Also, the sync option in the Account section is unlocked. I found a use case, where you can log in, switch sync to false, and then log out, keeping the sync set to `false`. On the other hand, currently, it is impossible to log in without sync turned off. It is pointless to block this setting before the user is logged in, as it might be "hacked" anyway.

After code review, the PR requires manual testing.

### DIFF
--- a/src/pages/settings/index.js
+++ b/src/pages/settings/index.js
@@ -36,6 +36,9 @@ store
       };
     }
 
+    // Sync options with background
+    chrome.runtime.sendMessage({ action: 'syncOptions' });
+
     mount(document.body, Settings);
   })
   .catch(() => {
--- a/src/utils/debounce.js
+++ b/src/utils/debounce.js
@@ -1,0 +1,40 @@
+/**
+ * Ghostery Browser Extension
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+export default function debounce(fn, { waitFor, maxWait }) {
+  let delayedTimer;
+  let maxWaitTimer;
+
+  const clear = () => {
+    clearTimeout(delayedTimer);
+    clearTimeout(maxWaitTimer);
+    delayedTimer = undefined;
+    maxWaitTimer = undefined;
+  };
+
+  let args = [];
+  const run = () => {
+    clear();
+    fn(...args);
+
+    args = [];
+  };
+
+  return (...latestArgs) => {
+    args = latestArgs;
+
+    if (maxWait > 0 && maxWaitTimer === undefined) {
+      maxWaitTimer = setTimeout(run, maxWait);
+    }
+    clearTimeout(delayedTimer);
+    delayedTimer = setTimeout(run, waitFor);
+  };
+}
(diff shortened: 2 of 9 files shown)


smalluban:
@AdamGhst, I did my testing between Firefox and Chrome, but please do a round of smokes related to the sync process.

src/pages/settings/index.js
```diff
@@ -36,6 +36,9 @@ store
       };
     }
 
+    // Sync options with background
+    chrome.runtime.sendMessage({ action: 'syncOptions' });
```
chrmod, line 40: should we await here? we don't want UI to flicker in case option change

smalluban: We already discussed awaiting the sync process in another context, and our decision was to not wait for it. It can be long or failing for many reasons (bad internet connection etc) - The settings page are available immediately for the user (they are local).

AdamGhst (changes requested):
Setting Sync stops working. 

Steps to reproduce: 
1. Login to Ghostery Account both browsers. Setting Sync is ON on both. 
2. Chrome - FF Setting sync works great.
3. Turn off the sync option in Chrome and leave it enabled in FF.
4. I make changes to the Ghostery settings in Chrome. No reaction in FF. As expected.
5. Turn the sync option back on in Chrome. 
6. Make changes to the Ghostery settings in Chrome. 
7. Syncing stops working. No changes in FF after reloading the Ghostery settings page. 

The same happens other way around. (FF - Chrome)

smalluban:
@AdamGhst I made some tweaks, but even without them, I cannot reproduce the issue with sync off/on. However, with the latest changes, it is even better, as the sync process wasn't triggered when switching the toggle of that option (only after a refresh of the page). This is why you might have thought that it does not work.

src/utils/debounce.js
```diff
@@ -0,0 +1,43 @@
...
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+export default function debounce(fn, { waitFor, maxWait }) {
```
smalluban, line 12: To not introduce not needed changes, I did not use this new util in `content_scripts/adblocker.js` on purpose. I'll make another PR when this is merged.

AdamGhst:
> @AdamGhst I made some tweaks, but even without them, I cannot reproduce the issue with sync off/on. However, with the latest changes, it is even better, as the sync process wasn't triggered when switching the toggle of that option (only after a refresh of the page). This is why you might have thought that it does not work.

It is very, very possible that what I found was misunderstood from my side. Looking forward to test it again. :]

chrmod (commented):
It generally looks good. Can we please test it manually before the PR is merged? @AdamGhst @Dimi-GE

AdamGhst (approved):
Manual tests with Option Sync has been repeated. 
I came trough scenarios I found last time, and there is no behavior I was writing about in last run of testing. 
For now, this feature flies like a gold, works like a charm, run like clockwork! 

Tested on Chrome and Firefox. 
macOS. 
PC Win 10.

philipp-classen (approved):

PR #2074, opened 2024-11-19
fix(adblocker): inject scriplets with `browser.contentScripts` on Firefox

chrmod:
**WARNING:** This has to be tested very carefully.

On Firefox, scriptlet injection run into timing issues which makes it impossible to inject certain scriptlets before page scripts.

Examples we try to fix are:
* https://gorhill.github.io/uBlock/tests/scriptlet-injection-filters-1.html
* https://picrew.me

The only way to ensure the scriptlet is always executed before the page script is to use `browser.contentScripts` API which available on Firefox only.

Open questions:
* is injecting to `allFrames` safe? for example, can scriptlets from different hostnames be injected due to `allFrames`?
* do we need more triggering methods like `webRequest.onResponseStarted`?
* when content scripts should be cleaned? 
* do we handle redirects correctly? for example redirect from http://example.com to https://www.example.com ?
* can we speed up by not matching cosmetic filters for scriptlet injection once content scripts are registered? 

TODO:
* [ ] support selective blocking
* [x] support pause

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -183,6 +183,7 @@ const contentScripts = (() => {
           matchAboutBlank: true,
           matchOriginAsFallback: true,
           runAt: 'document_start',
+          world: 'MAIN',
         });
         map.set(hostname, contentScript);
       } catch (e) {
@@ -232,7 +233,7 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
     }
 
     if (__PLATFORM__ === 'firefox') {
-      contentScript += `(function () { ${func.toString()} })(...${JSON.stringify(args)})`
+      contentScript += `(${func.toString()})(...${JSON.stringify(args)});\n`
       continue;
     }
 
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -86,6 +86,9 @@ export async function reloadMainEngine() {
     await engines.create(engines.MAIN_ENGINE);
     console.info('[adblocker] Main engine reloaded with no filters');
   }
+  if (__PLATFORM__ === 'firefox') {
+    contentScripts.unregisterAll();
+  }
 }
 
 let updating = false;
@@ -163,7 +166,50 @@ export const setup = asyncSetup('adblocker', [
   ),
 ]);
 
-function injectScriptlets(filters, tabId, frameId) {
+const contentScripts = (() => {
+  const map = new Map();
+  return {
+    async register(hostname, code) {
+      this.unregister(hostname);
+      try {
+        const contentScript = await browser.contentScripts.register({
+          js: [
+            {
+              code,
+            },
+          ],
+          allFrames: true,
+          matches: [`https://*.${hostname}/*`, `http://*.${hostname}/*`],
+          matchAboutBlank: true,
+          matchOriginAsFallback: true,
+          runAt: 'document_start',
+        });
+        map.set(hostname, contentScript);
+      } catch (e) {
+        console.warn(e);
+        this.unregister(hostname);
+      }
+    },
+    isRegistered(hostname) {
+      return map.has(hostname);
+    },
+    unregister(hostname) {
+      const contentScript = map.get(hostname);
+      if (contentScript) {
+        contentScript.unregister();
+        map.delete(hostname);
+      }
+    },
+    unregisterAll() {
+      for (const hostname of map.keys()) {
+        this.unregister(hostname);
+      }
+    },
+  };
+})();
+
+function injectScriptlets(filters, tabId, frameId, hostname) {
+  let contentScript = '';
   for (const filter of filters) {
     const parsed = filter.parseScript();
 
@@ -177,9 +223,16 @@ function injectScriptlets(filters, tabId, frameId) {
 
     const scriptletName = `${parsed.name}${parsed.name.endsWith('.js') ? '' : '.js'}`;
     const scriptlet = scriptlets[scriptletName];
+    const func = scriptlet.func;
+    const args = parsed.args.map((arg) => decodeURIComponent(arg));
 
     if (!scriptlet) {
       console.warn('[adblocker] unknown scriptlet with name:', scriptletName);
+      continue;
+    }
+
+    if (__PLATFORM__ === 'firefox') {
+      contentScript += `(function () { ${func.toString()} })(...${JSON.stringify(args)})`
       continue;
     }
 
@@ -193,8 +246,8 @@ function injectScriptlets(filters, tabId, frameId) {
           tabId,
           frameIds: [frameId],
         },
-        func: scriptlet.func,
-        args: parsed.args.map((arg) => decodeURIComponent(arg)),
+        func,
+        args,
       },
       () => {
         if (chrome.runtime.lastError) {
@@ -202,6 +255,19 @@ function injectScriptlets(filters, tabId, frameId) {
         }
       },
     );
+  }
+
+  if (__PLATFORM__ === 'firefox') {
+    if (filters.length === 0) {
+      contentScripts.unregister(hostname);
+    } else if (!contentScripts.isRegistered(hostname)) {
+      contentScripts.register(
+        hostname,
+        contentScript,
+      );
+    } else {
+      // do nothing if already registered
+    }
   }
 }
 
@@ -224,6 +290,9 @@ function injectStyles(styles, tabId, frameId) {
 }
 
 async function injectCosmetics(details, config) {
+  const isBootstrap = config.bootstrap;
+  const scriptletsOnly = Boolean(config.scriptletsOnly);
+
   try {
     setup.pending && (await setup.pending);
   } catch (e) {
@@ -237,6 +306,8 @@ async function injectCosmetics(details, config) {
   const domain = parsed.domain || '';
   const hostname = parsed.hostname || '';
 
+  if (scriptletsOnly && contentScripts.isRegistered(hostname)) return;
+
   if (isPaused(options, hostname)) return;
 
   const tabHostname = tabStats.get(tabId)?.hostname;
@@ -245,7 +316,6 @@ async function injectCosmetics(details, config) {
   }
 
   const engine = engines.get(engines.MAIN_ENGINE);
-  const isBootstrap = config.bootstrap;
 
   {
     const { matches } = engine.matchCosmeticFilters({
@@ -282,8 +352,12 @@ async function injectCosmetics(details, config) {
       }
     }
 
-    if (isBootstrap && scriptFilters.length > 0) {
-      injectScriptlets(scriptFilters, tabId, frameId);
+    if (isBootstrap) {
+      injectScriptlets(scriptFilters, tabId, frameId, hostname);
+    }
+
+    if (scriptletsOnly) {
+      return;
     }
 
     const { styles } = engine.injectCosmeticFilters(styleFilters, {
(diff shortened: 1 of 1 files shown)


src/background/adblocker.js
```diff
@@ -86,6 +86,7 @@ export async function reloadMainEngine() {
     await engines.create(engines.MAIN_ENGINE);
     console.info('[adblocker] Main engine reloaded with no filters');
   }
+  contentScripts.clear();
```
smalluban, line 89: This is unrelated to reloading main engine, or if it is, it should be commented - for now it looks like leftover code.

src/background/adblocker.js
```diff
@@ -196,25 +219,53 @@ async function injectScriptlets(scripts, tabId, frameId) {
...
-      if (chrome.runtime.lastError) {
-        console.warn(chrome.runtime.lastError);
+  if (__PLATFORM__ === 'firefox') {
+    if (scripts.length === 0) {
+      contentScripts.delete(hostname);
+    } else if (!contentScripts.has(hostname)) {
+      try {
+        const contentScript = await browser.contentScripts.register({
+          js: [
+            { code: 'console.log(1111)' },
```
smalluban, line 229: Leftover?

src/background/adblocker.js
```diff
@@ -196,25 +219,53 @@ async function injectScriptlets(scripts, tabId, frameId) {
...
-      },
-      func: scriptletInjector,
-      args: [encodeURIComponent(scriptlets)],
-    },
-    () => {
-      if (chrome.runtime.lastError) {
-        console.warn(chrome.runtime.lastError);
+  if (__PLATFORM__ === 'firefox') {
+    if (scripts.length === 0) {
+      contentScripts.delete(hostname);
```
smalluban, line 224: Why not put the `delete()` method logic here? The concept of `contentScripts` hides logic, but the unregister is related to this function.

src/background/adblocker.js
```diff
@@ -167,7 +168,29 @@ export const setup = asyncSetup([
...
+  const map = new Map();
+  return {
+    set(key, value) {
+      map.set(key, value);
+    },
+    has(key) {
+      return map.has(key);
+    },
+    delete(key) {
+      const contentScript = map.get(key);
```
smalluban, line 181: The only difference from "normal" map is this function, which should be part of the `injectScriptlets`.

src/background/adblocker.js
```diff
@@ -167,7 +168,29 @@ export const setup = asyncSetup([
  * Cosmetics injection
  */
 
-async function injectScriptlets(scripts, tabId, frameId) {
+const contentScripts = (() => {
```
smalluban, line 171: `contentScripts` is not persistent, and with every reload of the background process, the map will be empty. The same hostname can be injected many times.

chrmod: content scripts registered by the `browser.contentScript` are also non persisted and live only as long as the background process. So this is correct

src/background/adblocker.js
chrmod: this is very much related to reloading of the main engine. When filters change we have to clear the content script so the new version can be applied

src/background/adblocker.js
```diff
@@ -258,6 +309,7 @@ async function injectCosmetics(details, config) {
 
   const engine = engines.get(engines.MAIN_ENGINE);
   const isBootstrap = config.bootstrap;
+  const scriptletsOnly = Boolean(options.scriptletsOnly);
```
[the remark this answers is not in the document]
chrmod: it was meant to be config.scriptletsOnly

src/background/adblocker.js
chrmod: we do delete in mutliple places

src/background/adblocker.js
```diff
@@ -167,7 +168,29 @@ export const setup = asyncSetup([
...
+    set(key, value) {
+      map.set(key, value);
+    },
+    has(key) {
+      return map.has(key);
+    },
+    delete(key) {
+      const contentScript = map.get(key);
+      contentScript.unregister();
+      map.delete(key);
```
philipp-classen, line 183:
```suggestion
      if (contentScript) {
        contentScript.unregister();
        map.delete(key);
      }
```

src/background/adblocker.js
smalluban: Ok, I checked.

>Content scripts are also unregistered automatically when the page that created them is destroyed. For example, if they are registered from the background page they will be unregistered automatically when the background page is destroyed, and if they are registered from a sidebar or a popup, they will be unregistered automatically when the sidebar or popup is closed.

src/background/adblocker.js
smalluban: At least add `__PLATFORM__` check, so `contentScripts` can be tree-shaked.

src/background/adblocker.js
smalluban: In one function. Instead of proxying map, just create two functions `clearContentScripts()` and `deleteContentScript(hostname)`, and "normal" at the module level `Map` to keep track of hostnames.

chrmod: what's the problem with this solution? this is just encapsulation

smalluban: It looks that this is some kind of map, but it does side effects in functions, which normally don't do that. It is not obvious that calling `contentScripts.delete()` it will also unregister, until you read the implementation. If you would create function `unregisterContentScripts()` (my new idea of naming), it might contain `map.delete(key)`, but the logic is explicit.

chrmod: but then someone could modify the map and forget to unregister the content script

smalluban: The current implementation looks much better.

src/background/adblocker.js
```diff
@@ -196,6 +241,21 @@ async function injectScriptlets(scripts, tabId, frameId) {
     script.remove();
   }
 
+  if (__PLATFORM__ === 'firefox') {
+    if (scripts.length === 0) {
+      contentScripts.unregister(hostname);
```
philipp-classen, line 246: I'm wondering if this part of the code is necessary or how it should be best handled. 
When the engine updates, it should automatically clear the content scripts (contentScripts.unregisterAll). There might be potential for race conditions, but in principle, it should eventually take care of unregistering.

Is the code here intended especially for pause (which won't trigger an engine reload)? If so, would be good to leave a comment explaining this.

GRadziejewski:
Tested version: https://github.com/ghostery/ghostery-extension/pull/2074#issuecomment-2488599030

Checked on:
```
Windows 11, Firefox 132.0.2, GBE 10.4.15 and 10.4.10
Windows 10, Firefox 132.0.2, GBE 10.4.15 and 10.4.10
```

Websites::

```
https://example.com/ - OK
https://stre4mplay.one/ - OK
https://www.officedepot.com/ - OK
https://picrew.me/ja/ - OK
https://www.foxbusiness.com/ - OK
https://edition.cnn.com/ - OK
https://www.br.de/index.html - OK
https://indeed.com/ - OK
https://pasteboard.co/ - OK
https://los40.com/ - OK
https://www.telegraph.co.uk/ - OK
https://www.verizon.com/ - On 10.4.10 version it's not working, on 10.4.15 it's working!
https://humanbenchmark.com/ - OK
https://www.politico.com/ - OK
https://kusonime.com/ - OK
https://modapk.link/ - OK
https://xxxymovies.com/ - OK
https://www.jetpunk.com/ - OK
https://imgur.com/ - OK
https://www.hentai-party.net/ - OK
https://ccurl.net/ - OK
https://mysexgames.com/ - OK
https://www.ancient-origins.net/ - OK
https://hideandseek.world/ - OK
https://pewgame.com/ - OK
https://haonguyen.top/ - OK
https://crazyblog.in/ - OK
https://cutearn.net/ - OK
rshrt.com - > https://reshort.com/ - OK
https://gplinks.org/ - OK
https://easysky.in/ - OK
https://veganab.co/ - OK
https://www.gyanitheme.com/ - OK
https://fitdynamos.com/ - OK
https://postazap.com/ - OK
https://blog24.me/ - OK
https://atglinks.com/ - OK
https://streamelements.com/ - OK
https://www.facebook.com - OK
https://infinityscans.xyz/ - OK
https://www.youtube.com/ - OK
```

src/background/adblocker.js
```diff
@@ -224,6 +296,9 @@ function injectStyles(styles, tabId, frameId) {
 }
 
 async function injectCosmetics(details, config) {
+  const isBootstrap = config.bootstrap;
+  const scriptletsOnly = Boolean(config.scriptletsOnly);
```
smalluban, line 300: We are in control over the config, and `undefined` is just fine, you don't need to convert it to Boolean.

src/background/adblocker.js
```diff
@@ -237,6 +312,8 @@ async function injectCosmetics(details, config) {
   const domain = parsed.domain || '';
   const hostname = parsed.hostname || '';
 
+  if (scriptletsOnly && contentScripts.isRegistered(hostname)) return;
```
smalluban, line 315: This `scripletsOnly` is not what the name suggests - it is a flag for `EXPERIMENTAL_SCRIPTLETS_INJECTION` - you can't use `scripletsOnly` in Chrome if you would wish for example.

There is an unrelated bug in Firefox requiring the separation of injecting styles and scriptlets (styles must be injected not in `onCommited`, as then we have permission for an issue in iframes). So as this function becomes harder to reason about, this is a moment to do so...

src/background/adblocker.js
```diff
@@ -153,6 +158,12 @@ export const setup = asyncSetup('adblocker', [
   OptionsObserver.addListener(
     'experimentalFilters',
     async (value, lastValue) => {
+      if (__PLATFORM__ === 'firefox') {
+        EXPERIMENTAL_SCRIPTLETS_INJECTION = value;
```
smalluban, line 162: Please create another observer to separate logic. You are using the same option, but for another purpose.

smalluban (changes requested):

src/background/adblocker.js
chrmod: not quite right, this block is just an optimization that prevent a matching if content scripts are already registered. `scriptletsOnly` main role is to prevent style injection a bit below.

But yes, the condition should be improved

GRadziejewski:
To test this, turn on the experimental filters.

GRadziejewski:
Tested version: https://github.com/ghostery/ghostery-extension/pull/2074#issuecomment-2539415134

Checked on:
```
Windows 10, Firefox 134.0 GBE 10.4.18
```

Websites:
```
https://example.com/ - OK
https://stre4mplay.one/ - OK
https://www.officedepot.com/  - OK
https://www.foxbusiness.com/ - OK
https://edition.cnn.com/ - OK
https://www.br.de/index.html - OK
https://indeed.com/ - OK
https://pasteboard.co/ - OK
https://los40.com/ - OK
https://www.telegraph.co.uk/ - OK
https://www.verizon.com/ - OK
https://humanbenchmark.com/ - OK
https://www.politico.com/ - OK
https://kusonime.com/ - OK
https://modapk.link/ - OK
https://xxxymovies.com/ - OK
https://www.hentai-party.net/ - OK
https://mysexgames.com/ - OK
https://www.ancient-origins.net/ - OK
https://hideandseek.world/ - OK
https://pewgame.com/ - OK
https://haonguyen.top/ - OK
https://crazyblog.in/ - OK
https://cutearn.net/ - OK
https://reshort.com/ - OK
https://gplinks.org/ - OK
https://easysky.in/ - OK
https://veganab.co/ - OK
https://www.gyanitheme.com/ - OK
https://fitdynamos.com/ - OK
https://atglinks.com/ - OK
https://www.facebook.com - OK
https://www.reddit.com/ - OK
https://www.youtube.com/ - OK
```

LGTM

PR #2082, opened 2024-11-21
Ad-blocking: :has support

chrmod:


philipp-classen:
For context: this pull request upgrades to the latest adblocker library version and opts-in to the native :has selector that came in https://github.com/ghostery/adblocker/pull/4455.

philipp-classen (approved):

PR #2107, opened 2024-11-29
fix(engines): use options observer instead of internal listener

smalluban:
The update listener of the engines utility was already decoupled from options here https://github.com/ghostery/ghostery-extension/blob/78533f01cc0bb179eef8b49f520ca4fbf2b5a42d/src/background/custom-filters.js

This PR removes the feature completely. The exception filters were listening to the change of the TrackerDB to update the DNR rules. It can be also done by observing the `filtersUpdatedAt` option. Then, the mechanism in engines can be removed.

### DIFF
--- a/src/background/exceptions.js
+++ b/src/background/exceptions.js
@@ -137,12 +137,14 @@ async function updateFilters() {
     .filter(({ id }) => id >= 2000000)
     .map(({ id }) => id);
 
-  await chrome.declarativeNetRequest.updateDynamicRules({
-    addRules,
-    removeRuleIds,
-  });
+  if (addRules.length || removeRuleIds.length) {
+    await chrome.declarativeNetRequest.updateDynamicRules({
+      addRules,
+      removeRuleIds,
+    });
 
-  console.info('[exceptions] Updated DNR rules');
+    console.info('[exceptions] Updated DNR rules');
+  }
 }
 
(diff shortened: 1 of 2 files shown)


src/background/exceptions.js
```diff
@@ -142,10 +142,20 @@ async function updateFilters() {
...
   // Update exceptions filters every time TrackerDB updates
-  engines.addChangeListener(engines.TRACKERDB_ENGINE, updateFilters);
+  // It happens when all engines are updated
+  OptionsObserver.addListener(
+    'filtersUpdatedAt',
+    async function updateExceptions(value, lastValue) {
+      // Only update exceptions filters if the value has changed and is set to timestamp.
+      // It will happen only after successful update of the engines.
+      if (lastValue !== undefined && value !== 0) {
+        await updateFilters();
```
chrmod, line 157: this was previously happening for trackerdb updates only. 
IDK if it this better or worse now. Do we want to run it when Custom Filters change?

smalluban: Yes, technically, it was more accurate. However, it only applies to exceptions, and most users will have none, or very few. The case is to stop keeping complex logic just for one place. Even though the change is not accurate, I think it is worth simplifying the structure and re-use patterns as much as possible.

I added a condition for only touching DNR rules if there is a change - in the current implementation triggering an update for users without exceptions does nothing.

chrmod (approved):

PR #2117, opened 2024-12-04
Release v10.4.17

smalluban:
63e2ce773ccef440c49188e1a0eac770d7e5ea9f fix(trackers-preview): e2e tests & UI clean up (#2108)
7bf56c2b638e1c566b95f2d5b72a3763289a81f5 fix: minor updates to the build script and assets (#2113)
0e551f542d0ae4e9b76b70d569dc08348c147255 Temporarily disable anti-tracking fingerprinting removal in URLs (#2102)
d1af660a362af5a0fc85ca8dd81aec7c716cfdcb fix(build): bypass downloading resources on sequential builds (#2096)
5f6f83c64fe5903ed49be0eb2c48c6f5cc460db1 fix(settings): clean up contributor message for Safari (#2100)


AdamGhst (changes requested):
SMOKE TESTS 

macOS
Chrome - passed
FF - passed 
Safari - passed

Windows
Chrome - passed
FF - FAILED Never Consent Test. (Never Consent button does not work. Never-Consent triggers even if feature is on or off)

smalluban:
@AdamGhst I doubt that feature can work on macOS Firefox, but it fails on Windows. This is the same Firefox. The never-consent feature has to be tested in a clean browser. If you accidentally open the page before disabling never-consent, you will no longer see the CMP popup, even if you disable the feature.

I've just re-tested on my machine, and everything works as expected. It also passes tests, which run on Linux. Another argument is that this release does not change anything with the never-consent feature (compared to current production).

AdamGhst (approved):

AdamGhst:
> @AdamGhst I doubt that feature can work in Firefox on macOS, but it fails on Windows. This is the same Firefox. The never-consent feature has to be tested in a clean browser. If you accidentally open the page before disabling never-consent, you will no longer see the CMP popup, even if you disable the feature.
> 
> I've just re-tested on my machine, and everything works as expected. It also passes tests, which run on Linux. Another argument is that this release does not change anything with the never-consent feature (compared to current production).

After rechecking it looks like I had a false positive. My PC machine is causing some problems. I need to investigate. After reinstalling the release couple of times and re checking, everything worked perfectly.

AdamGhst:
Smoke tests. 
macOS
FF - all passed
Chrome - all passed
Safari - all passed

Win
FF and Chrome - all passed

PR #2119, opened 2024-12-05
fix(adblocker): inject scriptlets from source code

chrmod:
instagram.com and facebook.com introduced strict CSP that prevent current scriplet injection method that we use currently. To bypass those restrictions we have to pass scritpelts as code with help of `chrome.scripting.executeScript`. 

This however require us to ship the scriptlets and evaluate them in the background context. 

A new repository to provide scriptlet code is created https://github.com/ghostery/scriptlets. As its code comes from other project like uBO, there is no need for any kind of versioning scheme so we will use git commit checksums instead and install from github directly.

As this means we have to ship GPL code with the extension, we are forced to switch license from MPL-v2 to GPL-v3.

Please test on facebook.com, instagram.com and youtube.com.


package.json
```diff
@@ -19,6 +19,7 @@
     "locales": "npm run locales:setup && npm run locales.src",
     "xcode-export-locales": "rm -rf ./xcode/en.xcloc && xcodebuild -exportLocalizations -project ./xcode/Ghostery.xcodeproj -localizationPath ./xcode",
     "release": "./scripts/release.sh",
+    "udpate-scriptlets": "deno scripts/update-scriptlets.js --tagName 1.61.3b4 > ./src/rule_resources/scriptlets.js",
```
smalluban, line 22: If downloading scripts requires Deno, and we cannot avoid it - the file should be generated on the server and downloaded like Trackers Preview data. We can provide just what this script outputs.

chrmod: if the POC works then we can follow suggestion of @philipp-classen and move the scriptlets to a separate repository

src/background/adblocker.js
```diff
@@ -23,11 +24,18 @@ import * as trackerdb from '/utils/trackerdb.js';
 import * as OptionsObserver from '/utils/options-observer.js';
 import Request from '/utils/request.js';
 import asyncSetup from '/utils/setup.js';
-import { debugMode } from '/utils/debug.js';
 
 import { tabStats, updateTabStats } from './stats.js';
 import { getException } from './exceptions.js';
 
+const scriptlets = new Map();
+for (const [name, scriptlet] of Object.entries(SCRIPTLETS)) {
```
smalluban, line 32: I would move this to the `@ghostery/scriptlets` as default export (a final map).

chrmod: done - please review related PR https://github.com/ghostery/scriptlets/pull/1

src/background/adblocker.js
```diff
@@ -15,6 +15,7 @@ import {
   updateResponseHeadersWithCSP,
 } from '@ghostery/adblocker-webextension';
 import { parse } from 'tldts-experimental';
+import SCRIPTLETS from '@ghostery/scriptlets';
```
smalluban, line 18: I don't have strong opinion, but I would change to `scriptlets`

philipp-classen (approved):

PR #2120, opened 2024-12-06
PoC: integration with ghostery/adblocker#4499 and uBo origin

seia-soto:
This PoC proposes the use of https://github.com/ghostery/adblocker/pull/4499 while integrating with origin uBo scriptlet sources

refs https://github.com/ghostery/adblocker/pull/4499
refs https://github.com/ghostery/ghostery-extension/pull/2119


chrmod:
@seia-soto to be used with `chrome.scripting.executeScript` scriptlets need to have no external references. That mean, those have to be prepare on the built time.

PR #2123, opened 2024-12-06
fix(telemetry): clean up unused code & minor storage fixes

smalluban:
This PR does not change the logic of sending signals and storing signal timestamps. The main goal is to clean up code that relies on values we don't provide, so their values are always "the same". 

* The query parameters were only re-ordered to decide whether we need them at all or if we can clean them up too.
* Metrics data is stored in one storage key
* Custom config can be replaced with direct access to options

I've tested the following cases in both Chrome and Firefox:
* Clean install - initial ping - ping after a day (manually updated timestamp)
* Clean install with utms
* Update from production with and without UTMs

It might be easier to look at the final code rather than compare it - you can click a dot to open PR in online VS Code.

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -16,83 +16,80 @@ import { debugMode } from '/utils/debug.js';
 import asyncSetup from '/utils/setup.js';
 import * as OptionsObserver from '/utils/options-observer.js';
 
-import Metrics from './metrics.js';
+import Metrics, { processUrlQuery } from './metrics.js';
 
-async function loadStorage() {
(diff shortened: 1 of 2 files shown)


philipp-classen (approved):

chrmod (approved):
We can remove `_getSearchExtensionMetrics`

smalluban:
@chrmod As you want to remove some parameters, can we also remove all obsolete values? Those are hardcoded, and have no value to us.

smalluban:
For the sake of commits log I will create another PR with clean up of the parameters.

PR #2136, opened 2024-12-10
Adblocker: disable engine compression

chrmod:
With @ghostery/adblocker 2.3.0 it is now possible to disable engine compression on merge. This helps to reduce the overall merge time (including storage) from ~670ms to ~370ms. Additionally, uncompressed engine should perform better on matching. The cost is increased engine size with when compressed was an UintArray with size of 72055507, and when uncompressed it size is 8942525.

### DIFF
--- a/src/utils/engines.js
+++ b/src/utils/engines.js
@@ -459,7 +459,12 @@ export function replace(name, engineOrEngines) {
   const engines = [].concat(engineOrEngines);
   const engine =
     engines.length > 1
-      ? FiltersEngine.merge(engines, { skipResources: true })
+      ? FiltersEngine.merge(engines, {
+          skipResources: true,
+          overrideConfig: {
+            enableCompression: false,
+          },
+        })
       : engines[0];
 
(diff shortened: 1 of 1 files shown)


philipp-classen (approved):

smalluban (approved):
The change looks fine. If the e2e tests pass, it should be safe to merge - my only concern is a state where a user has a compressed version and changes to the new code (by changing some options or updating engines). You can't test it without updating from prod.

src/background/custom-filters.js
```diff
@@ -146,7 +146,7 @@ function updateEngine(text) {
     cosmeticFilters,
     networkFilters,
     preprocessors,
-    config: engines.get(engines.MAIN_ENGINE).config,
```
smalluban, line 149: If it is important, please change that in separate PR.

smalluban:
I understand now the change with fixes engine, but it won't work, as only main engine is available after restart of the background process.

seia-soto (approved):

PR #2164, opened 2024-12-23
feat(panel): Add the "Report a broken page" view

smalluban:
The PR adds the CTA for paused domains, which opens the "Report a broken page" in the panel. 

* It looks like `activeTab` permission does not trigger a browser popup when the extension is updated. However, we can also try out the optional permission if it is better.
* We need to match operating systems between the extension and support form - I think we should update fields to follow extension logic - `mac`, `win`, `ios`, `ipados`,`linux`, etc... and version only as a separate field (currently we have `windows_10` and other as os name)
* Copy is a draft version - needs review and improvements
* I tested the screenshot feature on Chrome, Firefox, and Safari on Mac and iOS, and it works as expected
* Open question about the link "Report a broken page" in the menu - should it also open the form in the panel or should it open the website as it does now?

WARNING: the PR logic is finished and it's connected to the real end-point. Please use `npm start -- --staging` when testing to avoid spaming the support channel.


src/background/broken-page-report.js
```diff
@@ -0,0 +1,67 @@
...
+        const formData = new FormData();
+        const browserInfo = await getBrowserInfo();
+
+        formData.append('support_ticket[user_name]', '');
+        formData.append('support_ticket[user_email]', msg.email);
+        formData.append(
+          'support_ticket[subject]',
+          `[GBE] Broken page report: ${msg.url}`,
+        );
+        formData.append('support_ticket[message]', msg.description);
```
chrmod, line 28: We should pass the product version via the form. Until it is supported by the website lets add it to the message. 

Lets consider if we want to add also: 
* list of detected activities (no urls) 
* extension settings 
* block list version (checksums)

src/background/broken-page-report.js
```diff
@@ -0,0 +1,67 @@
...
+        formData.append('support_ticket[selected_browser]', browserInfo.name);
+        formData.append('support_ticket[browser_version]', browserInfo.version);
+        formData.append('support_ticket[selected_os]', browserInfo.osVersion);
+        formData.append('support_ticket[os_version]', '');
+
+        if (msg.screenshot) {
+          const screenshot = await chrome.tabs.captureVisibleTab(null, {
+            format: 'jpeg',
+            quality: 100,
```
chrmod, line 37: currently the height defaults to the view port height, can we pass a height of the full page to capture everything?

src/background/broken-page-report.js
```diff
@@ -0,0 +1,67 @@
...
+          body: formData,
+        }).then((res) => {
+          if (!res.ok || res.status > 204) {
+            throw new Error(
+              `Sending report has failed with status: ${res.status}`,
+            );
+          }
+        });
+
+        sendResponse();
```
chrmod, line 57: should we add some error handling so in case of a problem we let people know they can report the problem via the website?

chrmod (commented):
Looks great so far!

src/background/broken-page-report.js
smalluban: The `rect` option is only supported in Firefox (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extensionTypes/ImageDetails#rect), and I think it might be tricky to detect the height of the whole page, and still I am not sure if without manual "scrolling" you can take a screenshot of the part, which is not visible. I would keep it as it is for now.

src/background/broken-page-report.js
smalluban: Good point. That's technically easy, I will add a message after error handling, like "Error... However, you can always report the broken page via our [website form]"

smalluban:
*  Updated copy from @mdsandu
* Added version, options and trackers to description
* Additional info when error occurs

I think the PR is ready for final review for merging.

smalluban:
There is a last open question - what about the link to report a broken page in the menu section of the panel? I am unsure if it should be changed, as the user might want to report "some" page - it does not have to be the one it is now... And if we change that, there will be no direct link to the form on the ghostery.com

mdsandu:
pls post current screens to resolve the last issue. thanks

smalluban:
Updated copy:

Zrzut ekranu 2025-01-14 o 09 24 57
Zrzut ekranu 2025-01-14 o 09 29 48
Zrzut ekranu 2025-01-14 o 09 30 35

Dimi-GE:
Screenshots from Zendesk:

image

image

image

image

Dimi-GE (approved):

mdsandu:
@smalluban two suggested changes to cover your concern
same text in 2 locations clarifying (hopefully)

- [ ] on screen 1 add "If the issue persists or you’d like to report a different page as broken, please use the report form on ghostery.com" below "Include a screenshot of the current page" with a bit of spacing in between. 


- [ ] on screen 2/fail screen: expand the red text to same message "If the issue persists or you’d like to report a different page as broken, please use the report form on ghostery.com"

let's see how this feels?

smalluban:
Keeping the message below is a good idea, as a case with an error might be rare. However, I don't like displaying it twice and hiding the second one (Ui flickering). I would keep only point 1 - display the message all the time regardless of the error (which can never be shown).

Zrzut ekranu 2025-01-16 o 08 01 34
Zrzut ekranu 2025-01-16 o 08 04 03

mdsandu:
hi @smalluban, not sure I fully understand the final state you are suggesting
how will the fail state look like (I am not a fan of showing the same message twice on a screen)
let's clarify offline

smalluban:
> I would keep only point 1 

It's simple - just keep the message below the "include screenshot" all the time, and not display another (or switch) when error happens.

mdsandu:
ah ok, now I get it! great! let's move on!

smalluban:
I tested Chrome, and there is no confirm popup after update with added `activeTab` permission. Using the same technique I am able to block the extension when adding `geolocation` permission (to be sure that I test it right).

I assume that only Chrome is affected by the "blocking popup" when permissions are added.

chrmod (approved):

PR #2171, opened 2024-12-31
fix: Use full hostname & unify pause logic

### LINKED ISSUE #2168: Ghostery is not paused
**Description**
When Ghostery is set to paused on an open page, the extension is not paused - it continues to work.

**Precondition**: 
new profile / cleared history / newly installed extension

**Steps to Reproduce**
1. Install Ghostery
2. Open https://cnn.com/
3. Open Ghostery panel
4. Pause Ghostery for 1 hour
5. Page reloaded
6. Open Ghostery panel - the extension is not paused on this website
7. Open Ghostery Settings / Websites
image

8.  Open https://cnn.com/ again
9. Open Ghostery panel
10. Pause Ghostery for 1 hour
11. Page reloaded
12. Open Ghostery panel - the extension is paused on this website
13. Open Ghostery Settings / Websites again
image

https://github.com/user-attachments/assets/4643f62e-aece-447f-b79b-d1d61bfdc997

**Expected results**
Ghostery should be paused.

```
Browser: Chrome 131.0.6778.205
OS: Windows 10
GBE: 10.4.19

Browser: Chrome 131.0.6778.205
OS: Windows 11
GBE: 10.4.19
```

smalluban:
Fixes #2168 and unifies pause behavior on subdomains if the top domain is paused.

* The tab stats setup should be done in `onCommited` instead of `onBeforeNavigate` to properly trigger setup after initial 30X redirects - the `cnn.com` or `onet.pl` sites are good examples - the `onCommited` is triggered once, but after redirects, with the final URL
* The removal of the `www.` subdomain was mainly caused by the above bug to avoid situations where not the latest (and correct) hostname was added to paused websites
* Instead of removing the `www` subdomain, we can only hide it in the panel header, as the browsers hide it in the URL address bar
* The DNR allow rule for `exmaple.com` works also for `abc.exmaple.com` - because of that this PR also unifies the behavior of the `isPaused()` function and `isTrusted()` for exceptions in Firefox - for the consistent logic through all the platforms. Unification helps with no-stress migration from the current production logic - if a user added `onet.pl`, it will work with unified logic as well, as the `www.onet.pl` contains the `onet.pl` (for isPaused and exceptions).

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -357,15 +357,17 @@ function isTrusted(request, type) {
   const exception = getException(metadata?.id || request.hostname);
 
   if (exception) {
-    const tabHostname = request.sourceHostname.replace(/^www\./, '');
-
     // The request is trusted if:
     // - tracker is blocked, but tab hostname is added to trusted domains
     // - tracker is not blocked and tab hostname is not found in the blocked domains
     if (
       exception.blocked
-        ? exception.trustedDomains.includes(tabHostname)
-        : !exception.blockedDomains.includes(tabHostname)
+        ? exception.trustedDomains.some((id) =>
+            request.sourceHostname.endsWith(id),
+          )
+        : !exception.blockedDomains.some((id) =>
+            request.sourceHostname.endsWith(id.sourceHostname),
+          )
     ) {
       return true;
     }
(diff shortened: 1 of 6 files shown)


AdamGhst (approved):
Works its magic!

PR #2176, opened 2025-01-08
fix(e2e): optional WTM tests with external URLs

smalluban:
The WTM feature requires loading an external URL. It seems that sometimes browsing an external URL fails, so we need to add an optional condition to proceed if the page is loaded.


chrmod (commented):
How about we inject the Trackers Preview to a test page instead? 
Ideally we should extend Trackers Preview to Bing, DDG, Brave.

smalluban:
@chrmod Sorry, I missed your comment, I planned to just merge it ;) 

Injecting somewhere else is not that "easy", as this is a static config of content script, which detects the page with proper content, etc... Another solution would be to mock the SERP statically, and somehow provide it with proper domain name in the testing environment...

PR #2179, opened 2025-01-10
fix(session): Safari bug with delayed set of cookies

smalluban:
Fixes the bug on iOS Safari 18.x, which makes the account section on settings page out of date after sucessful login.

Reproduce path (on production):

1. Open Account section on Settings page
2. Click login button
3. After going back to Settings, the page still shows a status to login in

This PR also changes a patch for another bug to be feature-based, rather than fixed to the platform (in the case Safari would fix the bug). For example, the problem with starting point from wrong date for cookies is fixed in 18.x (src/utils/api.js#L60)


smalluban:
@AdamGhst As Safari requires builds, I'll test it. However, I would like you to test other platforms to be sure that login/logout and sync (it uses cookies) works as before.

AdamGhst (approved):
PR tested. 
macOS:
FF and Chrome - passed.
WIN:
FF, Chrome and Edge
Passed.

chrmod (approved):

PR #2200, opened 2025-01-21
fix(adblocker): use config flag for firefox content scriptlets injection

smalluban:
Moves from `experimentalFilters` to config flag `firefox-content-script-scriptlets` to enable experimental scriplets injection in Firefox.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -27,8 +27,28 @@ import asyncSetup from '/utils/setup.js';
 
 import { tabStats, updateTabStats } from './stats.js';
 import { getException } from './exceptions.js';
+import Config, {
+  FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
+} from '/store/config.js';
 
 let options = Options;
+
+let ENABLE_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS = false;
+if (__PLATFORM__ === 'firefox') {
+  // FYI: Testing the flag by dev tools requires a reload of the background script
+  store.resolve(Config).then((config) => {
+    ENABLE_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS = config.hasFlag(
+      FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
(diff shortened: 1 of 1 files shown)


src/background/adblocker.js
```diff
@@ -167,23 +184,6 @@ export const setup = asyncSetup('adblocker', [
   ),
 ]);
 
-let ENABLE_FIREFOX_CONTENT_SCRIPT_SCRIPLET_INJECTION = false;
```
smalluban, line 170: The variable is used above that part of the code. I recommend to define variables on top of the code which uses them.

src/background/adblocker.js
```diff
@@ -27,9 +27,73 @@ import asyncSetup from '/utils/setup.js';
 
 import { tabStats, updateTabStats } from './stats.js';
 import { getException } from './exceptions.js';
+import Config, {
+  FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
```
chrmod, line 31: it would be easier to review if you would keep the original name

src/background/adblocker.js
```diff
@@ -27,9 +27,73 @@ import asyncSetup from '/utils/setup.js';
...
+} from '/store/config.js';
 
 let options = Options;
 
+const contentScripts = (() => {
```
chrmod, line 36: same here, keeping the code in original position would make it easier to review

src/background/adblocker.js
smalluban: It's a waste to make another PR just for changing names... ;)

src/background/adblocker.js
smalluban: I make position changes only when I think it is important - if we define variables, if it is possible, they should be defined before they are used by another part of the code.

This is PR, which touches this feature, so this is the best place to make such a change. For sake of visibility I'll make those changes in separate commits next time, so it will be easier to check logic change.

chrmod: two commits would work perfectly well, lets do that in the future

chrmod (approved):

PR #2202, opened 2025-01-22
fix: typo in `injectCosmetics` config

seia-soto:
It should be `isBootstrap` not `bootstrap`. Confirmed while debugging a website.

image


smalluban (changes requested):
There is a rename of the config `bootstrap` in L315:
```js
const { bootstrap: isBootstrap, scriptletsOnly } = config;
```

PR #2284, opened 2025-03-05
fix(safari): use svg icon for toolbar

### LINKED ISSUE #2271: iphone: ghostery icon seems blurry, pls investigate
https://ghostery.slack.com/archives/C03MP7U2GJ0/p1740478974555529

Image

smalluban:
Fixes #2271.

It must be a Safari bug, as I saw a blurry icon of other extensions. Also, it is not well documented, that Safari has started supporting SVG format for a toolbar icon - when creating a new project in Xcode (https://developer.apple.com/documentation/safariservices/creating-a-safari-web-extension) the current default for action icon is SVG image.

I've checked iOS 17.x, iOS 18.x, and Safari 18.x on macOS, and it looks that all of them supports correctly SVG images.


smalluban:
Safari requires adding a slight border to the inactive icon - I've checked Firefox and Safari on macOS and iOS, and the border is not visible.

chrmod (approved):

Issue #2227, opened 2025-01-29
Adding Bing Search Engine Redirect Protection


Dimi-GE:
Preventing Bing from redirecting result links through their servers.

chrmod:
Looks like we can extract target url from the BING redirect, for example:
https://www.bing.com/ck/a?!&&p=6cd30a967ff2ee999ca5216def7c82e3627199cda5c93e3bd31eb936cf94e2f2JmltdHM9MTc0MjI1NjAwMA&ptn=3&ver=2&hsh=4&fclid=3a90dfd6-373d-6c61-0733-cae9361c6d76&u=a1aHR0cHM6Ly93d3cuZmlyZXByb2JlLm5ldC9wbA&ntb=1

Has a parameter `u` with value `a1aHR0cHM6Ly93d3cuZmlyZXByb2JlLm5ldC9wbA`. If we drop first two characters (`a1`) there rest is base 64 encoded url: `atob('aHR0cHM6Ly93d3cuZmlyZXByb2JlLm5ldC9wbA') === "https://www.fireprobe.net/pl"`

Linhao-Qian:
Hi @chrmod @smalluban ! I’ve implemented this feature request — please feel free to review it when convenient.  
Also, while working on it, I noticed that the Google Search Engine Redirect Protection wasn’t functioning correctly, so I fixed that as well. The pull request link is : https://github.com/ghostery/ghostery-extension/pull/2335
Looking forward to your thoughts and suggestions!

smalluban:
Fixed by #2335

PR #2335, opened 2025-03-29
feat (issue #2227) : Adding Bing Search Engine Redirect Protection

Linhao-Qian:
~~I did two things in this commit:~~
~~1. I fixed the bug that Google Search Engine Redirect Protection doesn't work. The 'url' search param has been replaced by 'q'. I found this bug and it has not been reported before.~~
~~2. I added Bing Search Engine Redirect Protection according to the discussion in the issue #2227~~

Apologies for the earlier confusion. Upon further investigation, I realized that the original Google Search Engine Redirect Protection is functioning correctly, and the parameter should indeed be 'url', not 'q'. Therefore, this pull request now exclusively focuses on adding Bing Search Engine Redirect Protection.


ChrisZHHG:
Hi @smalluban, I've reverted the changes in the locales folder and made a minor modification to expand the range of Bing URL matching, since it can vary (e.g., cn.bing.com, search.bing.com, etc.). Thanks!

smalluban:
Please ensure code styling by the `npm lint` command.

src/content_scripts/prevent-serp-tracking.js
```diff
@@ -17,8 +17,27 @@ function safeLinkClick(event) {
 
   el.removeAttribute('ping');
 
-  const targetUrl =
-    el.pathname === '/url' && new URL(el.href).searchParams.get('url');
+  let targetUrl = null;
+  // Google Search: extract direct link from redirect URL
+  if (el.pathname === '/url') {
+    targetUrl = new URL(el.href).searchParams.get('q');
```
philipp-classen, line 23: Can you please explain with an example why now "q" instead of "url"? Also, why chaing the regexp for Google? Perhaps we should split this PR in two parts? Because I think they are solving unrelated topics. As I understand, one is a potential bug fix for an unreported bug (perhaps you can also share more information about what is exactly breaking). The other is Bing support.

Linhao-Qian: Sorry to mix the "google redirect fix" and the "bing redirect support" in this PR. I didn't split them because I think they are all about search engine redirect protection.

An example of Google Redirect URL is "https://www.google.com/url?q=https://www.example.com&sa=U&ved=2ahUKEwi&usg=AOvVaw0". If you try to go to "https://www.google.com/url?url=https://www.example.com&sa=U&ved=2ahUKEwi&usg=AOvVaw0" in the browser, the chrome browser will say that is an invalid url. That said, the search param of Google Redirect URL is "q" instead of "url".

I changed the regexp to make the prevent-serp-tracking.js file running when we are on the bing search page. Is this causing a conflict with current logic?

And, should I submit another issue to report the Google Redirect Protection problem and split that fix from this PR?

philipp-classen: Can you give us concrete steps on how we can reproduce the problem with Google that you describe?

Linhao-Qian: 1. Make sure the Ghostery extension and its Search Engine Redirect Protection is on;
2. Find an element with a Google Redirect Url. It may be an <a> tag and has an attribute like "href = https://www.google.com/url?q=https://www.example.com&sa=U&ved=2ahUKEwi&usg=AOvVaw0". If you cannot find such an element, you can open the console, choose an <a> tag and change its href attribute to the link above.
3. Click the element, you will find that google still redirect you to the redirect warning page, which means the Google Search Engine Redirect Protection doesn't work.
4. Change the code "new URL(el.href).searchParams.get('url')" into "new URL(el.href).searchParams.get('q')" (prevent-serp-tracking.js line 21)
5. After changing the code, repeat the steps above. You will find that Ghostery can rewrite the Google Redirect Url to the target url directly when clicking the element, and you will not see the redirect warning page any more.

The root cause is that Google Redirect URL uses 'q' instead of 'url' as its search param. 

Hope this can make the question clearer.

Linhao-Qian: I'm really sorry for the confusion earlier, @philipp-classen 

After revisiting the implementation, I realized I previously misunderstood how Google Search Engine Redirect Protection works. Although the redirect URL format (https://www.google.com/url?q=https://www.example.com&sa=U&ved=2ahUKEwi&usg=AOvVaw0) is valid in current Google search engine, Google actually employs another redirect URL format as shown below:
```html
<a jsname="UWckNb" class="zReHs" href="https://openai.com/" 
   data-ved="2ahUKEwjah9TP6dGMAxXHAjQIHaLXDDEQFnoECBwQAQ" 
   ping="/url?sa=t&source=web&rct=j&opi=89978449&url=https://openai.com/&ved=2ahUKEwjah9TP6dGMAxXHAjQIHaLXDDEQFnoECBwQAQ">
</a>
``` 
Given this, I've removed the unnecessary code related to Google and streamlined the PR. I'd appreciate it if you could review the Bing Search Engine Redirect Protection feature at your convenience.

Thank you very much!

smalluban (approved):
Currently, it looks like the changes are additive to support Bing. LGTM.

PR #2348, opened 2025-04-04
fix(telemetry): add "Toolbar pinned" signal

### LINKED ISSUE #2342: Pull "isOnToolbar" from extension --> share of users with extension pinned
chrome.action.getUserSettings https://developer.chrome.com/docs/extensions/reference/api/action#method-getUserSettings

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/getUserSettings#isontoolbar

smalluban:
Fixes #2342

FYI: Firefox will not send signal, as we use manifest V2, and Safari does not support that feature at all.

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -70,7 +70,10 @@ const setup = asyncSetup('telemetry', [
       EXTENSION_VERSION: version,
       storage: metrics,
       saveStorage,
-      getConf: () => store.resolve(Options),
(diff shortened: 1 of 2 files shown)


chrmod (approved):

mdsandu:
for Safari I know, we didn't create a design for that. Did not make the connection V2/V3 re signal in FF but clear now. thanks

PR #2363, opened 2025-04-11
Address AMO review - fix trackers preview version

chrmod:


scripts/download-wtm-stats.js
```diff
@@ -13,7 +13,8 @@ import { writeFileSync, existsSync } from 'fs';
 import { resolve } from 'node:path';
 
 const TARGET_PATH = resolve('src', 'rule_resources', 'wtm-stats.js');
-const DATA_URL = 'https://whotracks.me/data/trackers-preview.json';
+const DATA_URL =
```
smalluban, line 16: We must add a script to update to the latest version within the `npm run release` script (as it also makes some changes to the code and commits them). Otherwise, in practice, it will never be updated again, and it will be out of date very soon. 

The commit hash should be placed in some config file, not a JS script (it should not be updated). As this now relies on the github repo, the easiest way would be to add `whotrackme/whotracks.me` GitHub repo to npm dependencies and take the file from there.

smalluban: And then in `npm run release` add `npm install whotracksme/whotracks.me#master` (or how it should be written, to update to the latest commit hash).

chrmod: Yes, that would be a next step. This change is meant to resolve the review blocker only.

chrmod: btw. you don't want to add `whotrackme/whotracks.me` as dependency as repo contains multiple large files.

chrmod: Issue to remember https://github.com/ghostery/ghostery-extension/issues/2364

smalluban: If you mean the next release, it won't be a problem, as this is not made at the end of the month. For the current one (10.4.29) there is nothing we can do, as the source is already pushed, and it will load the latest version of the file.

smalluban (approved):

PR #2391, opened 2025-04-25
feat: Element picker

### LINKED ISSUE #873: Add the ability to block an element
### Description

The ability to select and block a custom element on the page would be great, to either account for the ads not blocked by Ghostery by default, or to cosmetically remove those empty spaces where blocked ads used to be

smalluban:
Fixes #873 https://github.com/ghostery/collaboration-kalin/issues/10

## Testing

* Selecting elements lives in the page world - it would be great to test it on a variety of pages
* The feature logic touches the custom filter, so they should be tested to see if they work as expected


src/background/custom-filters.js
```diff
@@ -28,6 +28,7 @@ import Options from '/store/options.js';
 import CustomFilters from '/store/custom-filters.js';
 
 import { setup, reloadMainEngine } from '/background/adblocker.js';
+import CustomContentBlocks from '/store/custom-content-blocks';
```
chrmod, line 31: please add `.js` - can we have a linter rule for it?

src/background/custom-filters.js
```diff
@@ -160,6 +161,21 @@ async function updateEngine(text) {
...
+  Object.entries(customContentBlocks.selectors).forEach(
+    ([hostname, selectors]) => {
+      if (selectors) {
+        selectors.forEach((selector) => {
+          text += `\n${hostname}##${selector}`;
+          customContentBlocksFilters++;
+        });
+      }
+    },
+  );
```
chrmod, line 177: Bit more readable
```
const filters = Object.entries(customContentBlocks.selectors).reduce(
  (all, [hostname, selectors]) => [
    ...all,
    ...selectors.map((selector) => `${hostname}##${selector}`),
  ],
  []
);

```
or 
```
const filters = [];
for (const [hostname, selectors] of Object.entries(customContentBlocks.selectors)) {
  for (const selector of selectors) {
    filters.append(`${hostname}##${selector}`)
  }
}

text += filters.join(`\n`);
```

src/background/helpers.js
```diff
@@ -23,6 +23,24 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
     case 'openPrivateWindowWithUrl':
       chrome.windows.create({ url: msg.url, incognito: true });
       break;
+    case 'openElementPicker':
+      chrome.scripting.executeScript(
+        {
+          injectImmediately: true,
+          world: chrome.scripting.ExecutionWorld?.ISOLATED ?? 'ISOLATED',
```
chrmod, line 30: We had this pattern elsewhere, maybe worth to share the code?

src/store/custom-content-blocks.js
```diff
@@ -0,0 +1,43 @@
...
+import { store } from 'hybrids';
+
+const CustomContentBlocks = {
+  selectors: store.record([String]),
+  [store.connect]: {
+    get: async () => {
+      const { customContentBlocks = {} } = await chrome.storage.local.get([
+        'customContentBlocks',
+      ]);
+      return { selectors: customContentBlocks };
```
chrmod, line 21: would be nice if the key name indicate the we have a map from hostname to selectors. How about `hostnameSelectors` ? or maybe just `hostnames`?

src/store/custom-content-blocks.js
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { store } from 'hybrids';
+
+const CustomContentBlocks = {
```
chrmod, line 14: How about `ElementPickerSelectors` ?

chrmod (changes requested):

src/background/custom-filters.js
smalluban: It's my IDE, I need to check if I can force adding `.js` in the IDE, and will also look if there is an ESLint rule for that.

src/background/custom-filters.js
smalluban: You've changed the logic, the `selectors` variable can be `null`.

src/background/custom-filters.js
smalluban: Ok, my IDE has an option for that ;)

smalluban: For eslint we would need to add additional plugin, so I think it is not worth it.

src/background/helpers.js
smalluban: The util file for a function with one line only used in a specific case? :P Please no.

src/store/custom-content-blocks.js
smalluban: Ok, no strong opinion on that.

src/store/custom-content-blocks.js
smalluban: As the name of the model will be `ElementPickerSelectors` we can change to `hostnames` here.

chrmod (approved):

PR #2392, opened 2025-04-28
feat: scripting on response started

seia-soto:
This improves scripting on Chromium platforms by listening from both `webNavigation.onCommitted` and `webRequest.onResponseStarted`. In many times, especially, when the page is loaded from cache, `webRequest.onResponseStarted` fired in prior to `webNavigation.onCommitted`.

To be sure this doesn't perfectly solve late injection issue when you navigate from URL. When pasting URL to Omnibox, none of Tabs, Web Navigation, Web Request API properly fired right after the document creation.

With this change, users are expected problems to be solved after a few times of refresh if the cause was late injection. From testing myself, even one time of refresh was enough in most cases.

Besides the injection timing adjustment, we need to take care of hashing function. It's defined as `computeDocumentId` function. The hash IDs are expected to live extremely short as it's removed once both events are fired. However, to ensure less collinsion better, use of commonly known hash functions like crc32 is better. Since we're mixing two integers, none of 4 basic operations help. Alternatively, the below function can help us to transform number into the array of flatten decimal digits with appendix of the length of its string representation.

```
export function createHashableStream(a: number, b: number) {
  const arr: number[] = [];
  let c: number = 0;
  while (a > 0) {
    a = (a / 10) >>> 0;
    arr.push(a % 10);
  }
  arr.push(c);
  c = 0;
  while (b > 0) {
    b = (b / 10) >>> 0;
    arr.push(b % 10);
  }
  arr.push(c);
  return arr;
}
```

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -28,6 +28,7 @@ import asyncSetup from '/utils/setup.js';
 
 import { tabStats, updateTabStats } from './stats.js';
 import Config, {
+  FLAG_CHROMIUM_SCRIPTING_ONRESPONSESTARTED,
   FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
 } from '/store/config.js';
 
@@ -93,6 +94,23 @@ if (__PLATFORM__ === 'firefox') {
     }
   });
 }
+
+let ENABLE_CHROMIUM_SCRIPTING_ONRESPONSESTARTED = false;
+if (__PLATFORM__ === 'chromium') {
+  store.resolve(Config).then((config) => {
+    const enabled = config.hasFlag(FLAG_CHROMIUM_SCRIPTING_ONRESPONSESTARTED);
+    if (!enabled) contentScripts.unregisterAll();
+
+    ENABLE_CHROMIUM_SCRIPTING_ONRESPONSESTARTED = enabled;
+  });
+}
+
+// Use `documentId` field from Chromium `webRequest` and `webNavigation` events
+// to determine if cosmetic injection is already happened. Note that this is
+// not available on Firefox yet. If we need to listen from multiple events on
+// Firefox, we need to calculate a hash of tab and frame to generate `documentId`
+// on the fly.
+const executedDocumentIds = new Set();
 
 function getEnabledEngines(config) {
   if (config.terms) {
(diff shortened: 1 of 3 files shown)


src/background/adblocker.js
```diff
@@ -94,6 +95,57 @@ if (__PLATFORM__ === 'firefox') {
...
+const FNV_1A_OFFSET_BASIS = 2166136261;
+
+/**
+ * Computes a unique id corresponding to specific tab and frame.
+ * `documentId` may not available from `webRequest` event callbacks as the
+ * document may not present at the time of event occurrence.
+ * @param {number} tabId
+ * @param {number} frameId
+ */
+function computeDocumentId(tabId, frameId) {
```
seia-soto, line 126: This function can fail if `tabId` and `frameId` is identical.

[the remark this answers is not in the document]
seia-soto: fixed on [`ea10341` (#2392)](https://github.com/ghostery/ghostery-extension/pull/2392/commits/ea10341d6c9fc70e4111adaf5ee72ccddfa7d487)

src/background/adblocker.js
```diff
@@ -94,6 +95,37 @@ if (__PLATFORM__ === 'firefox') {
...
+
+// Use `documentId` field from Chromium `webRequest` and `webNavigation` events
+// to determine if cosmetic injection is already happened. Note that this is
+// not available on Firefox yet. If we need to listen from multiple events on
+// Firefox, we need to calculate a hash of tab and frame to generate `documentId`
+// on the fly.
+/**
+ * @type {Set<string>}
+ */
+const executedDocumentIds = new Set();
```
chrmod, line 116: This may not work. Does `executeScript` guarantee the scriptlet execution? What if document is not created yet?

I think we should try to inject many times, but be sure that the scriptlet is run only once. That may require so flag in the execution environment. But such flag should not be detectable by a page script as it could be used to detect the adblocker.

chrmod (changes requested):

src/background/adblocker.js
seia-soto: `executeScript` doesn't guarantee the script execution in the page. Also, the timing never matched by running `executeScript` continuously by using loop until they actually respond with results. I believe this is by the design of Chromium engine. Even they doesn't guarantee some time related function correctly.

In the given events, the documents are already all created. From the main frame type of on response started event, the document was always available. Also, navigation on committed event was not touched.

chrmod:
replaced by https://github.com/ghostery/ghostery-extension/pull/2455

PR #2413, opened 2025-05-07
fix(panel): pause state for global domain

smalluban:
Fixes the panel pause state for the case, when a user adds global domain by hand. The expected behavior is to show paused state for all subdomains.

## Testing

* Ensure that pausing by clicking on pause button works as before - adds proper full domain with subdomain and shows correct state
* Add global domain in settings by hand and check how the pause state is displayed in panel and "redo" action works (for example `onet.pl`, and go to `www.onet.pl`, or some subdomain like `kobieta.onet.pl`)

### DIFF
--- a/src/pages/panel/views/main.js
+++ b/src/pages/panel/views/main.js
@@ -60,8 +60,7 @@ async function togglePause(host, event) {
 
   if (paused) {
     const pausedHostname = Object.keys(options.paused)
-      .sort()
-      .reverse()
+      .sort((a, b) => b.localeCompare(a))
       .find((domain) => stats.hostname.endsWith(domain));
 
     store.set(options, {
--- a/src/pages/panel/views/main.js
+++ b/src/pages/panel/views/main.js
@@ -13,7 +13,7 @@ import { html, store, router, msg } from 'hybrids';
 
 import { getCurrentTab, openTabWithUrl } from '/utils/tabs.js';
 
-import Options, { GLOBAL_PAUSE_ID } from '/store/options.js';
+import Options, { getPausedDetails, GLOBAL_PAUSE_ID } from '/store/options.js';
 import TabStats from '/store/tab-stats.js';
 import * as exceptions from '/utils/exceptions.js';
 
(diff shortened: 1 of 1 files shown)


AdamGhst (approved):
Adding main domain (onet.pl) to allowlist causes subdomain to be paused too. 
Unpausing Ghostery in any subdomain (while main onet.pl is still allowlisted) makes the main domain (onet.pl) cleared out from the allowlist, which unpauses Ghostery for the whole onet.pl and its subdomains. 

Tested on macOS 
OPERA and FF.

src/pages/panel/views/main.js
```diff
@@ -53,17 +53,29 @@ function reloadTab(host, event) {
...
+  host.alert = '';
+
   const { paused, pauseType } = event.target;
+  const { options, stats } = host;
 
-  host.alert = '';
+  if (paused) {
+    const pausedHostname = Object.keys(options.paused)
+      .sort()
+      .reverse()
```
chrmod, line 64: no need to reverse if you change the sorting function

src/pages/panel/views/main.js
```diff
@@ -53,17 +53,29 @@ function reloadTab(host, event) {
...
+      .find((domain) => stats.hostname.endsWith(domain));
```
chrmod, line 65: why to sort and reverse if you do a find? pessimistically you have to iterate full list 3 times.

src/pages/panel/views/main.js
smalluban: Yes, but that is more readable - I don't like sorting function like `b - a`, which indicates nothing meaningful. The `sort` and `reverse` methods are highly optimized, and this is a callback of user action, so the performance of the execution is not an issue.

src/pages/panel/views/main.js
smalluban: We have to, as otherwise, if somehow a user adds `onet.pl` and `www.onet.pl`, and he is on `www.onet.pl`, we must remove the second entry from the paused list. However, it would be chosen if `onet.pl` is before.

chrmod (approved):

PR #2426, opened 2025-05-12
fix(build): add versioning for data dependencies

### LINKED ISSUE #2364: Version control data files
To answer AMO requirements for 100% reproducible builds we should introduce a lock file to fix version of statics assets like:
* Adblocker engines
* WTM statistics
* Anti-Tracking whitelists

Additionally, a it would be helpful to have a script to update those upon release automatically.

smalluban:
Fixes #2364.

* Adds `dataDependencies` custom field to `package.json` (it is a safe and recommended way) to keep versions of the data assets
* Updates the `release.sh` script to update data dependencies on each release 
* The anti-tracking whitelist has the potential to be added as well, but currently it is not being used at all (it is not a part of the build), so I skipped it
* The adblocker engines use a unified URL to the CDN, which does not provide versioning, so currently, there is no way to fix the version of the downloaded engines. However, from the perspective of the issue's goal, the AMO reviews did not complain about `.dat` files, only `.js` files that are imported and part of the executed code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,9 @@
     "plotly.js-basic-dist": "^3.0.1",
     "tldts-experimental": "^7.0.7"
   },
+  "dataDependencies": {
+    "wtm-stats": "ded8cb4a9bad0fa1fb0b60996382ba6527127c33"
+  },
   "engineStrict": true,
   "engines": {
     "npm": ">=9.6.7"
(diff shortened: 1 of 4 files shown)


chrmod (approved):
Looks good.

You can add adblocker resources to the `dataDependencies` - engines have urls that include checksums.

PR #2455, opened 2025-05-27
Adblocker: on chrome inject scriptlets faster

chrmod:


src/background/adblocker.js
```diff
@@ -425,6 +425,18 @@ chrome.webNavigation.onCommitted.addListener(
   { url: [{ urlPrefix: 'http://' }, { urlPrefix: 'https://' }] },
 );
 
+if (__PLATFORM__ === 'chromium') {
+  chrome.webRequest.onResponseStarted.addListener(
```
smalluban, line 429: Should we disable `L419` then on Chrome? It looks like it doubles the `injectCosmetics` on the `onCommited` event listener.

chrmod: no, we need both. there are valid naviagations that wont trigger onResponseStarted but will trigger onCommited

smalluban: Ok. However, should we flag it not to execute again, or do we not mind it?

chrmod: Currently we dont mind. This has to be tested

smalluban (approved):

src/background/adblocker.js
```diff
@@ -29,6 +29,7 @@ import asyncSetup from '/utils/setup.js';
 import { tabStats, updateTabStats } from './stats.js';
 import Config, {
   FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
+  FLAG_CHROMEIUM_INJECT_COSMETICS_ON_RESPONSE_STARTED,
```
smalluban: "CHROMEIUM" -> "CHROMIUM"

However, I am not sure why this is only for Chrome, and if the flag is meant to be turned on only on Chrome, we still need to configure it; the name does not help, but might be confusing.

smalluban (changes requested):

src/background/adblocker.js
smalluban: The `chrome.webRequest.onResponseStarted` API is available in Firefox as well. Why do we only target Chromium?

chrmod: `onResponseStarted` on Firefox acts too fast - the document does not exist yet so the scripts cannot be executed.

src/background/adblocker.js
```diff
@@ -556,3 +557,27 @@ if (__PLATFORM__ === 'firefox') {
     ['blocking', 'responseHeaders'],
   );
 }
+
+if (__PLATFORM__ === 'chromium') {
```
smalluban, line 561: There is no `chromium` platform.

smalluban (approved):

PR #2466, opened 2025-06-02
fix(element-picker): remove action on websites list

smalluban:
Fixes https://github.com/ghostery/ghostery-extension/pull/2456#pullrequestreview-2881407389


AdamGhst (approved):
All Purr-fect.
Tested on macOS and Win10. 
Chrome, Opera, Edge, FF,

PR #2474, opened 2025-06-05
fix(wdio): move to Chrome for Testing

smalluban:
Starting from v137, the offical Chrome builds deprecated `--load-extension` flag (https://support.google.com/chrome/a/answer/7679408?sjid=9819695566660205428-EU#chromeBrsrE137)

There is a new way to load extensions by `Extensions.loadUnpacked` CDP protocol, but it requires some flags, which currently don't work with WDIO (https://chromedevtools.github.io/devtools-protocol/tot/Extensions/#method-loadUnpacked)

> Developers can still use the --load-extension switch in non-branded builds such as Chromium and Chrome For Testing.

The workaround is to move to [Chrome for Testing](https://github.com/GoogleChromeLabs/chrome-for-testing).

### DIFF
--- a/tests/e2e/wdio.conf.js
+++ b/tests/e2e/wdio.conf.js
@@ -119,10 +119,12 @@ export const config = {
     },
     {
       browserName: 'chrome',
+      browserVersion: 'stable',
       'goog:chromeOptions': {
         args: (argv.debug ? [] : ['headless', 'disable-gpu']).concat([
           `--load-extension=${CHROME_PATH}`,
           '--accept-lang=en-GB',
+          '--no-sandbox',
         ]),
       },
     },


tests/e2e/wdio.conf.js
```diff
@@ -119,10 +119,12 @@ export const config = {
     },
     {
       browserName: 'chrome',
+      browserVersion: 'stable',
       'goog:chromeOptions': {
         args: (argv.debug ? [] : ['headless', 'disable-gpu']).concat([
           `--load-extension=${CHROME_PATH}`,
           '--accept-lang=en-GB',
+          '--no-sandbox',
```
smalluban, line 127: A fix by https://github.com/webdriverio/webdriverio/issues/14168, otherwise, we have an odd error:
> WebDriverError: session not created: probably user data directory is already in use

tests/e2e/wdio.conf.js
```diff
@@ -119,10 +119,12 @@ export const config = {
     },
     {
       browserName: 'chrome',
+      browserVersion: 'stable',
```
smalluban, line 122: Forces to use Chrome for Testing (it internally uses https://pptr.dev/browsers-api#cli, so we need to point to "stable"), otherwise it will try to use a local version of Chrome (which is the normal production channel).

PR #2489, opened 2025-06-13
fix(element-picker): separate engine for element picker selectors

smalluban:
* Moves out element picker selectors from custom filters to a separate engine
* Unifies last value variable names for observing changes in various models

I discovered that if there are a lot of custom filters, adding a single blocked element may block the browser UI, or at least be very slow. Before the change, it required recompiling the entire set of custom filters, which included network filters that had to be converted.

PR tested in Chrome and Firefox, and the migration process too. I recommend testing the custom filters and element picker feature in the next release.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -109,7 +109,11 @@ function getEnabledEngines(config) {
       list.push(engines.FIXES_ENGINE);
     }
 
-    list.push(engines.CUSTOM_ENGINE);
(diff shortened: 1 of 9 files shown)


chrmod (approved):
Overall looks correct. The unnecessary variable renaming made the PR harder to review.
Would be best to test it manually. @AdamGhst can you help here?

smalluban:
I merged the PR. I think the most efficient way will be to test it within the release.

PR #2501, opened 2025-06-24
fix(e2e): use Chrome Canary Testing channel

smalluban:
Github Actions setup uses local version of Chrome if the main version matches, but we need to force to use Testing version of Chrome.

The way is to use Canary channel, as is always has newer major version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


smalluban:
I figured out that the cause of the issue was a problem with loading the extension (broken DNR rule). However, the CI does not show any log that the extension wasn't loaded.

PR #2503, opened 2025-06-26
fix(utils): add support for Brave and Oculus in `browser-info` utility

smalluban:
### Summary
Updates the `/utils/browser-into.js` file to add support for Brave and Meta Quest Browser.

### New tokens
* Brave: `br`
* Meta Quest Oculus Browser: `oc`

### Highlights
* There were two different logics for detecting the browser name, one in the `getBrowserInfo()` function and another in the `getBrowserName()` function. The latter was created for UI purposes, but it might confuse, so I unified the source code to use the logic from the original `getBrowserInfo()` with added support for Brave and Meta.
* Methods and properties set on `cachedGetBrowserInfo` are not used. 
* I checked the usage of the `getRawBrowserInfo` - this was used only by `AliveMessageGenerator` from reporting, and it uses only the `browser`, `version`, and `os` properties (the `platform` field is not used). It also duplicates the logic for Ghostery browsers overwrites. We can safely pass the default function (`getBrowserInfo`). The only difference is with the `os` field, which now respects the `other` value, rather than passing whatever was there, which is better in my opinion.
* The `displayName` field is not used, so I removed it and introduced the `browser` field (for `AliveMessageGenerator` needs)

### Testing
I tested detection in Brave browser, and Chrome using the collected userAgent from the Quest 2 device.


src/utils/browser-info.js
```diff
@@ -14,39 +14,72 @@ import Bowser from 'bowser';
...
+
+    // INFO: bowser detects Oculus as `Chrome`,
+    // so we need to check for OculusBrowser specifically before Chrome
+    if (navigator.userAgent.includes('OculusBrowser')) {
+      return { browser: 'Oculus', name: 'oculus', token: 'oc' };
+    }
+
+    // Brave's user agent detects as `Chrome`,
+    // so we need to check for Brave specifically
+    if (navigator.brave?.isBrave) {
```
philipp-classen: We could move the checks ones that have the lowest risk of false-positive to the start.

For instance, if you get `navigator.brave?.isBrave` then should be an extremely precise check. So, it would make sense to try that first. Same for OculusBrowser browser.

Perhaps this order (trying to move the most specific and thus reliable checks to the front)?
* Brave
* Oculus
* Yandex
* Opera
* Edge
* Chrome

Also, for Edge you might have to check for "Edg/" (Edge is the old engine, I think). For instance, this is the user agent that I get on Edge:

```
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36 Edg/138.0.0.0
```

smalluban: OK, we can change the order. We are using the `bowser` package, which produces the full name of the browser, which uses an internally complex regexp. This is not a check on the raw userAgent.

smalluban: Yandex can be last, also Opera, as the check is done with `bowser` parsing result, which contains the proper name of the browser (the order does not matter). As we expect there are more Chrome users, we can order by chance of luck ;)

smalluban:
Tests will fail until the https://github.com/whotracksme/webextension-packages/pull/162 is merged and bumped.

PR #2518, opened 2025-07-03
Adblocker: extended selectors

chrmod:
Enable extended selectors but only behind the remote configuration flag.

Based on https://github.com/ghostery/adblocker/blob/master/packages/adblocker-webextension-cosmetics/src/index.ts

New adblocker version brings:
* support for native `:has`
* support ecss
* `:xpath`
* `:matches-css*`
* `:matches-path`
* `:upward`


src/content_scripts/adblocker/extended-selectors.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+import { querySelectorAll } from '@ghostery/adblocker-extended-selectors';
+
+let UPDATE_EXTENDED_TIMEOUT = null;
+const PENDING = new Set();
+const EXTENDED = [];
+const HIDDEN = new Map();
+
+function cachedQuerySelector(root, selector, cache) {
```
philipp-classen, line 19: What is the semantic of this function? If it is a straightforward cache, why does it do extra things like deduplicating and merging with previous, cached results?

Why not something straightforward like:

```
function cachedQuerySelector(root, selector, cache) {
  const entry = cache.get(root);
  const cachedElements = entry?.get(selector);
  if (cachedElements) {
    return cachedElements;
  }

  const selected = querySelectorAll(root, selector.ast);
  if (selector.attribute !== undefined) {
    if (entry) {
      entry = new Map();
      cache.set(root, entry);
    }
    entry.set(selector, selected);
  }
  return selected;
}
```

Also, not obvious to me why the `selector.attribute !== undefined` guard is needed. I kept it here, but it is also counter-intuitive for a caching function to filter.

chrmod: that code was copied from the adblocker-content project. Now replaced by direct import

philipp-classen: The closest that I see is in adblocker-webextension-cosmetics: https://github.com/ghostery/adblocker/blob/aebce5f04137f6a0ec12fc64e826f497e04c1cdf/packages/adblocker-webextension-cosmetics/src/index.ts#L74

But the semantic still looks weird to me. Because it is not just a cache, but doing extra things. For instance, not always caching and having a code path that merges sets. Even though now looking a it, could be that the merging set part is dead, because this check may always hold (cachedSelected === undefined):
https://github.com/ghostery/adblocker/blob/aebce5f04137f6a0ec12fc64e826f497e04c1cdf/packages/adblocker-webextension-cosmetics/src/index.ts#L96

is always true, meaning it always starts with an empty Set, so there is no merging. (But then the code would be expressed in a confusing way (instead it seems to effectively only copying the result set and caching that).

But the types in the other project help (so my draft code does not work, since the cache is two-levels deep):

  root: Element,
  selector: ExtendedSelector,
  cache: Map<Element, Map<ExtendedSelector, Set<Element>>>,
  
Why it does not cache if "selector.attribute === undefined", is also not obvious to me.

philipp-classen: But in other words, I wonder if the code (also in the other project) could not be simplified to something like that:
* assuming the interface can be changed (querySelectorAll never returns duplicates, so keeping the array looks reasonable)
* always caching results (not only partially for some inputs)
* removing the Set merging (assuming the path is dead)

```
function cachedQuerySelector(
  root: Element,
  selector: ExtendedSelector,
  cache: Map<Element, Map<ExtendedSelector, [Element]>>,
): [Element] {
  const cachedSelectors = cache.get(root);
  const cachedElements = cachedSelectors?.get(selector);
  if (cachedElements !== undefined) {
    return cachedElements;
  }

  const selected = querySelectorAll(root, selector.ast);
  if (cachedSelectors === undefined) {
    cachedSelectors = new Map();
    cache.set(root, cachedSelectors);
  }
  cachedSelectors.set(selector, selected);

  return selected;
}
```

chrmod: Perhaps. In this iteration I focuses on bringing back the eCSS functionality as it was tested back in a day. In next step I wish we remove the copied code and rely on the `@adblocker` package instead. From there, we should optimize the library code.

src/background/adblocker.js
```diff
@@ -30,6 +30,7 @@ import { tabStats, updateTabStats } from './stats.js';
 import Config, {
   FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
   FLAG_CHROMIUM_INJECT_COSMETICS_ON_RESPONSE_STARTED,
+  FLAG_ENABLE_EXTENDED_SELECTORS,
```
smalluban: The "enable" is obious, please shorten it to `FLAG_EXTENDED_SELECTORS`

src/background/adblocker.js
```diff
@@ -589,3 +598,9 @@ if (__PLATFORM__ === 'chromium') {
     { urls: ['http://*/*', 'https://*/*'] },
   );
 }
+
+let ENABLE_EXTENDED_SELECTORS = false;
```
smalluban: Move this code above L411, as it is used before it is defined

smalluban (changes requested):

src/content_scripts/adblocker/extended-selectors.js
philipp-classen: But wouldn't it make more sense to simplify this here (we have to test anyway)? That it was copied from another place doesn't proof that it is working here as well.

Unless I'm misreading it, there are two observations:

1) only partly caching the results ("if (selector.attribute !== undefined)") for no obvious reason is either dead code or a potential performance bug
2) same for the set merging come: either it is dead (most likely), or it has a really weird semantic (combining results of different "querySelectorAll" calls). If a caching function returns different results after flushing a cache, it is warning sign to me.

philipp-classen: Look it into again. I think, I can proof that it is dead and can be simplified at least to this version with steps that are all safe:

```
function cachedQuerySelector(root, selector, cache) {
  const cachedElements = cache.get(root)?.get(selector);
  if (cachedElements !== undefined) {
    return cachedElements;
  }
  const selected = new Set(querySelectorAll(root, selector.ast));

  if (selector.attribute !== undefined) {  // <-- remains (IMHO, should be removed, but would change the semantic)
    let cachedSelectors = cache.get(root);
    if (cachedSelectors === undefined) {
      cachedSelectors = new Map();
      cache.set(root, cachedSelectors);
    }
    cachedSelectors.set(selector, selected);
  }
  return selected;
}
```

It is always the version that I proposed. The only difference is that the version above also removes the "if" (to cache only parts of the results, which I think does not make sens) and does not convert querySelectorAll results to a set (because the original Array is already guaranteed to be free of duplicates, so I don't see the need to convert)

(This is the chain of transformations. Basically, it gets easier if you create an temporary if-else and copy code in two paths, then it becomes clearly on how to simplify:
https://gist.github.com/philipp-classen/40582f12f02844a1fb74f31aa5df8ff9

philipp-classen: To sum up:
* We used the simplified version below now (which should be semantically indistinguishable)
* For the other changes, which both would be changing the semantics, we are pushing it them back to follow-ups (likely in the adblocker repository):
  - removing the extra Set wrapping would change the interface (and would result in adjustments to all callers)
  - switching from caching the partial function to the complete function may introduce side-effects like memory leaks; until we understand *why* these are not cached by the old implementation, always recomputing will be at least correct

PR #2528, opened 2025-07-08
fix(safari): match min Safari version with target deployment

smalluban:
Currently, we target ios 18+ and macos 13.5+ (as the oldest macOS version, which supports Safari 18). We should align the min Safari version in the manifest with the targets.


chrmod (approved):
iOS 17 has around 5% market share and it is shrinking. Given the amount of fixes for webextension in iOS 18 - it is worth to focus on the latest platform

Issue #2532, opened 2025-07-09
Sync settings made by Distraction Eraser


szl1973:
It would be great, if the website settings (made by distraction eraser) could be synced between devices.

chrmod:
Synchronization of website settings is something we are thinking about. It is sadly non-trivial due to privacy implication. Ghostery does not want to get domain names the users have custom filters for. So to sync, we look to implement end to end encryption with a user password, so that settings, when stored on Ghostery servers wont be readable for Ghostery.

smalluban:
The current way to synchronize is to manually download a backup and apply it on another browser. The fix for distraction eraser filters has been added to `v10.5.0`. 

@szl1973 Feel free to re-open if you have any other issues with the subject.

PR #2536, opened 2025-07-10
fix(backup): open exported config in the new tab on iOS

### LINKED ISSUE #2537: Safari on iOS fails to export extension settings
In the Safari, when I try to export extension settings into a file, I get the following error:

Image

smalluban:
Fixes #2537
Relates to #2532

Safari on iOS has a bug that prevents downloading the generated `Blob` object. To mitigate the issue, the only way is to open a new tab and display the content. Then, the user might copy the text and save it (however, this is not trivial). Unfortunately, there is no better way.

When doing the research, I have found a helpful library `file-saver`, which does not fix the iOS issue, but it is valuable to avoid other edge cases. I replaced our custom code with a call to the `saveAs` function from the library.

Additionally, the PR adds support for element picker selectors in the backup file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@whotracksme/reporting": "^7.4.0",
     "apexsankey": "github:ghostery/apexsankey",
     "bowser": "^2.11.0",
+    "file-saver": "^2.0.5",
     "hybrids": "^9.1.18",
     "idb": "^8.0.2",
     "jwt-decode": "^4.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
     "@whotracksme/reporting": "^7.4.0",
     "apexsankey": "github:ghostery/apexsankey",
     "bowser": "^2.11.0",
+    "file-saver": "^2.0.5",
```
chrmod, line 54: This library is very small and not maintained in years. Maybe better to copy the relevant source?

smalluban: Still it has millions of downloads, and hard to say what part will help. As you wrote, the library is small, so it won't impact much. It is imported only in settings page.

PR #2538, opened 2025-07-11
feat(onboarding): add notification for pinning extension in the toolbar

smalluban:
Fixes https://github.com/ghostery/collaboration-kalin/issues/34

Zrzut ekranu 2025-07-11 o 10 21 31


src/pages/notifications/pin-it.js
```diff
@@ -0,0 +1,78 @@
...
+switch (getBrowser().name) {
+  case 'opera':
+    imageUrl = pinExtensionOpera;
+    pathname = `/ghostery-ad-blocker-opera#how-do-i-pin-ghostery-to-the-opera-toolbar`;
+    break;
+  case 'edge':
+    imageUrl = pinExtensionEdge;
+    pathname = `/ghostery-ad-blocker-edge#how-do-i-pin-ghostery-to-the-microsoft-edge-toolbar`;
+    break;
+  default:
```
chrmod, line 37: what about Brave? I guess we should not display notification on that platform

src/pages/notifications/pin-it.js
```diff
@@ -0,0 +1,78 @@
...
+
+import Options from '/store/options.js';
+
+import * as notifications from '/utils/notifications.js';
+import { HOME_PAGE_URL } from '/utils/urls.js';
+import { getBrowser } from '/utils/browser-info.js';
+
+import pinExtensionChrome from './assets/pin-extension-chrome.png';
+import pinExtensionEdge from './assets/pin-extension-edge.png';
+import pinExtensionOpera from './assets/pin-extension-opera.png';
```
chrmod: what will this do? convert to base64? please use chrome.runtime.getURL instead

src/background/pin-it.js
```diff
@@ -0,0 +1,49 @@
...
+      // The notification was already shown maximum times
+      onboarding.pinIt >= NOTIFICATION_SHOW_LIMIT ||
+      // The notification was already shown recently
+      (onboarding.pinItAt &&
+        Date.now() - onboarding.pinItAt < NOTIFICATION_DELAY)
+    ) {
+      return false;
+    }
+
+    openNotification({
```
chrmod, line 45: what if another notification is already shown?

src/pages/notifications/pin-it.js
smalluban: No, the import has a URL to the file. This is a common pattern in bundlers, as the file is moved to the assets directory when building.

src/pages/notifications/pin-it.js
smalluban: Brave works pretty the same as Chrome
Zrzut ekranu 2025-07-14 o 09 45 15

src/background/pin-it.js
smalluban: The content script prevents displaying more than one notification at a time. However, I added a quick fix to what's new, as its version for notification was set regardless of the actual display of the notification.

src/pages/notifications/pin-it.js
chrmod: would be nice to call the variable in a way that would indicate we deal with URL. For example `chromeImageUrl`

src/pages/notifications/pin-it.js
chrmod: yes, but the screenshot does not match, right?

src/pages/notifications/pin-it.html
```diff
@@ -0,0 +1,15 @@
...
+<html data-theme="light">
+  <head>
+    <title>Pin it</title>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width" />
+    <link rel="icon" type="image/png" href="/ui/assets/favicon.png" />
+    <link rel="icon" type="image/svg+xml" href="/ui/assets/favicon.svg" />
+    <link rel="preload" as="image" href="./assets/pin-extension-chrome.png" />
+    <link rel="preload" as="image" href="./assets/pin-extension-edge.png" />
+    <link rel="preload" as="image" href="./assets/pin-extension-opera.png" />
```
chrmod, line 11: why to preload images that we don't use?

chrmod (approved):
Added a few comments but those are nit-picks. I'm good to merge it as is

src/pages/notifications/pin-it.html
smalluban: It's better than preloading nothing ;) There is no reliable way to preload only the required image (the `__PLATFORM__` works only in JS land).
However, preloading prevents the notification layout from being bumped.

src/pages/notifications/pin-it.js
smalluban: It's almost the same. The icon is the same, the message is identical. Of course we can add another image, but I think the current image is good enough.

PR #2565, opened 2025-07-22
fix(whats-new): Set current version after installation

smalluban:
To prevent displaying "What's new" to users who clear storage at the end of the session, we need to set the current version if it's not already set. 

IMPORTANT TESTING NOTICE: With this change, it won't be possible to test What's new notification/page by restarting the browser after installation.  To trigger it, the extension must change its minor version, eg, from 10.5.x to 10.6.x. It must be changed manually in the `manifest.json` file in the file system where the unpacked extension is loaded in the browser.

@AdamGhst, please update the test case for What's new with the above instruction.

### DIFF
--- a/src/background/whats-new.js
+++ b/src/background/whats-new.js
@@ -34,6 +34,13 @@ chrome.runtime.onStartup.addListener(async () => {
     return;
   }
 
+  // After installing the extension version is 0,
+  // so we need to set it to the current version and return early
+  if (options.whatsNewVersion === 0) {
+    store.set(options, { whatsNewVersion });
+    return;
+  }
+
   const tabs = await chrome.tabs.query({ currentWindow: true });
   const activeTab = tabs.find((tab) => tab.active);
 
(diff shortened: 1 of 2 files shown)


src/background/whats-new.js
```diff
@@ -34,6 +34,13 @@ chrome.runtime.onStartup.addListener(async () => {
     return;
   }
 
+  // After installing the extension version is 0,
+  // so we need to set it to the current version and return early
+  if (options.whatsNewVersion === 0) {
```
philipp-classen, line 39: Can you say, why it starts with 0? Is it a default that is set somewhere?

philipp-classen: I think, I found it here: https://github.com/ghostery/ghostery-extension/blob/a87196329210249b5442ddaf8988a26585778ced/src/store/options.js#L108

smalluban: Yes, it's set in options definition.

philipp-classen (approved):

PR #2470, opened 2025-06-03
Adblocker: local DNR converter

chrmod:

PR #2568, opened 2025-07-22
fix(dnr): use local DNR converter for network filters

### LINKED ISSUE #2556: Local DNR converter

smalluban:
Fixes #2556 and replaces #2470


scripts/build.js
```diff
@@ -432,7 +418,7 @@ const buildPromise = build({
             .replace('_virtual', 'virtual');
 
           const path = name.replace(pwd, '');
-          if (path.length > 100 && !argv['no-filename-limit']) {
```
chrmod, line 435: scary

src/background/exceptions.js
```diff
@@ -73,44 +65,27 @@ async function updateFilters() {
...
-          }
+    const result = await convert(filters);
 
-          rules.push({
-            ...rule,
-            priority: 2000000 + rule.priority,
-          });
+    for (const rule of result.rules) {
+      if (domains && domains.length) {
+        if (__PLATFORM__ === 'safari') {
```
chrmod, line 74: shouldn't we use convertToSafariFormat here?

chrmod (approved):
Generally looks great. I'm only little worried that on Safari it runs the converter in the service worker.

scripts/build.js
smalluban: There is no other way, as we don't control third-party dependencies. There should be fewer of them with better tree-shaking of the convert library.

src/background/exceptions.js
smalluban: It is now in the main convert utility - both checks if regex is supported and converts to Safari format. https://github.com/ghostery/ghostery-extension/pull/2568/files#diff-872e57c910d6cc8e3ca8a688cd4fdc3aa1e864a0092b218a09403859a090424aR35

src/pages/dnr-converter/index.js
```diff
@@ -0,0 +1,39 @@
...
+// Overwrite the declarativeNetRequest monkey patch for adguard library
+// The check is redone in the background script
+import './monkey-patch.js';
+
+import { convertWithAdguard } from '@ghostery/urlfilter2dnr';
+
+export async function convert(filters) {
+  try {
+    const result = await convertWithAdguard(filters);
+    result.errors.map((e) => `DNR: ${e.message}`);
```
philipp-classen: Isn't that a no-op?

Did you mean:

result.errors.map = result.errors.map(...)

src/pages/dnr-converter/index.js
```diff
@@ -0,0 +1,39 @@
...
+    return {
+      rules: [],
+      errors: [err],
+    };
+  }
+}
+
+chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
+  if (msg.action === 'dnr-converter:convert') {
+    convert(msg.filters).then((result) => sendResponse(result));
```
philipp-classen: Should we handle errors? Can the receiver get stuck if the operation does not send a response?

src/utils/dnr-converter.js
```diff
@@ -9,131 +9,88 @@
...
-    });
-
-    const iframe = document.createElement('iframe');
-    iframe.setAttribute('src', 'https://ghostery.github.io/urlfilter2dnr/');
-    iframe.setAttribute('style', 'display: none;');
-
-    documentConverter = new Promise((resolve, reject) => {
-      iframe.addEventListener('load', () => resolve(iframe));
-      iframe.addEventListener('error', reject);
+      closeOffscreenDocument();
```
philipp-classen: Would it be safer to call it in a finally block?

```
await setupOffscreenDocument();
try {
  result = await chrome.runtime.sendMessage(...)
} finally {
  closeOffscreenDocument();
}
```

src/pages/dnr-converter/index.js
smalluban: Yes, but now I think we don't need that. I will clear it

src/pages/dnr-converter/index.js
smalluban: It was like that before, but good point, we can add response in catch callback

src/utils/dnr-converter.js
```diff
@@ -9,131 +9,88 @@
...
+      }
+    }
   }
 
-  let requestCount = 0;
-
-  return async function convert(filter) {
-    const iframe = await createIframe();
-    const requestId = `request${requestCount++}`;
+  return result || { errors: ['Failed to initiate converter'], rules: [] };
```
philipp-classen: Looking at the function, "result" should be always truthy at this point (e.g. otherwise, the for loop should also throw when it runs "result.rules.entries"()"

Thus, I would think the "|| { errors ... }" fallback is dead.

src/utils/dnr-converter.js
smalluban: Hm.. good point too :)

src/utils/dnr-converter.js
```diff
@@ -9,131 +9,88 @@
...
-        {
-          action: 'convert',
-          converter: 'adguard',
-          filters: [requestId, filter],
-        },
-        '*',
-      );
-    });
-  };
+  offscreenTimeout = setTimeout(() => {
```
philipp-classen, line 74: Just for understanding: why delay the close operation? Would it cause problems to do "chrome.offscreen.closeDocument()" immediately? If not, can there be races if the one second is not enough?

Or is it intended to keep it open if there is a chance that multiple conversions will be triggered in short order? In other words, as an optimization to avoid multiple startup costs of the offscreen document?

smalluban: There is only one trigger to open offscreen document from the background script. The delay is to prevent to much open/close operations. If someone would add and remove tracker exceptions very fast, it might cause opening offscreen document, that was just closed.

philipp-classen: Makes sense. Could be worth adding this explanation as a comment.

PR #2593, opened 2025-07-30
feat(dnr): use dynamic DNR for fixes filters

### LINKED ISSUE #2476: Update DNR rules for `fix` filter list
We should fetch and convert fixes on the fly and apply them to the scope of dynamic rules. Currently, we are shipping them statically with a release.

smalluban:
Fixes #2476

Adds runtime fetch of the fixes DNR rules. 

For the compatibility layer, it falls back to the static rules `dnr-fixes.json` if the processing URL fails.

~~As there is no other way to detect if the dynamic fixes are applied, the code checks if there are any rules from the range. It means that the update/fetch procedure would run every time options are updated if there are no rules at all (I assumed this would never happen).~~

BIG UPDATE

* The `Resources` model has been added to keep `checksums` record (can keep more metadata in the future). 
* The checksum is also saved for every adblocker engine
* The code checks if the checksum is there, or if it has changed
* The broken page report options dump has been updated with added checksums and flags. From now on, the options only contain changed values (defaults are omitted).

### DIFF
--- a/src/background/dnr.js
+++ b/src/background/dnr.js
@@ -43,7 +43,7 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
       );
     }
 
-    // Add fixes rules
+    // Add latest fixes rules
     if (ids.length) {
       const fixesRulesIds = await getDynamicRulesIds(FIXES_ID_RANGE);
 
--- a/src/background/dnr.js
+++ b/src/background/dnr.js
@@ -10,7 +10,10 @@
  */
 
 import { ENGINES, getPausedDetails } from '/store/options.js';
+import { convertToSafariFormat } from '/utils/dnr-converter-safari.js';
+import { FIXES_ID_RANGE, getDynamicRulesIds } from '/utils/dnr.js';
 import * as OptionsObserver from '/utils/options-observer.js';
+import { ENGINE_CONFIGS_ROOT_URL } from '/utils/urls.js';
 
 if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
   const DNR_RESOURCES = chrome.runtime
@@ -21,28 +24,86 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
   // Ensure that DNR rulesets are equal to those from options.
   // eg. when web extension updates, the rulesets are reset
   // to the value from the manifest.
-  OptionsObserver.addListener(async function dnr(options) {
+  OptionsObserver.addListener(async function dnr(options, lastOptions) {
     const globalPause = getPausedDetails(options);
 
     const ids = ENGINES.map(({ name, key }) => {
       return !globalPause && options.terms && options[key] ? name : '';
     }).filter((id) => id && DNR_RESOURCES.includes(id));
 
-    if (ids.length) {
-      ids.push('fixes');
-
-      // Add regional network filters
-      if (options.regionalFilters.enabled) {
-        ids.push(
-          ...options.regionalFilters.regions
-            .map((id) => `lang-${id}`)
-            .filter((id) => DNR_RESOURCES.includes(id)),
-        );
-      }
-    }
-
     const enabledRulesetIds =
       (await chrome.declarativeNetRequest.getEnabledRulesets()) || [];
+
+    // Add regional network filters
+    if (ids.length && options.regionalFilters.enabled) {
+      ids.push(
+        ...options.regionalFilters.regions
+          .map((id) => `lang-${id}`)
+          .filter((id) => DNR_RESOURCES.includes(id)),
+      );
+    }
+
+    // Add fixes rules
+    if (ids.length) {
+      const fixesRulesIds = await getDynamicRulesIds(FIXES_ID_RANGE);
+
+      // Fixes should be set when engines are re-enabled and after each update.
+      // We can detect re-enabling by checking if both `fixes` static rules
+      // and dynamic fixes rules are not present.
+      if (
+        (!enabledRulesetIds.includes('fixes') && fixesRulesIds.length === 0) ||
+        lastOptions?.filtersUpdatedAt < options.filtersUpdatedAt
+      ) {
+        try {
+          await chrome.declarativeNetRequest.updateDynamicRules({
+            removeRuleIds: fixesRulesIds,
+          });
+
+          let addRules = await fetch(
+            `${ENGINE_CONFIGS_ROOT_URL}/dnr-fixes-v2/allowed-lists.json`,
+          )
+            .then((res) => res.json())
+            .then((list) => fetch(list.dnr.url))
+            .then((res) => res.json());
+
+          if (__PLATFORM__ === 'safari') {
+            addRules = addRules.reduce((acc, rule) => {
+              try {
+                acc.push(convertToSafariFormat(rule));
+              } catch {
+                // Ignore rules that cannot be converted
+              }
+              return acc;
+            }, []);
+          }
+
+          await chrome.declarativeNetRequest.updateDynamicRules({
+            addRules: addRules.map((rule, index) => ({
+              ...rule,
+              id: FIXES_ID_RANGE.start + index,
+            })),
+          });
+
+          console.info('[dnr] Updated dynamic fixes rules');
+        } catch (e) {
+          console.error('[dnr] Error while updating dynamic fixes rules:', e);
+
+          // As a fallback, add the static fixes rules from the resources
+          ids.push('fixes');
+        }
+      }
+    } else {
+      if (!enabledRulesetIds.includes('fixes')) {
+        // Remove all dynamic fixes rules
+        const removeRuleIds = await getDynamicRulesIds(FIXES_ID_RANGE);
+        if (removeRuleIds.length) {
+          await chrome.declarativeNetRequest.updateDynamicRules({
+            removeRuleIds,
+          });
+          console.info('[dnr] Removed dynamic fixes rules');
+        }
+      }
+    }
 
     const enableRulesetIds = [];
     const disableRulesetIds = [];
@@ -65,9 +126,9 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
           enableRulesetIds,
           disableRulesetIds,
         });
-        console.info('[dnr] Updated with rulesets:', ids.join(', '));
+        console.info('[dnr] Updated static rulesets:', ids.join(', '));
       } catch (e) {
-        console.error(`[dnr] Error while updating rulesets:`, e);
+        console.error(`[dnr] Error while updating static rulesets:`, e);
       }
     }
   });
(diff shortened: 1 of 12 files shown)


src/utils/dnr.js
```diff
@@ -0,0 +1,24 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+export const PAUSED_ID_RANGE = { start: 1, end: 1_000_000 };
+export const CUSTOM_FILTERS_ID_RANGE = { start: 1_000_000, end: 2_000_000 };
+export const EXCEPTIONS_ID_RANGE = { start: 2_000_000, end: 3_000_000 };
+export const FIXES_ID_RANGE = { start: 3_000_000, end: 4_000_000 };
+
+export const PAUSED_RULE_PRIORITY = 10_000_000;
+export const EXCEPTIONS_RULE_PRIORITY = 2_000_000;
```
smalluban, line 18: This comes from the current code. The rules are using the start index for the priority level. I expect that we can change that to another value, but it relies on the maximum priority level of "normal" rules.

In short, it is the same as it was before.

src/background/dnr.js
```diff
@@ -21,29 +24,87 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
...
+      if (
+        (!enabledRulesetIds.includes('fixes') && fixesRulesIds.length === 0) ||
+        lastOptions?.filtersUpdatedAt < options.filtersUpdatedAt
+      ) {
+        try {
+          await chrome.declarativeNetRequest.updateDynamicRules({
+            removeRuleIds: fixesRulesIds,
+          });
+
+          let addRules = await fetch(
```
chrmod: what if this fails?

smalluban: Then there is a `catch` block...

smalluban:
> After first update of fixes, we should probably disable the fixes ruleset.

This is exactly what will happen. The static rulesets are "set" every call, then we check if they are already enabled. With this code, if everything goes right with dynamic rules, the `fixes` static ruleset won't be added to `ids`, so either it won't be added, or it will be removed.

src/background/dnr.js
```diff
@@ -21,29 +24,87 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
...
-      ids.push('fixes');
-
-      // Add regional network filters
-      if (options.regionalFilters.enabled) {
-        ids.push(
-          ...options.regionalFilters.regions
-            .map((id) => `lang-${id}`)
-            .filter((id) => DNR_RESOURCES.includes(id)),
-        );
+      const fixesRulesIds = await getDynamicRulesIds(FIXES_ID_RANGE);
```
chrmod: this run on every options update?

It should definitely not run if we already have rules that are up to date. We should also have a schedule to update fixes every hour, same as we do with engines.

src/background/dnr.js
```diff
@@ -21,29 +24,87 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
...
+    // Add regional network filters
+    if (ids.length && options.regionalFilters.enabled) {
+      ids.push(
+        ...options.regionalFilters.regions
+          .map((id) => `lang-${id}`)
+          .filter((id) => DNR_RESOURCES.includes(id)),
+      );
+    }
+
+    // Add fixes rules
```
chrmod:
```suggestion
    // Add latest fixes rules
```

chrmod (changes requested):

src/background/dnr.js
smalluban: There is no other way to know if dynamic rules are there. This code must run on disable/enable engines operation, which requires to be in options observer.

src/background/dnr.js
```diff
@@ -21,29 +24,87 @@ if (__PLATFORM__ === 'chromium' || __PLATFORM__ === 'safari') {
...
+      if (
+        (!enabledRulesetIds.includes('fixes') && fixesRulesIds.length === 0) ||
+        lastOptions?.filtersUpdatedAt < options.filtersUpdatedAt
+      ) {
+        try {
+          await chrome.declarativeNetRequest.updateDynamicRules({
+            removeRuleIds: fixesRulesIds,
+          });
+
+          let addRules = await fetch(
```
philipp-classen: Generally, it is a good idea to still check for "response.ok", just to get some better error messages.

Currently, it can result in misleading errors like "Uncaught TypeError: Cannot read properties of undefined" or "Uncaught SyntaxError: Unexpected token".

src/background/dnr.js
chrmod: I've meant the full operation. Specifically the download.

chrmod: you can keep a flag if there was an update - you will probably need one to prevent updating too often

smalluban: The full operation is done only when it is really required - if you remove dynamic rules, you must download them again (re-enabling), or when engines have been updated (once an hour of active usage of the extension).

smalluban: See this line, read comment
https://github.com/ghostery/ghostery-extension/pull/2593/files#diff-32fa3e38e24d85bd0f82578c7083f3c30d8642c1e65b42f5c7fab1511a17bbbeR53

src/background/dnr.js
```diff
@@ -10,40 +10,129 @@
...
     .declarative_net_request.rule_resources.filter(({ enabled }) => !enabled)
     .map(({ id }) => id);
 
+  function getIds(options) {
+    if (!options.terms || getPausedDetails(options)) return [];
+
+    return ENGINES.reduce((acc, { name, key }) => {
+      if (options[key] && DNR_RESOURCES.includes(name)) acc.push(name);
+      return acc;
+    }, []);
```
chrmod:
```suggestion
    return ENGINES
      .filter(e => options[e.key] && DNR_RESOURCES.includes(e.name))
      .map(e => e.name);
```

src/background/dnr.js
```diff
@@ -10,40 +10,129 @@
...
+          console.info('[dnr] Updated dynamic fixes rules');
+        } catch (e) {
+          console.error('[dnr] Error while updating dynamic fixes rules:', e);
+
+          // As a fallback, add the static fixes rules from the resources
+          ids.push('fixes');
+        }
+      }
+    } else {
+      if (!enabledRulesetIds.includes('fixes')) {
```
chrmod: why not else if ?

src/background/dnr.js
```diff
@@ -10,40 +10,129 @@
...
+        // Remove all dynamic fixes rules
+        const removeRuleIds = await getDynamicRulesIds(FIXES_ID_RANGE);
+        if (removeRuleIds.length) {
```
chrmod, line 149: is this if needed?

src/background/dnr.js
```diff
@@ -10,40 +10,129 @@
...
+      ids.push(
+        ...options.regionalFilters.regions
+          .map((id) => `lang-${id}`)
+          .filter((id) => DNR_RESOURCES.includes(id)),
+      );
+    }
+
+    // Add latest fixes rules
+    if (ids.length) {
+      const fixesRulesIds = await getDynamicRulesIds(FIXES_ID_RANGE);
```
philipp-classen: Since I expect this to be a rather expensive operation, changing the order of the checks could save unnecessary computations:

Looking at
```
(!enabledRulesetIds.includes('fixes') && fixesRulesIds.length === 0) ||
        lastOptions?.filtersUpdatedAt < options.filtersUpdatedAt
```

and matching it as `(X && Y) || Z` only Y depends on fixesRulesIds.

(X && Y) || Z
<=>
Z || (X && Y)

And it is always false for every Y if !Z && !X

In other words, we need
```
const fixesRulesIds = await getDynamicRulesIds(FIXES_ID_RANGE);
```

only for Z || X, so replacing back

```
lastOptions?.filtersUpdatedAt < options.filtersUpdatedAt || !enabledRulesetIds.includes('fixes')
```

Something like like perhaps:

```
const force = lastOptions?.filtersUpdatedAt < options.filtersUpdatedAt;
if (force || !enabledRulesetIds.includes('fixes'))
  const fixesRulesIds = await getDynamicRulesIds(FIXES_ID_RANGE);
  if (fixesRulesIds.length === 0 || force) {
    try {
      await chrome.declarativeNetRequest.updateDynamicRules({
       removeRuleIds: fixesRulesIds,
      });
    ...
  }
}
```

Hope my math is correct. But basically, the idea is to evaluate "getDynamicRulesIds" only if really needed, and exhaust all other inexpensive checks before.

smalluban: Thanks for this deep understanding. I think I could write a comment, that we are going to introduce checksums, which saved can be checked instead of calling `getDynamicRulesIds`. This condition will change.

src/background/dnr.js
smalluban: If you mean the condition, yes, as there can be a case when we want to clear DNR rules, but nothing is already in place. I am not sure, but the `updateDynamicRules` can even throw an error if the list of IDs is empty.

smalluban:
The PR has been updated with support for checksums. The description has been updated. 

@chrmod @philipp-classen Please do another round of review.

src/utils/engines.js
```diff
@@ -282,9 +289,12 @@ export async function update(name) {
 
       // Save the new engine to memory and storage
       saveToMemory(name, engine);
-      saveToStorage(name);
+      saveToStorage(name, data.engines[ENGINE_VERSION].checksum);
```
philipp-classen, line 292: Should this have error handling? Also, if all other usage of saveToStorage just continue, but log errors, would introducing a "trySaveToStorage" function make sense? Which handles the logging, but is fire-and-forget, providing no guarantees beside best-effort.

src/utils/engines.js
```diff
@@ -385,10 +395,13 @@ export async function update(name) {
...
     if (updated) {
-      console.info(`[engines] Engine "${name}" updated`);
+      console.info(
+        `[engines] Engine "${name}" updated:`,
+        data.engines[ENGINE_VERSION].checksum,
+      );
 
       // Save the new engine to storage
-      saveToStorage(name);
+      saveToStorage(name, data.engines[ENGINE_VERSION].checksum);
```
philipp-classen, line 404: Same here. Could be switched to a "trySaveToStorage", which handles logging.

src/utils/engines.js
smalluban: It has, but inside the `saveToStorage` (try/catch). The name follows the convention of other targets (`saveToMemory`, `saveTo...`, `loadFrom....`). Your proposition might be valid, but I would keep this PR focused on the DNR fixes.

src/background/dnr.js
```diff
@@ -9,41 +9,158 @@
...
+          });
+          await store.set(resources, {
+            checksums: { [DNR_FIXES_KEY]: null },
+          });
+
+          console.info('[dnr] Removed dynamic fixes rules');
+        }
+      }
+    } else if (ids.length) {
+      ids.push('fixes');
```
chrmod, line 161: having a note that this is a default behavior for the first run may help in the future

chrmod (approved):

src/background/dnr.js
smalluban: It won't be if we set a flag to 100%.

PR #2603, opened 2025-08-04
feat(onboarding):  installation survey for Brave

smalluban:
Fixes https://github.com/ghostery/journal/issues/547

I tried a number of different ways, but the `onbeforeunload` event only allows opening a prompt. Messages to the background script are not sent (the tab is closed before the event can go out). We also cannot rely on the `tabId` returned when we open onboarding, as it might be closed and open again.

After all, the simplest and reliable way is to listen for `terms` change and removal of tabs.

### DIFF
--- a/src/background/onboarding.js
+++ b/src/background/onboarding.js
@@ -33,18 +33,7 @@ if (__PLATFORM__ === 'chromium' && getBrowser().name === 'brave') {
     async function braveSurvey(terms, lastTerms) {
       // Trigger only when user just have accepted terms
       if (lastTerms === false && terms === true) {
-        const [tab] = await chrome.tabs.query({
-          url: chrome.runtime.getURL('/pages/onboarding/index.html'),
-        });
-
-        const callback = async (tabId) => {
-          if (tabId === tab?.id) {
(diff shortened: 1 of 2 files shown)


mdsandu:
@smalluban the survey is not triggering on brave but "pin it" is at first website visit - and wow! that would be a lot of attention we ask of the user. we should disable pin it for brave for the time of survey 😬  (somehow I did not have brave as a browser where we trigger "pin it" top of mind as we did not mention it on the list)

used Brave 1.80.125/ Chromium 138.0.7204.184

smalluban:
Thanks for real user tests ;) If you keep open extension's dev tools, everything works as expected ;) I'll try to fix it, but it looks like opening a survey on closing the onboarding tab has too many edge cases (like if you wait 30 seconds and SW is killed).

smalluban:
We decided to go with a simpler trigger, just after clicking "Enable". Then, a user can close the survey and be back on the "Success page" of the onboarding.

mdsandu (approved):
survey trigger when clicking "enable" an moves user to that tab, setup successful opens in a separate tab during this. 
pin it does not trigger anymore

chrmod (approved):

PR #2628, opened 2025-08-13
feat(panel): add never-consent feedback

smalluban:


src/content_scripts/autoconsent.js
```diff
@@ -11,11 +11,38 @@
 
 import AutoConsent from '@duckduckgo/autoconsent';
 
+const FLAG_STORAGE_KEY = 'ghostery:autoconsent';
```
philipp-classen: One problem that I see is that it can be used for adblocker detection (by anti-adblockers) and for fingerprinting (trackers).

I'm not aware of a straightforward technique to transfer information in a safe and non-detectable way. So, I cannot offer an constructive advice. Message passing might be the best choice.

smalluban: That's a good point, I didn't take into account. Ill search if there is another way to store information in sync with browsing history of the user.

smalluban:
I moved the storage to the internal one using the `Resources` model. I realized that keeping the flag, even though the CMP can be shown again, is not completely wrong. If a user clears browsing history (or cookies expire), the never-consent will do it again, so the information is still valid. Most importantly, the information is no longer stored in the page world.

Currently, we are not showing that information, but I thought that instead of keeping the boolean, it would be better to save the timestamp. We can add clearing very old information, if we would like to avoid making this data too big or something else.

I also added a condition to hide the feedback if the extension is paused or never-consent is disabled.

src/background/autoconsent.js
```diff
@@ -94,6 +95,15 @@ chrome.runtime.onMessage.addListener((msg, sender) => {
       return initialize(msg, sender.tab, frameId);
     case 'eval':
       return evalCode(msg.snippetId, msg.id, sender.tab.id, frameId);
+    case 'optOutResult': {
+      if (msg.result === true) {
+        const { domain, hostname } = parse(sender.url);
+        store.set(Resources, {
+          autoconsent: { [domain || hostname]: Date.now() },
```
philipp-classen: Do you have an example when "domain" is missing, but a "hostname" exists? Also, I wonder if it could happen that both are missing? If it may happen, perhaps skipping the store operation, but printing a warning?

chrmod: yeah, I think better to skip if it is something weird

smalluban: It was meant to cover cases like IP addresses, or "weird" URLs :P However, this only runs if the autoconsent triggers, so we can assume that there will be a domain (I will add a condition for that, but it will only set for domains).

Issue #2630, opened 2025-08-14
Redirect resources path broken


seia-soto:
Chrome: 139.0.7258.66
Ghostery Browser Extension: 10.5.4

When resource redirect rule added, the path is not resolved correctly: `chrome-extension://ID/prefix/empty`

<img width="461" height="287" alt="Image" src="https://github.com/user-attachments/assets/34bf1c8c-7286-44a8-bf39-da88961ae70f" />

1. Open `https://www.nespresso.com/us/en/vertuo-coffee-pods`
2. Add `||akstat.io^$domain=nespresso.com,redirect=noopjs`
3. Check devtools network panel resulting network error

PR #2631, opened 2025-08-14
fix: feed redirect resource path option

seia-soto:
fixes https://github.com/ghostery/ghostery-extension/issues/2630

`urlfilter2dnr`'s `convertWithAdguard` requires `opts.resourcesPath` to be provided to update the resources' path for DNR.
The options are highly likely to be shared but I put in both places as it's not expected to be updated often.


smalluban (approved):
LGTM. However, I wonder if we were using `resourcesPath` before, as I don't remember we were sending this to the iframe ;)

PR #2661, opened 2025-08-27
fix: clean accounts

smalluban:
* Cleans up `session` and `api` calls
* Refactors sync to use `chrome.storage.sync` where it is fully supported
* Cleans up UI: contributor links, login links, etc..

The changes are spread across individual commits.


src/background/devtools.js
```diff
@@ -27,6 +27,9 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
           console.info('[devtools] Clearing main local storage');
           chrome.storage.local.clear();
 
+          console.info('[devtools] Clearing sync storage');
+          chrome.storage.sync.clear();
```
philipp-classen, line 31: It is an debug option, but might be still unexpected. Maybe this can be guarded (clicking multiple times or something like that?).

smalluban: After clicking once "Clear storage" you are instantly redirect to the onboarding, as you have no longer terms accepted. I am not sure if it even possible to click this more than once, as the button is disabled until the message from background process comes back to the UI.

src/utils/opera.js
```diff
@@ -33,13 +36,18 @@ export async function isSerpSupported() {
...
+        value: 'true',
+        domain: COOKIE_DOMAIN,
+        expirationDate: Date.now() / 1000 + 60 * 60 * 24 * 365 * 10,
+        httpOnly: true,
+      });
 
       isSupported = true;
     } catch {
-      // Clear out the cookie if it was set
-      setCookie('opera_serp_notification', undefined);
```
chrmod, line 41: why this was removed?

src/store/options.js
```diff
@@ -156,6 +145,14 @@ const Options = {
...
 };
 
+export const SYNC_OPTIONS = Object.keys(Options).filter(
+  (key) => !LOCAL_OPTIONS.includes(key),
+);
+
+export const REPORT_OPTIONS = SYNC_OPTIONS.filter(
+  (key) => !PROTECTED_OPTIONS.includes(key),
+).concat(['filtersUpdatedAt']);
+
```
chrmod: This is little hard to follow. Maybe we could use `Set.prototype.difference` to deal with that?

chrmod (changes requested):

src/utils/opera.js
smalluban: This cannot happen - look what can cause `catch` - this code would only run if the cookie wasn't set correctly - so there is no need to clear it out.

src/store/options.js
smalluban: It is just filtering out protected options and adding `filersUpdatedAt`, but I can use the spread operator; it might be more readable. This issue is related to formatting.

They are arrays, and `Set.prototype.difference` is available in Firefox from 127. We can't use it.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference#browser_compatibility

chrmod (approved):

smalluban:
@AdamGhst Please test this PR before merging. It would be best to do all smoke tests on one selected platform, can be Chrome. And 1-2 tests on the rest Safari will come with the release).

AdamGhst:
Tested on macOS and Win
Chrome - full smokes. 
FF - main functions. 

Works like a charm.

PR #2674, opened 2025-09-01
feat(dnr): use general rulesets in Safari

chrmod:


smalluban:
V10.5.8 (98) pushed to TestFlight.

smalluban:
I strongly recommend to merge this PR after v10.5.8 release. It will contain accounts clean up, and the last thing we want is to lock most of the users with them still visible in the extension.

smalluban (approved):
Tested on iOS 18.6.2 (iPhone XR and iPad Pro).

PR #2680, opened 2025-09-02
fix(engines): lazy load serialized engines from CDN

### LINKED ISSUE #2677: Initialize adblocker engines after installation
We no longer need to attach serialized engines to the distribution. They are obsolete at the moment of the public release, as it takes some time to be reviewed by the platform. In most cases, they must be updated immediately. Also, it makes the build not reproducible after a day. It is not possible to re-create the same distribution from the source code. 

The only drawback is the requirement for a user to fetch whole engines (not diffs) after installation.

smalluban:
Fixes #2677

Tested update from current production on Firefox and Chrome. Needs update tests for Safari on release PR.

### DIFF
--- a/src/utils/engines.js
+++ b/src/utils/engines.js
@@ -204,6 +204,8 @@ export async function update(name, { force = false } = {}) {
   // If the IndexedDB is corrupted, and there is no way to load the engine
   // from the storage, we should skip the update.
   // It can also happen if the engine has not finished init.
+  // The `force` option allows us to bypass this check, as when
+  // engine is being loaded from CDN for the first time the `loadFromStorage` returns null
   if (!force && (await loadFromStorage(name)) === null) {
     console.warn(
       `[engines] Skipping update for engine "${name}" as the engine is not available`,


src/pages/onboarding/views/main.js
```diff
@@ -26,7 +26,13 @@ const TERMS_AND_CONDITIONS_URL = `https://www.${GHOSTERY_DOMAIN}/privacy/ghoster
 function acceptTerms(host, event) {
   router.resolve(
     event,
-    store.set(Options, { terms: true, feedback: host.feedback }),
+    store.set(Options, {
+      terms: true,
+      feedback: host.feedback,
+      // Engines are initialized for the first time from CDN after onboarding,
+      // so we need to update the timestamp to current time
+      filtersUpdatedAt: Date.now(),
```
smalluban, line 34: This prevents running the update process just after loading the engine from the CDN. However, if the fetch from CDN fails, it will be repeated instantly next run of the SW. This value does not block it, as `loadFromStorage` will return `null` in `init()` function.

src/utils/engines.js
```diff
@@ -216,11 +200,11 @@ function check(response) {
   return response;
 }
 
-export async function update(name) {
+export async function update(name, { force = false } = {}) {
   // If the IndexedDB is corrupted, and there is no way to load the engine
   // from the storage, we should skip the update.
   // It can also happen if the engine has not finished init.
-  if ((await loadFromStorage(name)) === null) {
+  if (!force && (await loadFromStorage(name)) === null) {
```
chrmod, line 209: Can you document the force logic. By reading the code it is not obvious

smalluban: Sure, added.

chrmod (approved):

PR #2699, opened 2025-09-08
fix(exceptions): improve domain toggle logic for exceptions

### LINKED ISSUE #2664: Can't turn off GTM any more
I can't seem to turn off GTM any more in Firefox through the Ghostry dev menu

Image

The toggle switch toggels but as you can see from the screen shot it never turns off. (Note I know what turning off does I'm trying to turn it of during development rather than having to turn if off in the network tab of dev tools which is one of the main reasons for me to have ghostry)

I am using Ghostry 10.5.6.

smalluban:
Fixes #2664. When toggling the exception on the subdomain, and only the top-level domain is set, it should be used.

## Testing

Example of top-level domain without `www` and subdomain: `wordpress.org` and `learn.wordpress.org`. 

Add an exception for GTM on the top level, open subdomain, and try to toggle the same exception. The status should change to "Blocked on all websites".

### DIFF
--- a/src/utils/exceptions.js
+++ b/src/utils/exceptions.js
@@ -57,8 +57,12 @@ export function toggleDomain(options, trackerId, domain, force = false) {
 
   let domains = [domain];
   if (exception) {
-    domains = exception.domains.includes(domain)
-      ? exception.domains.filter((d) => d !== domain || force)
+    const targetDomain = exception.domains.includes(domain)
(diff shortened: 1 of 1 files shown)


chrmod (approved):

chrmod:
Thank you for the report. Let me move the issue to a repository that we use to deal with broken pages.

PR #2712, opened 2025-09-15
fix(managed): paused rules & minor clean up

smalluban:
* Fixes issue with updating a list of paused domains in managed mode, when `disableUserControl` is not turned on. The PR changes the condition to check the existence of `trustedDomains` in managed mode instead of relying on `disableUserControl`.
* Sets options related to managed config, rather than relying on config directly - much cleaner code
* Removes redundant action described in #2713

REBASE on merge

### DIFF
--- a/src/background/paused.js
+++ b/src/background/paused.js
@@ -77,8 +77,7 @@ OptionsObserver.addListener('paused', async (paused, lastPaused) => {
     (lastPaused ||
       // Managed mode can update the rules at any time - so we need to update
       // the rules even if the paused state hasn't changed
-      (__PLATFORM__ !== 'firefox' &&
-        (await store.resolve(ManagedConfig)).disableUserControl))
+      (await store.resolve(ManagedConfig)).trustedDomains.length > 0)
   ) {
     const removeRuleIds = await getDynamicRulesIds(PAUSED_ID_RANGE);
     const hostnames = Object.keys(paused);
--- a/src/store/options.js
+++ b/src/store/options.js
@@ -228,6 +228,14 @@ async function manage(options) {
     }
   }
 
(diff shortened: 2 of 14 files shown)


src/background/paused.js
```diff
@@ -77,8 +77,7 @@ OptionsObserver.addListener('paused', async (paused, lastPaused) => {
     (lastPaused ||
       // Managed mode can update the rules at any time - so we need to update
       // the rules even if the paused state hasn't changed
-      (__PLATFORM__ !== 'firefox' &&
-        (await store.resolve(ManagedConfig)).disableUserControl))
+      (await store.resolve(ManagedConfig)).trustedDomains.length > 0)
```
smalluban, line 80: It is not a 100% correct solution, as if the config will update from a list of domains to an empty list, it won't update. The correct solution requires calling `getDynamicRules()` (returns approximately 2k rules) and finding the rule for each SW start.

Still, I think it is good enough change - as if the `trustedDomains` are used, they should contain at least one domain preselected to be paused.

src/store/options.js
```diff
@@ -216,16 +212,14 @@ async function manage(options) {
 
     // Clear out the paused state, to overwrite with the current managed state
     options.paused = {};
+  }
 
-    // Add paused domains from the config
-    const config = await store.resolve(Config);
-    if (config.hasFlag(FLAG_PAUSE_ASSISTANT)) {
```
chrmod, line 222: why this is removed?

smalluban: I explained in the description, and in the linked PR - https://github.com/ghostery/ghostery-extension/pull/2713

chrmod (approved):

PR #2713, opened 2025-09-15
fix(pause-assistant): remove redudant addition of paused domains

smalluban:
I've found redundant logic - leftover from the initial logic of the Pause Assistant - It did not pause the domain upfront at that time.

In current logic, this line https://github.com/ghostery/ghostery-extension/blob/main/src/background/pause-assistant.js#L42 adds the domain from the remote config upfront regardless of the state of the `ManagedConfig`.

### DIFF
--- a/src/background/config.js
+++ b/src/background/config.js
@@ -12,8 +12,6 @@
 import { store } from 'hybrids';
 
 import Config from '/store/config.js';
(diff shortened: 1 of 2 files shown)


smalluban:
The commit cherry-picked to #2712

PR #2716, opened 2025-09-18
feat: support `scriptletGlobals.warOrigin`

seia-soto:
fixes https://github.com/ghostery/broken-page-reports/issues/1718
depends on https://github.com/ghostery/scriptlets/pull/26

This supports the injection of `scriptletGlobals` for scriptlets depending on redirect resources of the extension packages. See https://github.com/ghostery/scriptlets/blob/main/ubo.js#L19175-L19192 for the reference and check `ghostery/scriptlet` repository to align the changes.

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -267,7 +267,13 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
     }
 
     const func = scriptlet.func;
-    const args = parsed.args.map((arg) => decodeURIComponent(arg));
+    const scriptletGlobals = {
+      warOrigin: chrome.runtime.getURL('/rule_resources/redirects'),
+    };
+    const args = [
+      scriptletGlobals,
+      ...parsed.args.map((arg) => decodeURIComponent(arg)),
+    ];
 
     if (
       __PLATFORM__ === 'firefox' &&
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -268,7 +268,12 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
 
     const func = scriptlet.func;
     const scriptletGlobals = {
-      warOrigin: chrome.runtime.getURL('/rule_resources/redirects'),
+      // Request a real extension resource to obtain a dynamic URL to the resource.
+      // Redirect resources are defined with `use_dynamic_url` restriction.
+      // refs https://developer.chrome.com/docs/extensions/reference/manifest/web-accessible-resources#manifest_declaration
+      warOrigin: chrome.runtime
+        .getURL('/rule_resources/redirects/empty')
+        .slice(0, -6),
     };
     const args = [
       scriptletGlobals,
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -36,7 +36,7 @@ import Config, {
 
 let options = Options;
 
-const scriptletsGlobal = {
+const scriptletGlobals = {
   // Request a real extension resource to obtain a dynamic ID to the resource.
   // Redirect resources are defined with `use_dynamic_url` restriction.
   // The dynamic ID is generated per session.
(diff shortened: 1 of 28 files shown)


seia-soto:
Updated the lockfile, it includes:

refs https://github.com/ghostery/scriptlets/pull/27
refs https://github.com/ghostery/scriptlets/pull/26

seia-soto:
Corresponding test added and fixed quirks; ready for the final review.

smalluban:
https://github.com/ghostery/ghostery-extension/pull/2736 should fix the issue with creating a package.

src/background/adblocker.js
```diff
@@ -267,7 +267,18 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
     }
 
     const func = scriptlet.func;
-    const args = parsed.args.map((arg) => decodeURIComponent(arg));
+    const scriptletGlobals = {
+      // Request a real extension resource to obtain a dynamic URL to the resource.
+      // Redirect resources are defined with `use_dynamic_url` restriction.
+      // refs https://developer.chrome.com/docs/extensions/reference/manifest/web-accessible-resources#manifest_declaration
+      warOrigin: chrome.runtime
```
smalluban: Can the detection of the base URL be performed once per service worker lifetime and saved for subsequent calls to `injectScriptlets`?

tests/e2e/page.html
```diff
@@ -20,5 +20,16 @@ <h1>Test Page</h1>
       id="custom-filter"
       style="display: block; height: 100px; background: blue"
     ></div>
+
+    <!-- Scriptlets with web accessible resources -->
+    <script defer>
+      void (async function () {
+        const res = await fetch('ads.js');
+        const div = document.createElement('div');
```
smalluban, line 28: The div can be static and put in HTML code, or is it important to be added dynamically?

smalluban:
@seia-soto You can merge main or rebase. The PR with a fix is merged.

src/background/adblocker.js
```diff
@@ -267,7 +267,18 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
...
+        .getURL('/rule_resources/redirects/empty')
+        .slice(0, -6),
```
philipp-classen: I assume cutting off the last six characters is because you want to have the results for

`chrome.runtime.getURL('/rule_resources/redirects')`

And "/empty" is six characters. Though on the first glance, it is not obvious to me why the step is needed (appending "/empty" to have an existing one and then cutting it off). As I understand, chrome.runtime.getURL only constructs the URL but does not attempt to fetch it.

Anyway, if it is needed, it would be good to document that part.

smalluban: It looks like it just generates a URL, and it does not matter if the URL exists, so we can avoid `slice`. 

Zrzut ekranu 2025-09-25 o 14 20 33

seia-soto: I've already tried that and I found that it won't generate a dynamic ID in case the requested file path is not defined or found. Your URL indicates that that's our public extension ID which is trackable and not dynamic: `https://chromewebstore.google.com/detail/ghostery-tracker-ad-block/mlomiejdfkolichcflejclcbmpeaniij?hl=en`.

seia-soto: Also, the comment suggests it: `// Request a real extension resource to obtain a dynamic URL to the resource.`. The change was made here: https://github.com/ghostery/ghostery-extension/pull/2716/commits/fcbc8ece87885a369aaa434899037de0fd4c9287#diff-6481a5f7409b68532da99ab1713c40379d7978ca76596daff9237a31eb667543L271-R273

tests/e2e/page.html
seia-soto: I believe it's a reliable and clean way to notify that the request is finished and to deliver the response body. This way, we can just use `waitFor` instead of relying on complex function calls.

src/background/adblocker.js
seia-soto: I believe this can be cached:

> A dynamic ID is generated per session. That means it is regenerated when the browser restarts or the extension reloads.
>
> refs https://developer.chrome.com/docs/extensions/reference/manifest/web-accessible-resources

seia-soto: Updated by 3340820212c68f1aa8385012e344aba6ec48be23

PR #2719, opened 2025-09-19
fix(element-picker): force reload selectors when engine version mismatch

smalluban:
This PR fixes the issue when adblocker library updates, so loading the element picker engine trhows an error about version mistach. 

We have to force reload the engine if this happens.

I tested the change by updating production build with older version of adblocker by `npm run start:update chromium v10.5.6`. 

As the result, the selector runs correctly with the fix, but it does not if we update without it.

### DIFF
--- a/src/background/element-picker.js
+++ b/src/background/element-picker.js
@@ -19,9 +19,17 @@ import { setup, reloadMainEngine } from './adblocker.js';
 
 // Observe element picker selectors to update the adblocker engine
 store.observe(ElementPickerSelectors, async (_, model, lastModel) => {
-  if (!lastModel) return;
-
   let entries = Object.entries(model.hostnames);
+
+  // Skip update if there is no change in selectors
+  // and the engine is initialized if there are selectors.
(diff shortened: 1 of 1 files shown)


src/background/element-picker.js
```diff
@@ -19,10 +19,18 @@ import { setup, reloadMainEngine } from './adblocker.js';
...
-  if (!lastModel) return;
-
   let entries = Object.entries(model.hostnames);
 
+  // Skip update if there is no change in selectors
+  // and the engine is initialized if there are selectors.
+  // The engine can be `null` if adblocker detects version mismatch.
+  if (
+    !lastModel &&
+    (entries.length ? await engines.init(engines.ELEMENT_PICKER_ENGINE) : true)
```
chrmod: can we move the `init` outside of if statement? having nested ifs won't harm but it may be easier to understand that we take an action conditionally

smalluban: Updated

PR #2729, opened 2025-09-23
chore(environment): update nodejs and github node and checkout actions

smalluban:
Periodical update of environment tooling.


.github/workflows/artifact-upload.yml
```diff
@@ -24,11 +24,9 @@ jobs:
 
           echo $VERSION
           echo $SHA
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v5
         with:
           node-version-file: '.tool-versions'
```
smalluban, line 29: It is done automatically in `v5` - https://github.com/actions/setup-node?tab=readme-ov-file#breaking-changes-in-v5

PR #2738, opened 2025-09-25
feat(panel): edge mobile notification

smalluban:
Fixes https://github.com/ghostery/journal/issues/582

Zrzut ekranu 2025-09-26 o 09 20 01


smalluban:
@mdsandu Please test it out on Edge desktop.

scripts/generate-qr.js
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env node
+
+import fs from 'fs';
+import QRCode from 'qrcode';
```
philipp-classen: Do we need to install the dependency somehow?

If it is a script that is rarely used, we could put it in a separate folder with a package.json, or comment at least on the top, what external dependencies need to be installed.

src/pages/panel/store/notification.js
```diff
@@ -34,32 +39,41 @@ const NOTIFICATIONS = {
...
+  img: '',
   type: '',
   text: '',
   url: '',
   action: '',
   [store.connect]: async () => {
     const { terms, panel } = await store.resolve(Options);
 
+    // return NOTIFICATIONS.edgeMobile;
+
```
philipp-classen:
```suggestion
    // return NOTIFICATIONS.edgeMobile;

```

scripts/generate-qr.js
smalluban: Ah, it should not be there at all. I thought I removed it from this branch.

philipp-classen (approved):

PR #2754, opened 2025-09-30
fix(panel): update report an issue flow

smalluban:
Fixes https://github.com/ghostery/collaboration-kalin/issues/52


mdsandu:
@smalluban 
- [ ] If you chose a category, unselect screenshot and change category from the dropdown “windows OS looking lis - grey) the screenshot remains unselected - pls default to selected 


- [ ] This does not look like on the design - validation step is missing (alert in red) 
Bal

- [ ] It takes a long while for the report to be sent and it feels as if the Ghostery UI froze

smalluban:
> If you chose a category, unselect screenshot and change category from the dropdown “windows OS looking lis - grey) the screenshot remains unselected - pls default to selected

Can you write it in more than one sentence? I cannot understand what you mean.

> This does not look like on the design - validation step is missing (alert in red)

Yes. We use the native `required` attribute, and it matches OS requirements (it is rendered by the browser). It is tricky to make it look custom, but this is edge case. Also, it is not worth to implement custom behavior here - it will take to much time. Adam designed that, as it was required by the logic. I expect he don't mind. 

>  It takes a long while for the report to be sent and it feels as if the Ghostery UI froze

This is how it works - the delay only depends on sending a request to ghostery.com - I expect uploading the screenshot may take some time (it also depends on your internet connection).

If there is something I should change, I'll make another PR with fixes.

mdsandu:
-->  I expect he don't mind.- I mind because I spent time thinking about it and it's just dismissed without making any comment

--> Can you write it in more than one sentence? I cannot understand what you mean. --> visuals make it easier to understand - just try the path, you understand than what I mean. thanks

Screenshot 2025-10-01 at 10 24 45
Screenshot 2025-10-01 at 10 24 54
Screenshot 2025-10-01 at 10 25 00
Screenshot 2025-10-01 at 10 25 08
Screenshot 2025-10-01 at 10 25 47
Screenshot 2025-10-01 at 10 26 25

smalluban:
It is one form; if the user unselects screenshot and then changes category, it can't suddenly change another part of the form.  The user may not notice the change and send us a screenshot in a situation where they didn't want it.

And it is an edge case, where you first change the screenshot field, and then change the category, as we have created a separate step to be sure that the category is chosen correctly.

PR #2804, opened 2025-10-17
feat(pause-assistant): panel & settings page UI updates

smalluban:
Fixes https://github.com/ghostery/collaboration-kalin/issues/36

### Highlights

* Updates to the panel when the page is paused by the assistant
* Updates to the websites section of the settings page with a new badge and info on the website details
* Fixes issue with resolved BPR (removal of the domain from the config list)
* Proper clean up of domains when a feature flag or browsing assistant in settings is turned off
* Added `webkit` platform for remote config `filter` field
* Minor fixes for syncing remote config

### Testing

* Enable `pause-assistant` flag in settings page (dev tools). It should immediately add domains from the config (if we have them there for the current platform)
* Remove flag or disable/enable browser assistant in "Whotracks.me" section to check clean up domains
* Click on the domain name in dev tools to test the end of breakage, as the domain would be removed from the config
* To add a custom domain, use the `Test domain` action in dev tools
* `Force sync` action clears out flags, so after clicking it, the feature will be disabled (the flag is removed)

Zrzut ekranu 2025-10-20 o 09 02 51
Zrzut ekranu 2025-10-20 o 09 02 58
Zrzut ekranu 2025-10-20 o 09 03 21
Zrzut ekranu 2025-10-20 o 09 03 15

### DIFF
--- a/src/pages/settings/components/devtools.js
+++ b/src/pages/settings/components/devtools.js
@@ -98,17 +98,16 @@ async function testConfigFlag(host) {
   );
 
   await setConfig({
-    flags: flags.split(',').reduce(
-      (acc, flag) => {
-        flag = flag.trim();
-        if (flag) acc[flag] = { enabled: true };
-        return acc;
-      },
-      Object.keys(host.config.flags).reduce((acc, flag) => {
-        acc[flag] = { enabled: false };
-        return acc;
-      }, {}),
-    ),
+    flags: flags
+      .split(',')
+      .map((f) => f.trim())
+      .filter(Boolean)
+      .reduce(
+        (acc, f) => ((acc[f] = { enabled: true }), acc),
+        Object.fromEntries(
+          Object.keys(host.config.flags).map((k) => [k, { enabled: false }]),
+        ),
+      ),
   });
 }
 


src/pages/settings/components/devtools.js
```diff
@@ -92,16 +96,28 @@ async function testConfigFlag(host) {
...
     ].join(', '),
   );
-  if (!flags) return;
 
-  await store.set(host.config, {
-    flags: flags.split(',').reduce((acc, flag) => {
-      acc[flag.trim()] = { enabled: true };
-      return acc;
-    }, {}),
+  await setConfig({
```
chrmod, line 100: how about:
```js
const enabledFlags = new Set(
  flags.split(',').map(f => f.trim()).filter(Boolean)
);

const allFlags = Object.fromEntries(
  Object.keys(host.config.flags).map(flag => [
    flag,
    { enabled: enabledFlags.has(flag) },
  ])
);

await setConfig({ flags: allFlags });
```

src/background/pause-assistant.js
```diff
@@ -18,29 +18,65 @@ import Config, {
...
 
-  if (paused) {
+  if (Object.keys(paused).length) {
     store.set(Options, { paused });
   }
+}
+
+// Update paused domains when the config changes
+store.observe(Config, async (_, config, lastConfig) => {
+  updatePausedDomains(config, lastConfig);
```
chrmod, line 72: should we return early if we config was changed in a way that does not affect paused domains?

src/pages/settings/components/devtools.js
smalluban: I don't mind refactoring, but with this change, you cannot test new local flags, which you want to add to the config later (when you develop the feature).
The L107+ is for ensuring that not set flags (but set the remote config) will be set to `false`. There was no way to test removing the flag from the config if the remote config already enabled it.

src/background/pause-assistant.js
smalluban: The logic to return early is inside `updatePauseDomains()`.

AdamGhst (approved):
Awesome simple designs and flawless engineering. 

All perfect.

src/pages/settings/components/devtools.js
chrmod: that was not clear, perhaps we need a comment. 

Here is alternative implementation that should work the same:
```js
await setConfig({
  flags: flags
    .split(',')
    .map(f => f.trim())
    .filter(Boolean)
    .reduce(
      (acc, f) => ((acc[f] = { enabled: true }), acc),
      Object.fromEntries(
        Object.keys(host.config.flags).map(k => [k, { enabled: false }])
      )
    ),
});
```

chrmod (approved):

mdsandu:
the UI copy (panel + everything newly added associated with Bob) needs review

mdsandu:
screen time : Browsing Assistant 

title: Broken page report 
Desc: View a detailed report showing which ad blocker functionality may have caused the issue. 

title: How Browsing Assistant works 
Desc: Learn how Ghostery is automatically paused to avoid breakage.

PR #2815, opened 2025-10-24
fix(platform): update platform checks to improve code tree-shaking

smalluban:
The PR udpates conditions where we can first check the `__PLATFORM__` flag value to properly tree-shake code. 

There is also a fix for `sync.js` and related settings page to match the same condition (`isWebkit()` rather than `isSafari()`).

### DIFF
--- a/src/background/config.js
+++ b/src/background/config.js
@@ -28,7 +28,7 @@ function filter(item) {
     // Possible values: 'chromium', 'firefox', 'webkit'
     if (check && Array.isArray(platform)) {
       check = platform.includes(
-        __PLATFORM__ === 'chromium' && isWebkit() ? 'webkit' : __PLATFORM__,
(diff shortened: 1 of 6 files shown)


src/background/config.js
```diff
@@ -28,7 +28,7 @@ function filter(item) {
     // Possible values: 'chromium', 'firefox', 'webkit'
     if (check && Array.isArray(platform)) {
       check = platform.includes(
-        __PLATFORM__ === 'chromium' && isWebkit() ? 'webkit' : __PLATFORM__,
+        __PLATFORM__ !== 'firefox' && isWebkit() ? 'webkit' : __PLATFORM__,
```
smalluban, line 31: This is a semantic change, but we can completely move from value-based flag to `__FIREFOX__` as we now have only two build-based platforms - `chromium` and `firefox`.

PR #2834, opened 2025-11-05
chore(tests): add e2e job without feature flags for `package` labeled PRs

smalluban:
The PR adds a job to run end-to-end tests without flags when the "package" label is present. It should be sufficient to run those tests on extension using a package label, like we do with each release pull request. Currently, most flags are set to 100% (we are keeping the flags for safety reasons). 

* Added a new `e2e-package` job that runs Chrome and Firefox end-to-end tests when a pull request has the "package" label. This job depends on the completion of the existing `e2e` and `e2e-update` jobs and only runs for relevant PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    types: [opened, synchronize, reopened, labeled]
   workflow_dispatch:
     inputs:
       flags:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    types: [opened, synchronize, reopened, labeled]
```
philipp-classen, line 8: Not sure about the reason for the "labeled" part; but if labels are affecting the behavior of the tests, maybe we should also include "unlabeled" (removing a label)?

(Maybe it is sufficient as is, since it only cares about the presence of the "package" tag. But just asking to clarify.)

smalluban: I followed the workflow for building the package (it uses the `package` label as the condition): https://github.com/ghostery/ghostery-extension/blob/main/.github/workflows/artifact-upload.yml#L5

Also, I am not sure if there is an `unlabeled` type, but it does not make sense, as we want to trigger this workflow only if you add a label, not when you remove it.

Issue #2836, opened 2025-11-05
Use `documentId` for injecting cosmetics


smalluban:
When using the `onResponseStarted` event, injecting cosmetics can be too early, as the active frame/tabId points to the leaving document. On Chromium, we must use `documentId` to point to the correct target of the injection.

<img width="1502" height="294" alt="Image" src="https://github.com/user-attachments/assets/adf816de-8915-430a-8310-1c599caaaa58" />

Steps to reproduce (by @seia-soto):

1. Set chromium-inject-cosmetics-on-response-started flag in Ghostery (from Ghostery settings, dev tools)
2. Restart the browser
3. Open https://seia-soto.github.io/checkadb/index.html and apply custom filters from "scripting"
4. Check for the service worker log

smalluban:
When implementing the fix by replacing the `frameIds` option of `chrome.scripting` API with `documentIds` (where it is available), I discovered that `webRequest.onResponseStarted` or even `webRequest.onCompleted` does not include the `documentId` option, if the event fires before creation of the document.

How can it work now? The issue with using `frameIds` only occurs when the user navigates from one page to another, and the second page is not loaded from cache (the request for html body). Then running `chrome.scripting.executeScript()` or `chrome.scripting.insertCSS()` with a tabId/frameId pair fails, as it targets the wrong document (using `webRequest` events).

This issue might still apply to using `webNavigation.onCommited`, but I expect it is rare (I can't reproduce it just using this API). However, I would fix it by utilizing `documentId` where it is possible, regardless of the used APIs.

The example:

The event callback `details` data when loading `page.localhost:6789`.

The `details` on `webRequest.onCompleted()` (web page request):

<img width="458" height="341" alt="Image" src="https://github.com/user-attachments/assets/7d4e07e6-0890-4448-a4ea-0f856dd6f12e" />

As you can see, there is no `documentId` option. The document has not yet been created. 

Then we have `webNavigation.onCommited()` event:

<img width="543" height="333" alt="Image" src="https://github.com/user-attachments/assets/93f65aca-440b-402d-9f75-7758fa6c5d68" />

There is a `documentId` option!

After that, a request is made by the page, so we have another `webRequest.onCompleted()` event, but for a script file: 

<img width="522" height="392" alt="Image" src="https://github.com/user-attachments/assets/0894ab83-18cb-433b-8da6-e10efb5e4259" />

This time, we have `documentId` - the request is made after the document has been created.

[...]

PR #2837, opened 2025-11-06
fix(adblocker): utilize `documentId` for cosmetics injection

smalluban:
Relates to #2836.

This PR adds support for the `documentId` option of the `webNavigation.onCommited` event. This DOES NOT FIX the issue with `webRequest.onResponseStarted` event - it will break as before (explained in the https://github.com/ghostery/ghostery-extension/issues/2836#issuecomment-3496445543 comment).

Using `documentId` requires a logic change for applying global styles injection, as it can point to only a single frame at a time. Because of that, the logic must run for each frame on bootstrap, not only for the main frame. I expect that the call for global styles is quick (it does not relate to the page address, doesn't it?), and the performance mitigation only applies to the subframes, so we can keep the code unified for both platforms (I expect Firefox could have an old version, but it would complicate the code significantly).

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -230,6 +230,18 @@ export const setup = asyncSetup('adblocker', [
   ),
 ]);
 
+function resolveInjectionTarget(details) {
+  const target = { tabId: details.tabId };
+
+  if (details.documentId) {
+    target.documentIds = [details.documentId];
+  } else {
+    target.frameIds = [details.frameId];
+  }
+
+  return target;
+}
+
 const scriptletGlobals = {
   // Request a real extension resource to obtain a dynamic ID to the resource.
   // Redirect resources are defined with `use_dynamic_url` restriction.
@@ -240,7 +252,7 @@ const scriptletGlobals = {
     .slice(0, -6),
 };
 
-function injectScriptlets(filters, tabId, frameId, hostname) {
+function injectScriptlets(filters, hostname, details) {
   let contentScript = '';
   for (const filter of filters) {
     const parsed = filter.parseScript();
@@ -281,10 +293,7 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
         world:
           chrome.scripting.ExecutionWorld?.MAIN ??
           (__PLATFORM__ === 'firefox' ? undefined : 'MAIN'),
-        target: {
-          tabId,
-          frameIds: [frameId],
-        },
+        target: resolveInjectionTarget(details),
         func,
         args,
       },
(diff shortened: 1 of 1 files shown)


seia-soto (approved):
This change so far is a safer way to provide a cosmetic injection to users despite we need more concrete understanding and belief to the API. We recently fully disabled the `onResponseStarted` Chromium injection time and this fix will likely to address the all cases where the flag is still enabled so far. For now, this is a reasonable move to unbreak ongoing incidents.

src/background/adblocker.js
```diff
@@ -230,6 +230,18 @@ export const setup = asyncSetup('adblocker', [
   ),
 ]);
 
+function resolveInjectionTarget(details) {
+  const target = { tabId: details.tabId };
+
+  if (details.documentId) {
+    target.documentIds = [details.documentId];
+  } else {
+    target.frameIds = [details.frameId];
```
philipp-classen, line 239: The old code had in injectStyles a fallback for details.frameId being undefined:

```
   if (frameId !== undefined) {
    target.frameIds = [frameId];
  } else {
    target.allFrames = true;
  }
```

Do we also have to keep this special case here?

smalluban: It had to, as we were using that case for injecting global styles into all frames. This is no longer the case. There is no call to this function without `frameId` or `documentId` from `details`.

See this change: https://github.com/ghostery/ghostery-extension/pull/2837/files#diff-6481a5f7409b68532da99ab1713c40379d7978ca76596daff9237a31eb667543L443

philipp-classen: Could be that you are right. If I understand the old code, frameId is always present, but there is a special case when frameId=0:

In that case, it called injectStyles without passing frameId. Thus, frameId in the context of the existing injectStyles became undefined (and it fell back to all frames).

I have not confirmed it with the documentation, but I think it is correct that frameId is always present Then the new code is fine. My observation was based on the old code, which suggested (incorrectly) that it is an optional value when it is read without knowing the full context.

Only question left: is the former special case for frameId=0 (and then injecting to all frames) still needed?

src/background/adblocker.js
```diff
@@ -426,21 +429,21 @@ async function injectCosmetics(details, config) {
...
-      // This needs to be done only once per tab
       getBaseRules: true,
       getInjectionRules: false,
       getExtendedRules: false,
       getRulesFromDOM: false,
       getRulesFromHostname: false,
     });
 
-    injectStyles(styles, tabId);
+    injectStyles(styles, details);
```
philipp-classen, line 446: Note: this would be the place that used to inject to all frames, but now only injects to the current frame.

PR #2847, opened 2025-11-07
fix(settings): remove managed and assisted domains from the backup

smalluban:
Exported and imported settings should not intersect with managed and assisted paused domains.

* Removes the `managed` and `assist` paused domains
* Fix cleaning all paused domains, when importing settings - managed and assisted should not be touched

### DIFF
--- a/src/pages/settings/utils/backup.js
+++ b/src/pages/settings/utils/backup.js
@@ -33,7 +33,7 @@ export async function exportToFile() {
   cleanedOptions.paused = Object.entries(cleanedOptions.paused).reduce(
     (acc, [domain, details]) => {
       if (!details.managed && !details.assist && !details.revokeAt) {
-        acc[domain] = details;
+        acc[domain] = { revokeAt: 0 };
       }
       return acc;
     },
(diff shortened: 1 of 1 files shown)


src/pages/settings/utils/backup.js
```diff
@@ -25,10 +25,25 @@ export async function exportToFile() {
   const customFilters = await store.resolve(CustomFilters);
   const elementPickerSelectors = await store.resolve(ElementPickerSelectors);
 
+  const cleanedOptions = Object.fromEntries(
+    SYNC_OPTIONS.map((key) => [key, options[key]]),
+  );
+
+  // Remove managed, assist and temporary paused domains
+  cleanedOptions.paused = Object.entries(cleanedOptions.paused).reduce(
```
philipp-classen, line 33: Purely cosmetic, but perhaps filtering more be more natural then reducing here:

```
cleanedOptions.paused = Object.fromEntries(
  Object.entries(cleanedOptions.paused).filter(
    ([domain, details]) => !details.managed && !details.assist && !details.revokeAt
  )
);
```

smalluban: I'll update it and merge.

smalluban: I figured out that we can clean up `managed` and `assist` options from exported data, so I kept the `reduce`, as the item's value changes.

PR #2858, opened 2025-11-18
fix(adblocker): enhance ad-blocking in subframes

smalluban:
Fixes injecting cosmetics on subframes.

* On Firefox, injecting cosmetics within the `webNavigation.onCommited` event on subframes is too early, which makes the CSS not injected at all.
* The local frames (like `about:blank`) do not trigger the `webNavigation.onCommited` event at all for both platforms

Running bootstrap `injectCosmetics()` multiple times is safe on Firefox. The scriptlets are injected only once by the registration process. In Chromium, we can use `documentId` to detect if the bootstrap has already run, preventing injection scriptlets again. However, we will introduce this change in a separate pull request.


src/manifest.chromium.json
```diff
@@ -71,6 +71,7 @@
       "js": ["content_scripts/adblocker/index.js"],
       "matches": ["http://*/*", "https://*/*"],
       "run_at": "document_start",
+      "match_about_blank": true,
```
seia-soto, line 74: Just a question; does this improve the status on `documentId` not being available? or is this just to match new spec?

src/background/adblocker.js
```diff
@@ -308,19 +305,11 @@ function injectScriptlets(filters, tabId, frameId, hostname) {
 }
 
 function injectStyles(styles, tabId, frameId) {
-  const target = { tabId };
-
-  if (frameId !== undefined) {
-    target.frameIds = [frameId];
-  } else {
-    target.allFrames = true;
```
chrmod, line 316: This is most important part of this PR - we should not inject to all frames, this was just incorrect.

src/manifest.chromium.json
chrmod: This allows us to handle local iframes

Issue #1019, opened 2023-02-02
MV3: Redirect protection


chrmod:

yokoffing:
https://hackmd.io/fhA_CJGCRWaumNLBIckZeQ

chrmod:
May need https://github.com/w3c/webextensions/issues/744 https://github.com/w3c/webextensions/issues/610

PR #2881, opened 2025-11-24
feat: redirect protection

chrmod:
Implements a new redirect protection feature that blocks malicious redirects using the adblocker engine and DNR (Declarative Net Request) rules.

**Key changes:**
- New redirect protection page that intercepts and warns users about potentially malicious redirects
- Settings UI to enable/disable the feature and manage domain exceptions
- Cross-platform support (Firefox, Chromium) with platform-specific implementations
- DNR integration for managing exception rules dynamically
- Build script to generate redirect protection rulesets
- Unit tests for DNR redirect rules

Screenshot 2025-11-29 at 00 08 48

Screenshot 2025-11-29 at 00 08 58
Screenshot 2025-11-29 at 00 09 01
Screenshot 2025-11-29 at 00 09 13

Implements https://github.com/ghostery/ghostery-extension/issues/1019

### DIFF
--- a/src/pages/settings/views/privacy.js
+++ b/src/pages/settings/views/privacy.js
@@ -289,7 +289,7 @@ export default {
                         type="headline-xs"
                         layout="row gap:0.5 items:center"
                       >
-                        Redirect Tracking Protection
+                        Redirect Protection
                       </ui-text>
                       <ui-icon
                         name="chevron-right"
--- a/src/pages/settings/views/redirect-protection-add-exception.js
+++ b/src/pages/settings/views/redirect-protection-add-exception.js
@@ -44,7 +44,7 @@ export default {
           layout="column gap:3"
         >
           <ui-text type="label-l" layout="block:center margin:bottom">
-            Add redirect exception
+            Add exception
           </ui-text>
           <div layout="column gap items:start">
             <ui-text type="body-m" color="secondary">
--- a/src/pages/settings/views/redirect-protection.js
+++ b/src/pages/settings/views/redirect-protection.js
@@ -44,11 +44,11 @@ export default {
                 Back
               </ui-text>
             </settings-link>
-            <ui-text type="headline-m">Redirect Tracking Protection</ui-text>
+            <ui-text type="headline-m">Redirect Protection</ui-text>
             <ui-text type="body-l" mobile-type="body-m" color="secondary">
-              Ghostery blocks redirects to known tracking domains and shows a
-              warning page instead. You can choose to proceed or go back. Add
-              exceptions below for domains you trust.
+              Prevents websites from redirecting you through tracking services
+              or unknown destinations that may compromise your privacy or
+              security.
             </ui-text>
           </div>
           <settings-card type="content">
@@ -64,9 +64,7 @@ export default {
                     color="quaternary"
                     layout="size:3"
                   ></ui-icon>
-                  <ui-text type="headline-xs">
-                    Redirect Tracking Protection
-                  </ui-text>
+                  <ui-text type="headline-xs">Redirect Protection</ui-text>
                 </div>
               </div>
             </ui-toggle>
@@ -75,7 +73,7 @@ export default {
           html`
             <div layout="column gap:2">
               <div layout="row content:space-between items:center">
-                <ui-text type="label-l">Redirect exceptions</ui-text>
+                <ui-text type="label-l">Exceptions</ui-text>
                 <ui-button data-qa="button:redirect-protection:add">
                   <a href="${router.url(RedirectProtectionAddException)}">
                     Add


src/background/custom-filters.js
```diff
@@ -178,7 +182,14 @@ export async function updateCustomFilters(input, options) {
       result.errors.push(...errors);
     }
 
-    result.dnrRules = await updateDNRRules(rules);
+    // Apply redirect protection to DNR rules
+    const allOptions = await store.resolve(Options);
```
smalluban: AllOptions? Use `const { redirectProtection } =`

src/background/custom-filters.js
```diff
@@ -178,7 +182,14 @@ export async function updateCustomFilters(input, options) {
...
+    const rulesWithRedirects = applyRedirectProtection(rules, {
+      enabled: allOptions.redirectProtection?.enabled,
+      priority: 100,
```
smalluban: Hardcoded priority. Why 100 not 200?

src/background/dnr.js
```diff
@@ -42,6 +49,15 @@ if (__PLATFORM__ !== 'firefox') {
       );
     }
 
+    // Add redirect protection if enabled and any blocking is active
+    if (
+      ids.length &&
+      options.redirectProtection?.enabled &&
```
smalluban: Pointless `?.` - options model has structure, that allways requires fields to be present.

src/background/dnr.js
```diff
@@ -118,9 +142,16 @@ if (__PLATFORM__ !== 'firefox') {
                 }
               }
 
+              // Apply redirect protection to dynamic fixes
+              const rulesArray = Array.from(rules);
```
smalluban: usless const, only used once in the next line

src/background/dnr.js
```diff
@@ -51,12 +67,14 @@ if (__PLATFORM__ !== 'firefox') {
   OptionsObserver.addListener(async function dnr(options, lastOptions) {
     const ids = getIds(options);
 
-    if (
-      lastOptions &&
-      lastOptions.filtersUpdatedAt === options.filtersUpdatedAt &&
-      String(ids) === String(getIds(lastOptions))
-    ) {
-      // No changes in options triggering an update, skip updating rules
+    const hasChanges =
```
smalluban: This is not used anymore, the change should only add new condition. OptionsObserver has a utility to compare equaility.

src/background/dnr.js
```diff
@@ -167,6 +197,52 @@ if (__PLATFORM__ !== 'firefox') {
       ids.push('fixes');
     }
 
+    if (
+      ids.includes('redirect-protection') &&
+      options.redirectProtection?.enabled
```
smalluban: Remove `?.`

src/background/dnr.js
```diff
@@ -167,6 +197,52 @@ if (__PLATFORM__ !== 'firefox') {
...
+    ) {
+      const disabledDomains = options.redirectProtection.disabled || [];
```
smalluban: Pointless `||`

src/background/dnr.js
```diff
@@ -167,6 +197,52 @@ if (__PLATFORM__ !== 'firefox') {
...
+      const lastDisabledDomains =
+        lastOptions?.redirectProtection?.disabled || [];
```
smalluban: Remove `?.`

src/background/dnr.js
```diff
@@ -167,6 +197,52 @@ if (__PLATFORM__ !== 'firefox') {
...
+
+      const disabledDomainsChanged =
+        !lastOptions ||
+        JSON.stringify(disabledDomains.sort()) !==
```
smalluban: JSON.stringify, varialble sorted... The `options.redirectProtection.disabled should be record, not an array.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+  createRedirectProtectionExceptionRules,
+} from '/utils/dnr.js';
+
+function getRedirectUrlStorageKey(tabId) {
+  return `redirectUrl_${tabId}`;
+}
+
+async function updateOptionsWithDisabledHostname(hostname) {
+  const options = await store.resolve(Options);
+  const disabled = options.redirectProtection?.disabled || [];
```
smalluban: Remove `?.`

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+
+  if (!disabled.includes(hostname)) {
+    await store.set(Options, {
+      redirectProtection: {
+        ...options.redirectProtection,
+        disabled: [...disabled, hostname],
```
smalluban: Here it would be `disabled: { [hostname]: true }`

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+    });
+  }
+
+  return disabled.includes(hostname) ? disabled : [...disabled, hostname];
+}
+
+const allowedRedirectUrls = new Set();
+
+function shouldProtectRedirect(options, hostname) {
+  if (!options.redirectProtection?.enabled) {
```
smalluban: Remove `?.`

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+    return false;
+  }
+
+  const disabledDomains = options.redirectProtection.disabled || [];
```
smalluban: Pointless `[]`

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+  return !disabledDomains.some((domain) => hostname.endsWith(domain));
+}
+
+export function handleRedirectProtection(
+  details,
+  request,
+  options,
+  isTrusted,
+  getEngine,
```
smalluban: Dependency injection - we don't do that

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+function shouldProtectRedirect(options, hostname) {
+  if (!options.redirectProtection?.enabled) {
+    return false;
+  }
+
+  const disabledDomains = options.redirectProtection.disabled || [];
+  return !disabledDomains.some((domain) => hostname.endsWith(domain));
+}
+
+export function handleRedirectProtection(
```
smalluban: This function should be a part of adblocker.js, it is used only there, and uses many depndencies of that module.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+      '?url=' +
+      encodedUrl;
+
+    return { redirectUrl: protectionUrl };
+  }
+
+  return undefined;
+}
+
+export function allowRedirectUrl(url) {
```
smalluban: This is best example of misplace. This function is only used in adblocker.js, and it is here to expose this module variable.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+  allowedRedirectUrls.add(url);
+}
+
+export async function disableRedirectProtectionForHostname(hostname) {
```
smalluban: The same misplace issue.

src/background/adblocker.js
```diff
@@ -459,6 +464,15 @@ chrome.runtime.onMessage.addListener((msg, sender) => {
 
     injectCosmetics(details, msg);
   }
+
+  // Handle redirect protection messages (Firefox only)
+  if (__PLATFORM__ === 'firefox') {
```
smalluban: This should be saparate message listener in `background/redirect-protection.js`

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+  });
+}
+
+if (__PLATFORM__ !== 'firefox') {
+  chrome.webNavigation.onBeforeNavigate.addListener(
+    (details) => {
+      if (
+        details.frameId === 0 &&
+        details.url &&
+        !details.url.includes('redirect-protection')
```
smalluban: What if this is part of real URL?

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+      ) {
+        chrome.storage.session
+          .set({
+            [getRedirectUrlStorageKey(details.tabId)]: details.url,
```
smalluban: Why you don't use `utils/map.js`?

src/background/redirect-protection.js
```diff
@@ -0,0 +1,302 @@
...
+            error,
+          );
+          sendResponse({ success: false, error: error.message });
+        }
+      })();
+
+      return true;
+    }
+
+    if (message.action === 'disableRedirectProtection') {
```
smalluban: This should be done in place in `pages/redirect-protection` as they have access to `Options`.

The whole update of DNR, etc.. should be reaction to the newest state of the `Options`.

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,261 @@
...
+  } catch (error) {
+    console.error('[redirect-protection] Failed to allow redirect:', error);
+  }
+}
+
+async function alwaysAllow(host) {
+  if (!host.hostname) return;
+
+  try {
+    const response = await chrome.runtime.sendMessage({
```
smalluban: Update here `options.redirectProtection.disabled`

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,261 @@
...
+    }
+  } catch (error) {
+    console.error('[redirect-protection] Failed to disable protection:', error);
+  }
+}
+
+const App = {
+  targetUrl: {
+    value: '',
+    connect: async (host) => {
```
smalluban: This should be separate function, which returns `targetUrl`. You should not use `connect` to this, as will run every time element is moved in DOM (it won't happen here, but this is not meant to be used for this purpose).

This should be external async function, then set the value of property to this function (will return promise). In template use `html.resolve()` to utilize data from a promise.

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,261 @@
...
+          if (response?.url) {
+            host.targetUrl = response.url;
+          }
+        } catch (e) {
+          console.error('[redirect-protection] Failed to get URL:', e);
+        }
+      }
+    },
+  },
+  hostname: ({ targetUrl }) => {
```
smalluban: hostname and displayURL should be returned by this function too. Currently, both properties call `new URL()`, have try/catch, etc..

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,261 @@
...
+
+    if (response?.success) {
+      location.replace(host.targetUrl);
+    }
+  } catch (error) {
+    console.error('[redirect-protection] Failed to disable protection:', error);
+  }
+}
+
+const App = {
```
smalluban: We don't call main component `App`. Usually, this is a name of the module, or you can inline it at the last line of file.

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,261 @@
...
+          <div class="header" layout="margin:2:0">
+            <ui-icon name="logo-with-slogan" layout="height:3"></ui-icon>
+          </div>
+
+          <div class="modal" layout="column gap:2">
+            <ui-text type="headline-l" layout="block:center">
+              Tracking Redirect Alert
+            </ui-text>
+
+            ${targetUrl
```
smalluban: here use `html.resolve(target, ...)` if the `target` keeps result of the external function resolving your `targetUrl`, `displayUrl` and `hostname`

src/pages/settings/views/privacy.js
```diff
@@ -20,6 +20,7 @@ import RegionalFilters from './regional-filters.js';
 import ExperimentalFilters from './experimental-filters.js';
 import CustomFilters from './custom-filters.js';
 import Serp from './serp.js';
+import RedirectProtection from './redirect-protection.js';
 import { BECOME_A_CONTRIBUTOR_PAGE_URL } from '/utils/urls.js';
```
smalluban: We use single brackets.

src/pages/settings/views/redirect-protection-add-exception.js
```diff
@@ -0,0 +1,85 @@
...
+import Options from '/store/options.js';
+import Hostname from '../store/hostname.js';
+
+async function add({ options, hostname }, event) {
+  event.preventDefault();
+
+  router.resolve(
+    event,
+    store.submit(hostname).then(({ value }) => {
+      const disabled = options.redirectProtection?.disabled || [];
```
smalluban: Remove `?.`

src/pages/settings/views/redirect-protection.js
```diff
@@ -0,0 +1,153 @@
...
+
+export default {
+  [router.connect]: {
+    stack: [RedirectProtectionAddException],
+  },
+  options: store(Options),
+  render: ({ options }) => html`
+    <template layout="contents">
+      <settings-page-layout layout="column gap:4">
+        ${store.ready(options) &&
```
smalluban: Options in Settings are already `ready` - look at the `pages/settings/index.js`

src/pages/settings/views/redirect-protection.js
```diff
@@ -0,0 +1,153 @@
...
+                    type="primary"
+                    size="s"
+                    data-qa="button:redirect-protection:add"
+                  >
+                    <a href="${router.url(RedirectProtectionAddException)}">
+                      Add
+                    </a>
+                  </ui-button>
+                </div>
+                ${options.redirectProtection.disabled?.length
```
smalluban: Remove `?.`

src/pages/settings/views/redirect-protection.js
```diff
@@ -0,0 +1,153 @@
...
+                          <ui-text type="label-s" color="secondary">
+                            Website
+                            (${options.redirectProtection.disabled.length})
+                          </ui-text>
+                        </div>
+                        ${options.redirectProtection.disabled.map(
+                          (hostname) => html`
+                            <div
+                              layout="grid:1|max content:center padding:1 gap"
+                              style="border-bottom: 1px solid var(--color-divider-primary);"
```
smalluban: Inline style

src/pages/settings/views/redirect-protection.js
```diff
@@ -0,0 +1,153 @@
...
+                              </ui-action>
+                            </div>
+                          `,
+                        )}
+                      </div>
+                    `
+                  : html`
+                      <div
+                        layout="column center gap:2 padding:4"
+                        style="background: var(--background-secondary); border-radius: 8px;"
```
smalluban: Inline style

src/store/options.js
```diff
@@ -64,6 +64,12 @@ const Options = {
   // SERP protection
   serpTrackingPrevention: true,
 
+  // Redirect protection
+  redirectProtection: {
+    enabled: true,
+    disabled: [String], // List of hostnames where redirect protection is disabled
```
smalluban: This should be `disabled: record({})` - usually single access is more important than having prepared list. This is a pattern for all lists of domains in options.

tests/unit/redirect-protection.test.js
```diff
@@ -0,0 +1,243 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { applyRedirectProtection } from '../../src/utils/dnr.js';
+
+const tests = [
```
smalluban: Please use at least `node:test` package...

src/pages/settings/views/privacy.js
chrmod: not sure what you mean - probably the comment was meant for another place

.github/workflows/test.yml
```diff
@@ -27,8 +27,21 @@ jobs:
 
       - name: Lint
         run: npm run lint
+  unit:
+    runs-on: ubuntu-latest
```
smalluban, line 30: needs: [lint]

src/utils/dnr.js
```diff
@@ -13,12 +13,93 @@ export const PAUSED_ID_RANGE = { start: 1, end: 1_000_000 };
...
+};
 
 export const PAUSED_RULE_PRIORITY = 10_000_000;
 export const EXCEPTIONS_RULE_PRIORITY = 2_000_000;
+export const REDIRECT_PROTECTION_EXCEPTION_PRIORITY = 200;
+
+const MAX_DNR_PRIORITY = 1073741823;
+
+export function filterMaxPriorityRules(rules) {
+  return rules.filter((rule) => rule.priority !== MAX_DNR_PRIORITY);
```
smalluban: Shouldn't it be `>=`?

src/background/dnr.js
```diff
@@ -94,14 +108,16 @@ if (__PLATFORM__ !== 'firefox') {
 
             if (list.dnr.checksum !== resources.checksums['dnr-fixes']) {
               const rules = new Set(
-                await fetch(list.dnr.url).then((res) =>
-                  res.ok
-                    ? res.json()
-                    : Promise.reject(
-                        new Error(
-                          `Failed to fetch DNR rules: ${res.statusText}`,
+                filterMaxPriorityRules(
```
smalluban: We should filter those rules when generating rules on the server-side.

scripts/download-dnr-rulesets.js
```diff
@@ -54,17 +55,19 @@ for (const [name, target] of Object.entries(RULESETS)) {
...
           `Failed to fetch DNR rules for "${name}": ${res.status}: ${res.statusText}`,
         );
       }
 
-      return res.text();
+      return res.json();
     });
 
-    writeFileSync(outputPath, dnr);
+    dnr = filterMaxPriorityRules(dnr);
```
smalluban: We should filter out those rules on the server-side, so they never reach the extension code here.

src/background/custom-filters.js
```diff
@@ -170,14 +174,20 @@ export async function updateCustomFilters(input, options) {
 
   // Update DNR rules for Chromium and Safari
   if (__PLATFORM__ !== 'firefox') {
-    const { rules, errors } = await convert(
+    const { redirectProtection } = await store.resolve(Options);
```
smalluban: Move to line 186.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,230 @@
...
+import { store } from 'hybrids';
+
+import {
+  REDIRECT_PROTECTION_SESSION_ID_RANGE,
+  REDIRECT_PROTECTION_EXCEPTION_PRIORITY,
+  REDIRECT_PROTECTION_ID_RANGE,
+  getDynamicRulesIds,
+  createRedirectProtectionExceptionRules,
+} from '/utils/dnr.js';
+import AutoSyncingMap from '/utils/map.js';
```
smalluban, line 23: Group imports by `/store`, `/utils`. Add a new line after each group.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,230 @@
...
+          });
+        }
+      }
+    },
+  );
+}
+
+chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
+  if (message.action === 'getRedirectUrl') {
+    if (__PLATFORM__ === 'firefox') {
```
smalluban: We use the `switch` operator for listing different types of messages. Also, if a message type is used only by one platform, create a separate function for it. Tree-shaking will still have a top condition on Firefox now.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  if (message.action === 'allowRedirect') {
+    if (!message.url) {
+      sendResponse({ success: false, error: 'Missing URL' });
+      return false;
+    }
+
+    if (__PLATFORM__ === 'firefox') {
+      allowedRedirectUrls.add(message.url);
```
smalluban, line 199: Move this action to a separate function defined only for Firefox.

src/background/redirect-protection.js
```diff
@@ -0,0 +1,230 @@
...
+    }
+    const url =
+      sender.tab && sender.tab.id
+        ? redirectUrlMap.get(sender.tab.id) || null
+        : null;
+    sendResponse({ url });
+    return false;
+  }
+
+  if (message.action === 'allowRedirect') {
```
smalluban, line 192: Allow and disallow redirects are still side effects to the user action - if something goes wrong, the proper state is never re-created.

In our architecture, we should re-create the state as a listener to option changes. We can still look if the rules are there, and don't remove all and set all (we can be smart with that).

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,265 @@
...
+    return { targetUrl, hostname: '' };
+  }
+}
+
+async function allow(host) {
+  const { targetUrl } = await host.target;
+  if (!targetUrl) return;
+
+  try {
+    const response = await chrome.runtime.sendMessage({
```
smalluban: No messages passing - just update options.

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,265 @@
...
+  } catch (error) {
+    console.error('[redirect-protection] Failed to allow redirect:', error);
+  }
+}
+
+async function alwaysAllow(host) {
+  const { targetUrl, hostname } = await host.target;
+  if (!hostname) return;
+
+  const response = await chrome.runtime.sendMessage({
```
smalluban: No messages passing - just update options.

src/pages/redirect-protection/index.js
```diff
@@ -0,0 +1,265 @@
...
+            <ui-icon name="logo-with-slogan" layout="height:3"></ui-icon>
+          </div>
+
+          <div class="modal" layout="column gap:2">
+            <ui-text type="headline-l" layout="block:center">
+              Tracking Redirect Alert
+            </ui-text>
+
+            ${html.resolve(
+              target.then(({ targetUrl, hostname }) =>
```
smalluban: Use the `getTarget` function directly here; no need for the component's property.

smalluban (changes requested):

tests/e2e/spec/index.spec.js
```diff
@@ -14,3 +14,4 @@ import './main.spec.js';
 import './panel.spec.js';
 import './whotracksme.spec.js';
 import './advanced.spec.js';
+import './redirect-protection.spec.js';
```
smalluban: This file does not exist.

scripts/download-dnr-rulesets.js
chrmod: That's a option but it has own problems. I would make it work first with this approach and then when working, update the server that creates the filter.

src/utils/dnr.js
chrmod: we want to remove max priority only - it is a single value

src/pages/redirect-protection/index.js
chrmod: no, that wont work due to race conditions. it needs message passing to enforce order or execution

src/pages/redirect-protection/index.js
chrmod: same as below

src/background/redirect-protection.js
chrmod: those are session rules, meant for a very short time. We dont want to put the on Options. If anything, we should rather remove them ASAP, not sure if webNavigation as on Firefox will work as expected.

src/background/redirect-protection.js
chrmod: what will that bring?

src/background/redirect-protection.js
chrmod: There is nothing switch will help with here. And specifically in this case, we cannot add an `if` to switch block

tests/e2e/spec/index.spec.js
chrmod: it does now

src/pages/redirect-protection/index.js
chrmod: it is actually needed by the actions

package.json
```diff
@@ -11,7 +11,8 @@
     "lint": "npm run lint:src && npm run lint:build",
     "lint:src": "eslint",
     "lint:build": "npm run build firefox -- --silent && addons-linter dist",
-    "test": "npm run wdio && npm run wdio:update",
+    "test": "npm run test:unit && npm run wdio && npm run wdio:update",
+    "test:unit": "node --test",
```
smalluban, line 15: If we can't run like `node something /*.test.js`, add `tests/unit/index.js` and run this file.

scripts/build-redirect-protection-rules.js
```diff
@@ -0,0 +1,95 @@
...
+    allBlockingRules.push(...blockingRules);
+  }
+
+  console.log(`\nSummary:`);
+  console.log(`  Total rules processed: ${totalRulesProcessed}`);
+  console.log(`  Total blocking rules: ${totalBlockingRules}`);
+
+  const redirectRulesWithoutIds = applyRedirectProtection(allBlockingRules, {
+    enabled: true,
+    priority: 100,
```
smalluban: Why it is not a `REDIRECT_PROTECTION_EXCEPTION_PRIORITY` ? It's only place where the `prorioty` is overwritten. Otherwise, we can omit that config at all.

src/utils/dnr.js
```diff
@@ -39,3 +54,69 @@ export async function getDynamicRulesIds(type) {
...
+    priority: REDIRECT_PROTECTION_EXCEPTION_PRIORITY,
+    action: { type: 'allow' },
+    condition: {
+      urlFilter: `||${hostname}^`,
+      resourceTypes: ['main_frame'],
+    },
+  }));
+}
+
+export function applyRedirectProtection(
```
smalluban: The function should have the following arguments: `(options, rules)`

src/background/redirect-protection.js
```diff
@@ -0,0 +1,232 @@
...
+const redirectUrlMap = new AutoSyncingMap({
+  storageKey: 'redirectUrls:v1',
+  ttlInMs: 5 * 60 * 1000,
+});
+
+const REDIRECT_PROTECTION_PAGE_URL = chrome.runtime.getURL(
+  'pages/redirect-protection/index.html',
+);
+
+const allowedRedirectUrls = new Set();
```
smalluban, line 106: This will die on SW restart on Firefox. This is an edge case, but if the user opens tab with redirect, wait 30s, then this set will be gone...

chrmod: it's fine. it is meant to be short lived. We should actually clear the session rules on MV3 as well.

smalluban:
Zrzut ekranu 2025-12-18 o 15 35 18
Zrzut ekranu 2025-12-18 o 15 22 33
Zrzut ekranu 2025-12-18 o 15 22 14

chrmod (commented):
Looks good! Lets merge

PR #2882, opened 2025-11-25
feat: add `zap` and `ghostery` filtering modes

smalluban:
Fixes https://github.com/ghostery/collaboration-kalin/issues/51

Zrzut ekranu 2025-11-26 o 09 47 09
Zrzut ekranu 2025-11-26 o 09 46 45
Zrzut ekranu 2025-11-26 o 09 46 30

### DIFF
--- a/src/background/zapped.js
+++ b/src/background/zapped.js
@@ -78,7 +78,7 @@ if (__PLATFORM__ !== 'firefox') {
               priority: PAUSED_RULE_PRIORITY,
               action: { type: 'allowAllRequests' },
               condition: { excludedRequestDomains: excludedDomains },
-              resource_types: ['main_frame'],
+              resourceTypes: ['main_frame'],
             },
       ],
       removeRuleIds,


src/background/paused.js
```diff
@@ -69,64 +70,76 @@ OptionsObserver.addListener('paused', async (paused, lastPaused) => {
...
-    __PLATFORM__ !== 'firefox' &&
-    (lastPaused ||
+  if (__PLATFORM__ !== 'firefox') {
+    if (
+      // The background process starts and runs for each tab, so we can assume
+      // that this function is called before the user can change the paused state
+      // in the panel or the settings page.
+      !OptionsObserver.isOptionEqual(options.paused, lastOptions?.paused) ||
+      // Filtering mode has changed
+      (lastOptions && options.filteringMode !== lastOptions?.filteringMode) ||
```
philipp-classen:
```suggestion
      (lastOptions && options.filteringMode !== lastOptions.filteringMode) ||
```

src/pages/panel/views/main.js
```diff
@@ -98,6 +103,34 @@ async function togglePause(host, event) {
   `);
 }
 
+async function toggleZapped(host) {
+  const { options, stats, paused } = host;
+
+  const zappedHostname = Object.keys(options.zapped)
```
philipp-classen: If someone keeps using it, the list of "zapped" sites could grow large. So, the way how we store the data, could matter.

By reading this part in isolation, it is not clear to me why the keys have to be iterated in the order of their hostname. If there is no match or exactly one, it will not matter. And if there are multiple matches (and collision resolution is important), it is not obvious to me why "a.foo.com" would be better than "b.bar.com", for instance).

Also, it is not clear to me when reading this part, why the hostname that this part is trying to find, depends on both the list of zapped hostnames and the hostname in stats. Is "stats.hostname" the hostname of the page that is currently open?

src/pages/settings/views/websites.js
```diff
@@ -42,7 +46,14 @@ function revokeCallback(item) {
     if (item.assist) dismissAction(item.id, ACTION_PAUSE_ASSISTANT);
 
     store.set(ElementPickerSelectors, { hostnames: { [item.id]: null } });
-    store.set(options, { paused: { [item.id]: null }, exceptions });
+
+    if (options.filteringMode === FILTERING_MODE_GHOSTERY) {
+      store.set(options, { paused: { [item.id]: null }, exceptions });
+    }
+
+    if (options.filteringMode === FILTERING_MODE_ZAP) {
```
philipp-classen:
```suggestion
    } else if (options.filteringMode === FILTERING_MODE_ZAP) {
```

src/pages/settings/views/websites.js
```diff
@@ -57,29 +68,40 @@ export default {
...
+          id,
+          revokeAt,
+          assist,
+          managed,
+          exceptions: new Set(),
+          counter: 0,
+        }));
+    }
+
+    if (options.filteringMode === FILTERING_MODE_ZAP) {
```
philipp-classen:
```suggestion
    } else if (options.filteringMode === FILTERING_MODE_ZAP) {
```

src/pages/panel/views/main.js
smalluban: The storing mechanism is the same as for paused domains. I expect that in Zap mode is about blocking only selected sites (not many), as normally you pause selected sites. If someone has a lot of domains there, I think the user should use normal mode.
However, we don't have much choice. We can choose between a record and an array. The first is better, as usually we need information if the domain is there, so access by key is cleanest and fast enough (I expect [].includes and property[key] has similar performance). 

About the rest of the code. This is required for situations where the user enables (or pauses in ghostery mode) on a subdomain where the top domain is already enabled/paused. We must find the top-level domain first and act on that domain. This is why we sort (shortest to longest names), keeping all subdomains one after another. The logic is only to sort in the scope of one top domain; the order of different top domains does not matter.

src/background/pause-assistant.js
```diff
@@ -78,19 +87,32 @@ store.observe(Config, async (_, config, lastConfig) => {
...
 chrome.webNavigation.onCompleted.addListener(async (details) => {
   if (details.frameId === 0) {
-    if (!(await store.resolve(Options)).pauseAssistant) return;
-
     const managedConfig = await store.resolve(ManagedConfig);
     if (managedConfig.disableUserControl) return;
 
+    const options = await store.resolve(Options);
+    if (
+      options.filteringMode !== FILTERING_MODE_GHOSTERY ||
```
chrmod: I don't think I like the name `FILTERING_MODE` as it affects more than just "filtering". Perhaps `MODE` would be good enough? I will think about it

src/background/paused.js
```diff
@@ -11,41 +11,42 @@
...
-  'xmlhttprequest',
-  'ping',
-  'media',
-  'websocket',
-  'webtransport',
-  'webbundle',
-  'other',
-];
+OptionsObserver.addListener(async function pausedSites(options, lastOptions) {
+  if (options.filteringMode !== FILTERING_MODE_GHOSTERY) {
```
chrmod: `options.filteringMode === FILTERING_MODE_ZAP` - should be safer in case we introduce more modes

src/background/paused.js
```diff
@@ -69,64 +70,76 @@ OptionsObserver.addListener('paused', async (paused, lastPaused) => {
...
   }
 
-  // The background process starts and runs for each tab, so we can assume
-  // that this function is called before the user can change the paused state
-  // in the panel or the settings page.
-  if (
-    __PLATFORM__ !== 'firefox' &&
-    (lastPaused ||
+  if (__PLATFORM__ !== 'firefox') {
+    if (
```
chrmod: can we avoid indentation change here for the diff to be more readable?

src/background/zapped.js
```diff
@@ -0,0 +1,84 @@
+/**
```
chrmod, line 1: `zapped` or `zap` ? the mode is called `zap`

src/background/zapped.js
```diff
@@ -0,0 +1,84 @@
...
+import { FLAG_FILTERING_MODE } from '/utils/config-types.js';
+import {
+  getDynamicRulesIds,
+  PAUSED_ID_RANGE,
+  PAUSED_RULE_PRIORITY,
+  ALL_RESOURCE_TYPES,
+} from '/utils/dnr.js';
+import * as OptionsObserver from '/utils/options-observer.js';
+
+// Clear filtering mode and zapped data if the flag is removed
```
chrmod, line 26: why we clear? cannot we just keep it?

src/background/zapped.js
```diff
@@ -0,0 +1,84 @@
...
+          condition: {
+            excludedInitiatorDomains: excludedDomains,
+            excludedRequestDomains: excludedDomains,
+            resourceTypes: ALL_RESOURCE_TYPES,
+          },
+        },
+        {
+          id: 2,
+          priority: PAUSED_RULE_PRIORITY,
+          action: { type: 'allowAllRequests' },
```
chrmod: does it work as expected on Safari? Shouldn't we apply same approach we do for pausing?

src/pages/panel/views/main.js
chrmod: In general, and also for lookup on Firefox, we may want to use a more efficient data structure. But for first iteration a map is probably good enough.

src/pages/panel/views/main.js
```diff
@@ -98,6 +103,34 @@ async function togglePause(host, event) {
...
+    .sort((a, b) => b.localeCompare(a))
```
chrmod: This logic should not belong to a view. The fact you need to sort before lookup is not obvious and may be missed.

src/pages/panel/views/main.js
chrmod: A Reverse domain trie would be a good candidate

src/pages/panel/views/main.js
```diff
@@ -287,83 +329,103 @@ export default {
...
-            html`
-              <panel-pause
-                onaction="${globalPause ? revokeGlobalPause : togglePause}"
-                paused="${paused || globalPause}"
-                global="${globalPause}"
-                managed="${paused?.managed}"
-                assist="${paused?.assist}"
-                revokeAt="${globalPause?.revokeAt || paused?.revokeAt}"
-                data-qa="component:pause"
+        ${!options.terms &&
```
chrmod, line 328: code here becomes quite long, maybe its worth to encapsulate it in own component?

src/pages/panel/views/main.js
```diff
@@ -444,10 +506,10 @@ export default {
                                     ></ui-icon>`}
                                   </a>
                                 </ui-text>
-                                ${!paused &&
-                                !globalPause &&
-                                options.terms &&
+                                ${options.terms &&
```
chrmod, line 506: why there is a change here?

src/pages/settings/views/websites.js
```diff
@@ -42,7 +46,12 @@ function revokeCallback(item) {
     if (item.assist) dismissAction(item.id, ACTION_PAUSE_ASSISTANT);
 
     store.set(ElementPickerSelectors, { hostnames: { [item.id]: null } });
-    store.set(options, { paused: { [item.id]: null }, exceptions });
+
+    if (options.filteringMode === FILTERING_MODE_GHOSTERY) {
+      store.set(options, { paused: { [item.id]: null }, exceptions });
+    } else if (options.filteringMode === FILTERING_MODE_ZAP) {
+      store.set(options, { zapped: { [item.id]: null }, exceptions });
```
chrmod, line 53: is this actually removing a key from the map or sets its value to `null`?

src/store/managed-config.js
```diff
@@ -23,6 +23,11 @@ const ManagedConfig = {
   disableTrackersPreview: false,
   trustedDomains: [TRUSTED_DOMAINS_NONE_ID],
 
+  disableFilteringMode: (config) =>
```
chrmod: what does that mean?

src/store/options.js
```diff
@@ -17,15 +17,17 @@ import { isOpera, isSafari } from '/utils/browser-info.js';
...
+export const FILTERING_MODE_GHOSTERY = 'ghostery';
+export const FILTERING_MODE_ZAP = 'zap';
 
 export const ENGINES = [
   { name: 'ads', key: 'blockAds' },
   { name: 'tracking', key: 'blockTrackers' },
   { name: 'annoyances', key: 'blockAnnoyances' },
 ];
 
+export const GLOBAL_PAUSE_ID = '<all_urls>';
```
chrmod: why you change order?

src/store/options.js
```diff
@@ -88,12 +95,18 @@ const Options = {
   // Tracker exceptions
   exceptions: store.record({ global: false, domains: [String] }),
 
-  // Paused domains
+  // Paused domains (ghostery filtering mode)
   paused: store.record({ revokeAt: 0, assist: false, managed: false }),
 
+  // Zapped domains (zap filtering mode)
+  // Empty record for future additions (if any)
+  zapped: store.record(true),
```
chrmod, line 101: I have questions :)

src/utils/urls.js
```diff
@@ -14,12 +14,9 @@ import { debugMode, stagingMode } from './debug.js';
 
 export const GHOSTERY_DOMAIN = debugMode ? 'ghosterystage.com' : 'ghostery.com';
 
+export const TERMS_AND_CONDITIONS_URL = `https://www.${GHOSTERY_DOMAIN}/privacy/ghostery-terms-and-conditions?utm_source=gbe&utm_campaign=onboarding`;
 export const HOME_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/`;
 
-export const SIGNON_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/signin`;
```
chrmod: not related

chrmod (changes requested):

src/background/pause-assistant.js
smalluban: The UI name for the feature is called "Filtering Mode" (onboarding, settings page). Do you think we should change it as well?

src/background/paused.js
smalluban: Yep, good point.

src/background/zapped.js
smalluban: The file name reflects the options field name, which follows the `paused.js` file.

src/background/zapped.js
smalluban: We can keep it - it is not a technical requirement. This is meant to protect from migrating options. If, for some reason, we need to change the `options.zapped` structure, we can turn off the flag, and the structure will be empty. Also, it makes a clean state if we want to turn off the flag.

However, I don't have a strong opinion here.

src/background/zapped.js
smalluban: This code reflects exactly what we do in `paused.js`. The only difference is that we need to combine conditions, as otherwise it won't work (as the reverse of AND is OR).

src/pages/panel/views/main.js
smalluban: I agree, we can work on that. I discussed that with @philipp-classen today. I think we need a helper function in the scope of `store/options.js` like `getPausedDetails()`.

src/pages/panel/views/main.js
smalluban: Those parts relate to most of the properties passed to the template, so it would be a much of copy of the header of the component. It is the structure of the main view. Even though it is long, it is separated into features. Nothing here is reused in any other place. If something is a UI component, it is already separated.

src/pages/panel/views/main.js
smalluban: `options.terms` is an "entry" condition, used to show which part of the view is for the state before onboarding and after onboarding. This change makes it more visible.

src/pages/settings/views/websites.js
smalluban: It removes the `[item.id]` key from the record.

src/store/managed-config.js
smalluban: Filtering mode is a feature that allows a user to switch from default Ghostery mode to different modes (currently to only one other mode called "zap"). This flag disables this feature - so the user can't change `options.filteringMode` from its default value. 
In practice, the onboarding and settings page uses this computed property to hide the filtering mode.

src/store/options.js
smalluban: Cosmetic change, I'll revert.

src/utils/urls.js
smalluban: I'll create a separate PR.

src/background/paused.js
smalluban: After all, it is not good point ;) This file is for Ghostery mode, so if the mode is DIFFERENT THAN ghostery mode, we need to do a clean up. This is not related to zap directly (if there would be more modes).

src/background/paused.js
smalluban: Yes.

src/pages/panel/views/main.js
smalluban: I've created `getTopDomainFromRecord()` function in `options.js`. It unifies logic for domain lookup in `options.paused` and `options.zapped`.

src/background/zapped.js
chrmod: me neither, we can keep as is. also, having a way to clean may be useful

src/background/paused.js
```diff
@@ -11,41 +11,42 @@
...
-  'websocket',
-  'webtransport',
-  'webbundle',
-  'other',
-];
+OptionsObserver.addListener(async function pausedSites(options, lastOptions) {
+  if (options.filteringMode !== FILTERING_MODE_GHOSTERY) {
+    // Filtering mode has changed - clean up alarms
+    if (lastOptions && options.filteringMode !== lastOptions?.filteringMode) {
+      const alarms = (await chrome.alarms.getAll()).filter(({ name }) =>
```
chrmod: nit: no need for a variable here. As this is very simple you could replace filter with forEach and have everything in one callback

src/background/paused.js
```diff
@@ -118,15 +120,24 @@ OptionsObserver.addListener('paused', async (paused, lastPaused) => {
               resourceTypes: ['main_frame', 'sub_frame'],
             },
           },
+          {
+            id: 4,
+            priority: PAUSED_RULE_PRIORITY,
```
chrmod: why we add this rule? is this for Safari only?

src/background/paused.js
```diff
@@ -118,15 +120,24 @@ OptionsObserver.addListener('paused', async (paused, lastPaused) => {
...
+            action: { type: 'allowAllRequests' },
+            condition: {
+              requestDomains: globalPause ? undefined : hostnames,
+              resourceTypes: ['main_frame', 'sub_frame'],
```
chrmod: I skeptical if we should put sub_frame here (same in the rule above)

src/background/zapped.js
```diff
@@ -0,0 +1,84 @@
...
+
+    const removeRuleIds = await getDynamicRulesIds(PAUSED_ID_RANGE);
+    const excludedDomains = Object.keys(options.zapped);
+
+    await chrome.declarativeNetRequest.updateDynamicRules({
+      addRules: [
+        {
+          id: 1,
+          priority: PAUSED_RULE_PRIORITY,
+          action: { type: 'allow' },
```
chrmod: isn't `allowAllRequests` sufficient?

src/background/zapped.js
```diff
@@ -0,0 +1,84 @@
...
+          },
+        },
+        {
+          id: 2,
+          priority: PAUSED_RULE_PRIORITY,
+          action: { type: 'allowAllRequests' },
+          condition: {
+            excludedInitiatorDomains: excludedDomains,
+            excludedRequestDomains: excludedDomains,
+            resourceTypes: ['main_frame', 'sub_frame'],
```
chrmod: shouldn't this be just `main_frame`?

src/background/zapped.js
```diff
@@ -0,0 +1,84 @@
...
+    await chrome.declarativeNetRequest.updateDynamicRules({
+      addRules: [
+        {
+          id: 1,
+          priority: PAUSED_RULE_PRIORITY,
+          action: { type: 'allow' },
+          condition: {
+            excludedInitiatorDomains: excludedDomains,
+            excludedRequestDomains: excludedDomains,
+            resourceTypes: ALL_RESOURCE_TYPES,
```
chrmod: if applied to a domain owned by a tracker, doesn't this rule allow this tracker on all websites?

src/store/options.js
```diff
@@ -245,16 +257,35 @@ async function manage(options) {
   return options;
 }
 
+export function getTopDomainFromRecord(record, hostname = '') {
+  if (!hostname) return null;
+
+  const domain = Object.keys(record)
+    .sort((a, b) => b.localeCompare(a))
```
chrmod, line 265: please add a comment on why sorting is needed.

src/store/options.js
```diff
@@ -245,16 +257,35 @@ async function manage(options) {
   return options;
 }
 
+export function getTopDomainFromRecord(record, hostname = '') {
```
chrmod: how about `findClosestParentDomain`?  or just `findParentDomain`?  

I would prefer not to mention the `record` as it is implementational detail

src/pages/panel/views/main.js
```diff
@@ -98,6 +105,32 @@ async function togglePause(host, event) {
   `);
 }
 
+async function toggleZapped(host) {
+  const { options, stats, paused } = host;
+
+  const zappedHostname = getTopDomainFromRecord(options.zapped, stats.hostname);
+
+  await store.set(options, {
```
chrmod, line 111: Best to move business logic to the background, but if you rely on some side effects of having it in the view, then at least, please encapsulate it in a function and preferably move to non-view file.

src/pages/settings/views/websites-add.js
```diff
@@ -20,15 +23,20 @@ async function add({ options, hostname, pauseType }, event) {
...
-      });
+          return store.set(options, {
+            paused: {
+              [value]: {
+                revokeAt: pauseType && Date.now() + 60 * 60 * 1000 * pauseType,
+              },
+            },
+          });
+        case FILTERING_MODE_ZAP:
+          return store.set(options, { zapped: { [value]: {} } });
```
chrmod: the empty object here asks for some explanations. maybe the store should provide an utility function to create records, for example `store.createRecord(Options.zapped, {})` (in this case it could be `store.createRecord(Options.zapped)`)

chrmod: or event `store.createRecord(options, Options.zapped, value, {})`

src/store/options.js
```diff
@@ -245,16 +257,35 @@ async function manage(options) {
...
+    .sort((a, b) => b.localeCompare(a))
+    .find((d) => hostname.endsWith(d));
+
+  return domain || null;
+}
+
 export function getPausedDetails(options, hostname = '') {
   if (options.paused[GLOBAL_PAUSE_ID]) {
     return options.paused[GLOBAL_PAUSE_ID];
+  } else if (!hostname) {
```
chrmod: in which case it can be empty?

src/pages/onboarding/views/modes.js
```diff
@@ -0,0 +1,132 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { html, router, store } from 'hybrids';
+
+import modeGhosteryScreenshotUrl from '/ui/assets/mode-ghostery.svg';
```
chrmod, line 14: Didn't Kalin provide gifs?

src/ui/components/icon.js
```diff
@@ -80,6 +80,24 @@ const icons = {
...
     </svg>
   `,
+  'logo-in-box': /* html */`
```
chrmod, line 83: this file is huge.  Why cannot we keep them as separate files. None of the new ones uses current color

src/ui/components/icon.js
```diff
@@ -99,6 +117,90 @@ const icons = {
...
...8C53.7933 22.6449 53.7281 22.4413 53.7043 22.2409L53.0234 22.3216L52.3424 22.4022C52.3887 22.7933 52.5145 23.1668 52.7102 23.5198L53.3099 23.1873ZM53.0234 22.3216L53.7043 22.2409C53.6599 21.866 53.7404 21.5517 53.9437 21.2627L53.3828 20.8682L52.8219 20.4737C52.4186 21.0472 52.2594 21.7018 52.3424 22.4022L53.0234 22.3216ZM53.3828 20.8682L53.9429 21.2638C54.1272 21.0029 54.4103 20.8236 54.8839 20.7675L54.8033 20.0865L54.7226 19.4056C53.9327 19.4992 53.2688 19.841 52.8227 20.4726L53.3828 20.8682ZM54.8033 20.0865L54.8839 20.7675C55.5645 20.6869 55.9683 20.8059 56.2094 20.9892L56.6244 20.4433L57.0393 19.8974C56.4116 19.4203 55.6013 19.3015 54.7226 19.4056L54.8033 20.0865ZM56.6244 20.4433L56.2194 20.9967C56.521 21.2174 56.646 21.4493 56.6762 21.7048L57.3572 21.6242L58.0381 21.5435C57.957 20.8581 57.5934 20.3028 57.0293 19.89L56.6244 20.4433ZM57.3572 21.6242L56.6762 21.7048C56.7148 22.0304 56.6264 22.3599 56.3318 22.7221L56.8638 23.1548L57.3957 23.5875C57.8864 22.9843 58.127 22.2934 58.0381 21.5435L57.3572 21.6242ZM56.8638 23.1548L56.3388 22.7137C56.0927 23.0066 55.6667 23.2399 54.9416 23.3258L55.0223 24.0068L55.103 24.6877C56.0448 24.5761 56.8464 24.2415 57.3888 23.5958L56.8638 23.1548Z"
```
chrmod, line 173: this must impact first render of all UIs

tests/e2e/wdio.conf.js
```diff
@@ -95,18 +97,12 @@ export function buildForChrome() {
...
-      'spec/index.spec.js',
-    ],
-  ],
+  specs: ['spec/index.spec.js'],
   reporters: [['spec', { showPreface: false }]],
   logLevel: argv.debug ? 'error' : 'silent',
   mochaOpts: {
     timeout: argv.debug ? 24 * 60 * 60 * 1000 : 60 * 1000,
-    retries: 2,
+    retries: 1,
```
chrmod, line 105: can we make it 0? :D

src/pages/onboarding/views/modes.js
smalluban: I am aware that Kalin is making them, but he hasn't notified me that they are ready.

src/background/paused.js
smalluban: This PR is not meant to change logic here. I added rule number 4, as we have the same pattern above - action "allow", two splitted conditions `initiatorDomains` and `requestDomains` - if the `allowAllRequests` doesn't work with the `requestDomains` condition, we can remove it. It was my precaution.

src/background/paused.js
smalluban: We can address this in a separate PR (if so, please create an issue).

src/background/zapped.js
smalluban: From my testing, it wasn't, not sure why.

src/background/zapped.js
smalluban: https://github.com/ghostery/ghostery-extension/pull/2882#discussion_r2577025250

I copied (reverse) rules from `paused.js`

src/store/options.js
smalluban: Here: https://github.com/ghostery/ghostery-extension/blob/feat-filtering-mode/src/background/dnr.js#L30

When we check if there is a global pause set (only), not if it is just paused (any case).

src/store/options.js
smalluban: Ok.

src/store/options.js
smalluban: Ok.

tests/e2e/wdio.conf.js
smalluban: Yes, we can try it out. However, there is always a case of some glitch in the e2e execution, so I wanted to lower the number gradually. Let's keep `1`, and if in a few weeks we will not have flaky executions, we can try out `0`.

src/ui/components/icon.js
smalluban: It's still just plain text - not a big raster image. Most of these icons must be loaded in the panel/settings page (most used pages). There is no technical limitation to split this into separate files, but as we avoid the compilation step, they would be separate imports. I suggest avoiding it at least for performance reasons (loading time).

Logos are multiple color icons, which do not use `currentColor` (only monochromatic icons use it).

src/ui/components/icon.js
smalluban: Yes, but you would have to check how much. I expect that loading and compiling `icon.js` is negligible. I checked, and onboarding compiles all the files required to render a view in 18ms.

src/background/zapped.js
smalluban: It works the opposite, it allows everything except a list of domains... If you add a domain, it should be blocked, not allowed.

src/pages/settings/views/websites-add.js
smalluban: This empty object here is just a value... I am not sure why you want a special function for creating a plain object. We must set it as an empty object, as we want to keep this for future additions. It could be `true` as well, but we did it in this way on purpose.

src/pages/panel/views/main.js
smalluban: This function meets the pattern of "views". Those files contain a view component (no UI definitions, like styles etc), it is connected to the router. The file is rather a "controller + view" in MVC sense. 

We can move out the logic from here, but look at the function body - it is strictly connected with the context (host argument) and runs UI side effects - adding an alert. This belongs to this view and only this view. There is no other place in extension that toggles Zap mode.

src/background/zapped.js
smalluban: I checked the current solution, and the number of blocked trackers is exactly the same on normal mode and enabled domain in zapped mode.

src/background/zapped.js
```diff
@@ -0,0 +1,89 @@
...
+      return;
+    }
+
+    const removeRuleIds = await getDynamicRulesIds(PAUSED_ID_RANGE);
+    const excludedDomains = Object.keys(options.zapped);
+
+    await chrome.declarativeNetRequest.updateDynamicRules({
+      addRules: [
+        isWebkit()
+          ? // Safari/WebKit has a bug with setting exdludedRequestDomains,
```
philipp-classen:
```suggestion
          ? // Safari/WebKit has a bug with setting excludedRequestDomains,
```

src/background/zapped.js
```diff
@@ -0,0 +1,89 @@
...
+              priority: PAUSED_RULE_PRIORITY,
+              action: { type: 'allow' },
+              condition: { excludedInitiatorDomains: excludedDomains },
+            }
+          : {
+              id: 1,
+              priority: PAUSED_RULE_PRIORITY,
+              action: { type: 'allowAllRequests' },
+              condition: { excludedRequestDomains: excludedDomains },
+              resource_types: ['main_frame'],
```
philipp-classen: In these examples, it is camel-case:
https://developer.chrome.com/docs/extensions/reference/api/declarativeNetRequest

```suggestion
              resourceTypes: ['main_frame'],
```

PR #2884, opened 2025-11-27
fix(adblocker): use `documentId` for cosmetics injection

smalluban:
Fixes [#2836](https://github.com/ghostery/ghostery-extension/issues/2836)


src/background/adblocker.js
```diff
@@ -230,6 +231,26 @@ export const setup = asyncSetup('adblocker', [
...
+const INJECTION_TARGET_DOCUMENT_ID = resolveFlag(
+  FLAG_INJECTION_TARGET_DOCUMENT_ID,
+);
+
+function resolveInjectionTarget(details) {
+  const target = { tabId: details.tabId };
+
+  if (
+    __PLATFORM__ !== 'firefox' &&
+    INJECTION_TARGET_DOCUMENT_ID.enabled &&
```
smalluban, line 243: Resolving the flag is an asynchronous action. When SW restarts for the request, it might be too late for `onResponseStarted` event. We have to keep it in mind. The behavior differs between the flag set to 100%  and no flag.

However, it might be too late to set the `onResponseStarted` flag as well ;)

philipp-classen: I wonder if "resolveFlag" should use `undefined` instead of `false` until the value is known?

It is initialized here:
https://github.com/ghostery/ghostery-extension/blob/79069ef3adffa9ad70cc3269fb3c2430a55e25c9/src/store/config.js#L121

Then the caller could decide what the default should be:

INJECTION_TARGET_DOCUMENT_ID.enabled ?? true
or
INJECTION_TARGET_DOCUMENT_ID.enabled ?? false

Or print a warning if the value is accessed while it is not known yet.

smalluban: Flags if not resolved can't be true, as if the flag resolves to false, it can have unexpected results with one running with and then without the flag (immediately). 
I haven't tested those race conditions. It is about how fast local storage resolves data, and if this is before or after calling listeners of `webNavigation` and `webRequest`.

However, the most important reason is if we want to turn off the flag completely because of some bugs etc.. you can't start with a turned on flag...

smalluban:
I have found that we have never been correctly injecting cosmetics (at least CSS) into subframes in Safari. Here is the proof:

1. Safari v10.5.19 (before adblocker.js change to target each frame, so it uses `allFrames`)
Zrzut ekranu 2025-12-3 o 10 24 45

2. Safari v10.5.21 - this branch without flag (current production):
Zrzut ekranu 2025-12-3 o 10 26 38

3. Current branch with flag turned on:
Zrzut ekranu 2025-12-3 o 10 27 06

It looks like we have to use `documentId` in Safari. Otherwise, it does not work in subframes.

src/background/adblocker.js
philipp-classen: We can move this topic out of the changes here, because it is independent.

(In a nutshell, it is the realization that it is currently not possible to distinguish if a flag is "false", because it should be false or because it is not loaded yet. By having it start as "undefined" and then move towards "true"/"false", it would preserve that information. For instance, logging when races happen, is currently not possible.)

philipp-classen: Coming back to the actual question if the flag is ready here: from my tests, it looks like it is fast enough.

Tested with Chrome and adding logging (introducing locally the "undefined" state), killing the service workers and only observing through the chrome://serviceworker-internals/?devtools page.

No proof, but overall I think it is very likely that it is woken up fast enough by other listener before.

PR #2888, opened 2025-11-27
fix(engines): apply config and resources from fixes to memory-based engines

### LINKED ISSUE #2831: Engine resources aren't available if only custom filters engine is enabled
If the only enabled engine is "custom filters", the length of array will be `1` and no resources will be copied to the final engine.

https://github.com/ghostery/ghostery-extension/blob/25e762dbd3b4e754526ab181201d8134049f7da3/src/utils/engines.js#L436-L444

The behaviour can be tested on https://seia-soto.github.io/checkadb/ on Firefox with every recent Ghostery extension versions including #2829. The instructions are as below.

1. Disable all features: ad-blocking, anti-tracking, never-consent, and all regional filters
2. Apply `/gen/rediradsbygoogle.js^$redirect=googlesyndication_adsbygoogle.js` to the custom filters
3. Open the testing page and check the entry: `/gen/rediradsbygoogle.js^$redirect=googlesyndication_adsbygoogle.js` under `networking` section.

The expected result should be `pass` here.

smalluban:
Fixes #2831

### DIFF
--- a/src/background/adblocker.js
+++ b/src/background/adblocker.js
@@ -149,7 +149,7 @@ export async function reloadMainEngine() {
       `[adblocker] Main engine reloaded with: ${enabledEngines.join(', ')}`,
     );
   } else {
-    engines.create(engines.MAIN_ENGINE);
+    await engines.create(engines.MAIN_ENGINE);
     console.info('[adblocker] Main engine reloaded with no filters');
   }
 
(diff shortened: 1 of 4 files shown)


smalluban:
I consider if we can switch the Fixes engine to TrackerDB for loading config. Does it contain the same config as the rest? It is loaded at first start before onboarding.

PR #2900, opened 2025-12-02
Prevent duplicates in metrics

philipp-classen:
Avoid clients with broken storage (e.g. out-of-disk space) sending duplicated metrics. This change separates it in three phases:
1. Preparing the sending of requests
2. Writing to disk
3. Sending the requests

Its goal is to have "at most one" semantic. It does not attempt to resend messages.

It should also address potential races (multiple writers to storage). Now, it is sentimentalized (only one writer), so this should no longer be possible.

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -15,7 +15,6 @@ import Options from '/store/options.js';
 import { debugMode } from '/utils/debug.js';
 import asyncSetup from '/utils/setup.js';
 import * as OptionsObserver from '/utils/options-observer.js';
-import { captureException } from '/utils/errors.js';
 
 import Metrics, { processUrlQuery } from './metrics.js';
 
@@ -47,12 +46,7 @@ async function detectUTMs() {
 }
 
 async function saveStorage(storage) {
-  try {
-    await chrome.storage.local.set({ metrics: storage });
-  } catch (e) {
-    captureException(e);
-    throw e;
-  }
+  await chrome.storage.local.set({ metrics: storage });
 }
 
 let runner;
@@ -68,7 +62,7 @@ const setup = asyncSetup('telemetry', [
       metrics.utm_source = utms.utm_source || '';
       metrics.utm_campaign = utms.utm_campaign || '';
 
-      saveStorage(metrics);
+      await saveStorage(metrics);
     }
 
     runner = new Metrics({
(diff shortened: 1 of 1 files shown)


smalluban:
I looked more carefully, and it might not be enough for saving ping dates (Metrics saves it itself). Still, it will protect from sending too many install pings. 

However, the `this.saveStorage()` is called before sending requests. We should add `await` (in a few places), and it will cover all cases.

philipp-classen:
> The sufficient solution should be just adding `await` here https://github.com/ghostery/ghostery-extension/blob/main/src/background/telemetry/index.js#L71
> 
> For performance reasons, we don't wait until the storage is saved correctly. If we add `await`, the `Metrics` won't start if the storage is broken.

Why is performance - or I rather think you mean latency - relevant for sending metrics? "ping" is a synchronous operation that starts fire-and-forget updates in the background. The change in the PR is that they are transactional now (instead of having concurrent writers). From performance, it should be rather better because stores are batched. Now it is one save operation instead of multiple ones. Also, previously multiple ones were started and it implicitly relied on the semantic that the last write to the key wins. (Which might be true, but I'm not sure if it guaranteed.)

Regarding latency: for sending telemetry, I do not think latency is a concern, because it is a background operation. And it should not take long enough to run into the problem that the service worker gets killed either.

But the current approach of having multiple sources being able to trigger metrics (one dimension of writers) and then also having multiple writes within the operation, makes it very hard to reason about changes and the consistency of the in-memory and persisted state. By having only one writer and running all in-memory operations before persisting then, consistency is easier to verify.

One aspect, which is out-of-scope of this PR, is that having transactions and separating state changes would enable a retry mechanism if we intend to implement that at one point. To move from "at-most-one", as should be with the change here, closer to an "exactly-one" semantic).

smalluban:
The performance was my assumption. I think it might have been also another reason, like just now waiting for sending signal, as we assumed that it was just fine...

We have only a single place in the app, which sends a telemetry signal - https://github.com/ghostery/ghostery-extension/blob/main/src/pages/panel/index.js. We don't have any race conditions currently, as everything else is send at once on SW restart.

If we add `await` in every call to `this.saveStorage()` and here https://github.com/ghostery/ghostery-extension/blob/main/src/background/telemetry/index.js#L71, it would be a safe change to push it this week. If we need more logic changes, I don't mind, but we can do it separately.

PR #2915, opened 2025-12-10
fix(zap): use `allow` rule for DNR in all platforms

smalluban:
The `allowAllRequests` action does not work as expected on Chrome. Here we have a screenshot of the panel on youtube.com when it is disabled in Zap mode:

Zrzut ekranu 2025-12-10 o 10 42 57
Zrzut ekranu 2025-12-10 o 10 43 06

The `allow` rule used in WebKit works as expected.

### DIFF
--- a/src/background/zapped.js
+++ b/src/background/zapped.js
@@ -14,6 +14,7 @@ import { store } from 'hybrids';
 import Config from '/store/config.js';
 import Options, { MODE_DEFAULT, MODE_ZAP } from '/store/options.js';
 
+import { isWebkit } from '/utils/browser-info.js';
 import { FLAG_MODES } from '/utils/config-types.js';
 import {
   getDynamicRulesIds,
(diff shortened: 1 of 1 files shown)


smalluban:
I've found that the issue is elsewhere. The `resourceTypes` must be inside the `condition` key. We can update it like this:

```js
condition: {
  excludedRequestDomains: excludedDomains,
  resourceTypes: ['main_frame'],
},
```

However, I wonder whether the current fix is a better solution, as it simplifies the DNR rule and works in both Chromium and Safari.

src/background/zapped.js
```diff
@@ -63,23 +62,12 @@ if (__PLATFORM__ !== 'firefox') {
...
-              id: 1,
-              priority: PAUSED_RULE_PRIORITY,
-              action: { type: 'allowAllRequests' },
-              condition: { excludedRequestDomains: excludedDomains },
-              resourceTypes: ['main_frame'],
-            },
+        {
+          id: 1,
+          priority: PAUSED_RULE_PRIORITY,
+          action: { type: 'allow' },
```
chrmod: we have to use `allowAllRequests` so that requests in iframes are blocked correctly

PR #2928, opened 2025-12-18
chore(deps): bump adblocker packages to 2.13.1

seia-soto:
This is the follow-up of https://github.com/ghostery/ghostery-extension/pull/2925. Last time, see details in https://github.com/ghostery/adblocker/pull/5361.


seia-soto:
The tests to prevent the same incident in the future is placed in `adblocker` repository: https://github.com/ghostery/adblocker/pull/5364. Despite it requires the external dependency from npm, it's the first place to prevent the same incident.

PR #2941, opened 2025-12-22
feat(onboarding): installation survey

smalluban:
Fixes https://github.com/ghostery/journal/issues/615

The survey is triggered for all browsers except Brave when a user closes the onboarding page while Ghostery has been enabled and the feature flag is turned on (`onboarding-survey`). 

For the safest solution, the listener for tab removal is attached when the onboarding is shown automatically after the installation. This code runs only once if the onboarding page is successfully displayed.

### Testing

Before closing the onboarding page, but after enabling Ghostery, open the Settings page and enable the `onboarding-survey` feature flag. Then, go back to the onboarding page and close the tab. You should see the survey page.

### DIFF
--- a/src/background/onboarding.js
+++ b/src/background/onboarding.js
@@ -8,17 +8,45 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
+import { store } from 'hybrids';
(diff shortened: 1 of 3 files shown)


mdsandu:
tested @smalluban. it works as expected- triggering on FF, not triggering on Brave
(I tested the versions you provided last week- FYI)

PR #2950, opened 2025-12-29
feat(telemetry): Add zap parameter

chrmod:
Telemetry values for Zap:
  - -1 = flag disabled
  - 0 = default (never touched)
  - 1 = zap (never touched)
  - 2 = default + touched
  - 3 = zap + touched

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -11,7 +11,7 @@
 
 import { store } from 'hybrids';
 
-import Options, { MODE_ZAP, MODE_DEFAULT } from '/store/options.js';
+import Options from '/store/options.js';
 import Config from '/store/config.js';
 import { debugMode } from '/utils/debug.js';
 import asyncSetup from '/utils/setup.js';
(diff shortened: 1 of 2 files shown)


src/background/telemetry/metrics.js
```diff
@@ -192,7 +194,18 @@ export default class Metrics {
       // Date of install (former install_date)
       buildQueryPair('id', this.storage.installDate) +
       // Toolbar pinned
-      buildQueryPair('tp', Number(conf.userSettings?.isOnToolbar ?? -1));
+      buildQueryPair('tp', Number(conf.userSettings?.isOnToolbar ?? -1)) +
+      // ZAP mode (-1 = flag disabled, 0 = default, 1 = zap, 2 = switched from zap to default)
+      buildQueryPair(
+        'zap',
+        !conf.config.hasFlag(FLAG_MODES)
```
smalluban, line 201: We have 3 states, and you should not rely on the remote config flag.

Mode set to `default`: 0
Mode set to `zap`: 1
Mode at least once has changed from `zap` to `default`: 2

src/background/telemetry/index.js
```diff
@@ -98,6 +103,12 @@ OptionsObserver.addListener(async function telemetry({ terms, feedback }) {
   } else {
     chrome.runtime.setUninstallURL('https://mygho.st/fresh-uninstalls');
   }
+
+  if (lastOptions?.mode === MODE_ZAP && mode === MODE_DEFAULT) {
```
smalluban: Move this into the `if (terms) { ... }` condition. Then, you can remove `setup.pending`

src/background/telemetry/metrics.js
```diff
@@ -192,7 +194,18 @@ export default class Metrics {
       // Date of install (former install_date)
       buildQueryPair('id', this.storage.installDate) +
       // Toolbar pinned
-      buildQueryPair('tp', Number(conf.userSettings?.isOnToolbar ?? -1));
+      buildQueryPair('tp', Number(conf.userSettings?.isOnToolbar ?? -1)) +
+      // ZAP mode (-1 = flag disabled, 0 = default, 1 = zap, 2 = switched from zap to default)
```
smalluban: "2" also includes mode set to zap, but `mode` was touched. It is worth mentioning in the comment.

PR #2960, opened 2026-01-03
feat: add option "fixesFilters"

### LINKED ISSUE #2952: Disable Ghostery specific fixes
This is a feature request to add an ability to disable Ghostery-specific fixes on websites. It's helpful whenever to compare "actual" Ghostery behaviour and let me create a fix from scratch easily as I don't need to create exceptions every time.

The below is the list of changes required:

- Use "ads" engine as base engine
- An ability to exclude "fixes" engine when merging engines
- New value to express if the flag is enabled

Suggested place to add this feature toggle is in the developer settings as this is not recommended for major user base. We can also add a broken page reports flag to do this as well, which gives us better potential to fix the issue without making complicated changes.

seia-soto:
fixes #2952

image

### Adblocker Engine

- Launch (or `npm run start -- --clean`) and enable extension
- Enable "Experimental filters"
- Open https://ghostery.com and check if `.ds-footer-copyrights` is visible (`document.querySelector('.ds-footer-copyrights').getBoundingClientRect()`)
- Disable "Ghostery specific fixes" and repeat the previous step

### DNR

- Launch (or `npm run start -- --clean`) and enable extension
- Enable "Experimental filters"
- Open https://www.srf.ch/play/tv/srfglobal/video/netanjahu---kriegsverbrecher-oder-politgenie?urn=urn:srf:video:fce1e024-0ddf-417e-8dda-126cb93c8f68 and check if requests to `data.srf.ch` are blocked
- Disable "Ghostery specific fixes" and repeat the previous step

### DIFF
--- a/tests/e2e/spec/managed.spec.js
+++ b/tests/e2e/spec/managed.spec.js
@@ -24,13 +24,19 @@ import {
 async function setManagedConfig(config) {
   await browser.url(getExtensionPageURL('panel'));
 
-  await browser.execute((managedConfig) => {
-    if (managedConfig) {
-      chrome.storage.local.set({ managedConfig });
-    } else {
-      chrome.storage.local.remove('managedConfig');
-    }
-  }, config);
+  await browser.execute(
+    (managedConfigStr) => {
+      const managedConfig = managedConfigStr
+        ? JSON.parse(managedConfigStr)
+        : null;
+      if (managedConfig) {
+        chrome.storage.local.set({ managedConfig });
+      } else {
+        chrome.storage.local.remove('managedConfig');
+      }
+    },
+    config ? JSON.stringify(config) : null,
+  );
 
   await waitForIdleBackgroundTasks();
 
--- a/tests/e2e/utils.js
+++ b/tests/e2e/utils.js
@@ -43,9 +43,9 @@ async function sendMessage(msg) {
 
   const result = await browser.execute(
     browser.isChromium
-      ? (msg) => chrome.runtime.sendMessage(msg)
-      : (msg) => browser.runtime.sendMessage(msg),
-    msg,
+      ? (msg) => chrome.runtime.sendMessage(JSON.parse(msg))
+      : (msg) => browser.runtime.sendMessage(JSON.parse(msg)),
+    JSON.stringify(msg),
   );
 
   if (result !== 'done') {


src/pages/settings/components/devtools.js
```diff
@@ -148,6 +148,19 @@ export default {
             <ui-line></ui-line>
           `}
           <ui-text type="headline-s">Developer tools</ui-text>
+          ${store.ready(options) &&
```
smalluban: Options are always ready in the Settings page, as the initial file checks options first here: https://github.com/ghostery/ghostery-extension/blob/main/src/pages/settings/index.js#L30
We can clean up other `store.ready()` calls, but this is not related to the PR.

src/store/options.js
```diff
@@ -66,6 +66,7 @@ const Options = {
   // Experimental features
   autoconsent: { autoAction: 'optOut' },
   experimentalFilters: false,
+  brokenPageFilters: true,
```
smalluban: We call it `fixes` in the extension, so might be `fixesFilters`?

src/background/adblocker.js
```diff
@@ -100,7 +100,7 @@ function getEnabledEngines(config) {
       list.push(...config.regionalFilters.regions.map((id) => `lang-${id}`));
     }
 
-    if (list.length) {
+    if (config.brokenPageFilters && list.length) {
```
smalluban: You have forgotten about DNR rules.

smalluban (changes requested):
If you add/edit UI, please add a corresponding screenshot of the changes.

src/background/dnr.js
```diff
@@ -107,7 +122,10 @@ if (__PLATFORM__ !== 'firefox') {
                   ),
             );
 
-            if (list.dnr.checksum !== resources.checksums['dnr-fixes']) {
+            if (
+              list.dnr.checksum !== resources.checksums['dnr-fixes'] &&
+              !lastOptions?.fixesFilters
```
smalluban: I don't understand addition of this condition. The function might run when other option changes, so the `fixesFilters` is still true. In this case `lastOptions.fixesFilter === true` and your condition resolves to `false`, so there will be no update, but it should. For example, when `filersUpdatedAt` changes (L106).

src/store/options.js
```diff
@@ -66,6 +66,7 @@ const Options = {
   // Experimental features
   autoconsent: { autoAction: 'optOut' },
   experimentalFilters: false,
+  fixesFilters: true,
```
smalluban, line 73: Add `fixesFilters` to `LOCAL_OPTIONS`.

smalluban (changes requested):

src/background/dnr.js
seia-soto: I removed the condition because the local resource checksum will be `null` after changing `fixesFilters`.

smalluban: I still see the `!lastOptions?.fixesFilters`

smalluban (changes requested):

src/background/dnr.js
seia-soto: Sorry, I didn't push the commit. Now live.

src/background/dnr.js
```diff
@@ -85,7 +86,21 @@ if (__PLATFORM__ !== 'firefox') {
       const DNR_FIXES_KEY = 'dnr-fixes';
       const resources = await store.resolve(Resources);
 
-      if (ids.length) {
+      if (options.fixesFilters === false) {
```
smalluban: If you change the condition in L103 to `options.fixesFilters && ids.length`, you can skip this removal section altogether, as from L201 there is the same logic as here in the `else` section.

smalluban (changes requested):
Please provide a new spec `experimental.spec.js` with two tests, one where nothing changes and fixes work as expected (when experimental filters are turned on), and when `fixesFilters` are turned off.
In debug mode (which runs on e2e tests), the dev tools are automatically opened. You must only add the `data-qa` attribute to the `fixesFilters` toggle.

src/background/dnr.js
```diff
@@ -81,8 +83,7 @@ if (__PLATFORM__ !== 'firefox') {
     const config = await store.resolve(Config);
 
     // Add latest fixes rules
-    if (config.hasFlag(FLAG_DYNAMIC_DNR_FIXES)) {
-      const DNR_FIXES_KEY = 'dnr-fixes';
+    if (options.fixesFilters && config.hasFlag(FLAG_DYNAMIC_DNR_FIXES)) {
```
smalluban: You should add `options.fixesFilters` in L89 - this condition is only for the feature flag and will be removed soon.

src/background/dnr.js
```diff
@@ -31,6 +31,7 @@ if (__PLATFORM__ !== 'firefox') {
     .getManifest()
     .declarative_net_request.rule_resources.filter(({ enabled }) => !enabled)
     .map(({ id }) => id);
+  const DNR_FIXES_KEY = 'dnr-fixes';
```
smalluban, line 34: I can't find a reason to move it here.

src/background/dnr.js
seia-soto: I thought that condition were not related to the changes from the first impression. However, I now realised that the possible value of `resources.checksums[DNR_FIXES_KEY]` is `null` or `some string`.

src/background/dnr.js
seia-soto: Some values were written as `dnr-fixes` as string literal and there's no need to redefined the value every time the options observer callback is called. Placing here reduces the times that variable is declared and prevent the value accidentally being changed somewhere else by using constant.

seia-soto:
The changes were applied and the testing filters marked as dangerous are dropped here https://github.com/ghostery/broken-page-reports/pull/2076

smalluban (approved):

seia-soto:
Depends on https://github.com/ghostery/ghostery-extension/pull/2975

seia-soto:
@smalluban Please, review this one as this one just rebased. Just to make everything sure.

Issue #2954, opened 2025-12-31
Subframe-only scripting filters


seia-soto:
https://github.com/ghostery/adblocker/pull/5104 introduces subframe only scripting capabilities for the extension. One thing that needs to be carefully approached is that this might introduce a contextual scripting, which wasn't the case so far.

PR #2961, opened 2026-01-05
feat: subframe scripting

seia-soto:
fixes https://github.com/ghostery/ghostery-extension/issues/2954 refs https://github.com/ghostery/adblocker/commit/9d5e120473f746db6a060a15804ac97ad53b474c

- [x] Create ancestor chains
- [x] Create E2E test
- [x] Update ancestor list on frame creates
- [x] Handle duplicates on returning ancestor chain
- [x] Handle on tabs closed https://developer.chrome.com/docs/extensions/reference/api/tabs#event-onRemoved
- [x] Handle on tabs replaced https://developer.chrome.com/docs/extensions/reference/api/tabs#event-onReplaced
- [x] Describe the side-effects of using `documentId`
- [x] Describe the side-effects of using `chrome.runtime.onMessage` (missing `parentFrameId` information): we're not neglecting any information because we're trying to inject from multiple events.
- [x] Ensure the Firefox content scripts API usage with the ancestor chain (e.g. is `contentScript.isRegistered` safe?)
- [x] Test incomplete request situation which removes the whole tab information
- [x] Synchronise ancestors structure on the service worker restart (or when the context is lost)
- [x] Introduce AB flag https://github.com/ghostery/broken-page-reports/pull/2095
- [x] Group adblocker related files
- [x] Refactor to Class instance
- [x] Refactor tests to work with Class instance
- [x] Test documentId tracking

### DIFF
--- a/src/background/adblocker/index.js
+++ b/src/background/adblocker/index.js
@@ -14,7 +14,6 @@ import {
   filterRequestHTML,
   updateResponseHeadersWithCSP,
 } from '@ghostery/adblocker-webextension';
-import { parse } from 'tldts-experimental';
 import scriptlets from '@ghostery/scriptlets';
 import {
   FLAG_FIREFOX_CONTENT_SCRIPT_SCRIPTLETS,
@@ -31,7 +30,7 @@ import * as exceptions from '/utils/exceptions.js';
 import * as engines from '/utils/engines.js';
 import * as trackerdb from '/utils/trackerdb.js';
 import * as OptionsObserver from '/utils/options-observer.js';
-import Request from '/utils/request.js';
+import Request, { parseWithCache } from '/utils/request.js';
 import asyncSetup from '/utils/setup.js';
 
 import { tabStats, updateTabStats } from '../stats.js';
@@ -360,7 +359,7 @@ async function injectCosmetics(details, config) {
 
   const { tabId, frameId, parentFrameId, documentId, url } = details;
 
-  const parsed = parse(url);
+  const parsed = parseWithCache(url);
   const domain = parsed.domain || '';
   const hostname = parsed.hostname || '';
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,14 +9,14 @@
       "version": "10.5.28",
       "license": "GPL-3.0",
       "dependencies": {
-        "@duckduckgo/autoconsent": "^14.51.0",
+        "@duckduckgo/autoconsent": "^14.52.0",
         "@ghostery/adblocker": "^2.13.4",
         "@ghostery/adblocker-content": "^2.13.3",
         "@ghostery/adblocker-extended-selectors": "^2.13.3",
         "@ghostery/adblocker-webextension": "^2.13.4",
         "@ghostery/config": "^0.0.7",
         "@ghostery/scriptlets": "github:ghostery/scriptlets",
-        "@ghostery/urlfilter2dnr": "^1.6.5",
+        "@ghostery/urlfilter2dnr": "^1.6.6",
         "@github/relative-time-element": "^5.0.0",
         "@sentry/browser": "^10.38.0",
         "@whotracksme/reporting": "^7.10.0",
@@ -28,7 +28,7 @@
         "idb": "^8.0.2",
         "lottie-web": "^5.13.0",
         "plotly.js-basic-dist": "^3.3.1",
-        "tldts-experimental": "^7.0.19"
+        "tldts-experimental": "^7.0.22"
       },
       "devDependencies": {
         "@eslint/js": "^9.39.2",
@@ -38,11 +38,11 @@
         "@wdio/globals": "^9.12.7",
         "@wdio/mocha-framework": "^9.12.6",
         "@wdio/spec-reporter": "^9.23.2",
-        "addons-linter": "^9.5.0",
+        "addons-linter": "^9.6.0",
         "eslint": "^9.39.2",
         "eslint-config-prettier": "^10.1.8",
         "eslint-plugin-prettier": "^5.5.5",
-        "globals": "^17.2.0",
+        "globals": "^17.3.0",
         "license-report": "^6.8.1",
         "prettier": "^3.8.1",
         "sinon-chrome": "^3.0.1",
@@ -516,9 +516,9 @@
       }
     },
     "node_modules/@duckduckgo/autoconsent": {
-      "version": "14.51.0",
-      "resolved": "https://registry.npmjs.org/@duckduckgo/autoconsent/-/autoconsent-14.51.0.tgz",
-      "integrity": "sha512-FgaCJQGqbmc71N9LjTWtmVcTVJQbKBDTC9av8S78qOJkIT8t/l7PEsuiJGf7IfzWFjh0RGtY3wByi6wtW99XeA==",
+      "version": "14.52.0",
+      "resolved": "https://registry.npmjs.org/@duckduckgo/autoconsent/-/autoconsent-14.52.0.tgz",
+      "integrity": "sha512-Xg38Ypoe9S1ESZYSEVNKGF3Fn8hSEdTHsz9rQ97Pa+d8Z0HtgrE1OpBt5C/6ZUCMLSdn6BOcvaULt0A6/OTLsg==",
       "license": "MPL-2.0",
       "dependencies": {
         "@ghostery/adblocker": "^2.0.4",
@@ -1387,9 +1387,9 @@
       }
     },
     "node_modules/@ghostery/urlfilter2dnr": {
-      "version": "1.6.5",
-      "resolved": "https://registry.npmjs.org/@ghostery/urlfilter2dnr/-/urlfilter2dnr-1.6.5.tgz",
-      "integrity": "sha512-b3z1d2Wp9vpTNROzgw8advp0O2iwDYaI7jJaky9yj2Y9H49u+ava7QXwo5Erf63rBbMlCIFeA/ZLCyw3RfaKmA==",
+      "version": "1.6.6",
+      "resolved": "https://registry.npmjs.org/@ghostery/urlfilter2dnr/-/urlfilter2dnr-1.6.6.tgz",
+      "integrity": "sha512-VBEqR83pFjTBNgFS2Ro5fmhAw7BL7LI8UoGjRvpVjWX4kmErCatgXGagijLLujqMjJ9VkhXSjYBO4GA9+JJU4Q==",
       "license": "GPL-3.0",
       "dependencies": {
         "@adguard/re2-wasm": "^1.2.0",
@@ -2198,9 +2198,9 @@
       "license": "MIT"
     },
     "node_modules/@mdn/browser-compat-data": {
-      "version": "7.2.4",
-      "resolved": "https://registry.npmjs.org/@mdn/browser-compat-data/-/browser-compat-data-7.2.4.tgz",
-      "integrity": "sha512-qlZKXL9qvrxn2UNnlgjupk1sjz0X59oRvGBBaPqYtIxiM0q4m2D5ZF9P/T0qWm0gZYzb5jMZ1TpUViZZVv7cMg==",
+      "version": "7.3.0",
+      "resolved": "https://registry.npmjs.org/@mdn/browser-compat-data/-/browser-compat-data-7.3.0.tgz",
+      "integrity": "sha512-B7X53zLkrQOVNzsWg4d+iE5vbrreOYB9H/2Ikeit9LPb3tqRNRsEAxysQduFLODBJBSgFGr2FkJLq7tMQkqLZQ==",
       "dev": true,
       "license": "CC0-1.0"
     },
@@ -4094,20 +4094,20 @@
       }
     },
     "node_modules/addons-linter": {
-      "version": "9.5.0",
-      "resolved": "https://registry.npmjs.org/addons-linter/-/addons-linter-9.5.0.tgz",
-      "integrity": "sha512-jV3gaZ9tLvLJQZyyZTLbDVFoKPA1lA7LmsCxZG62OKf5Y2ZMdTposSMFffnnMMQsjqHY7TP3DwTPwRP4e80w9A==",
+      "version": "9.6.0",
+      "resolved": "https://registry.npmjs.org/addons-linter/-/addons-linter-9.6.0.tgz",
+      "integrity": "sha512-hoZFDg/4Ial7xw2SDrnQJ6iyLtf0KdP/NRhLP1bX+SKNmFLJskDIvWOCbDAC1qUq7cNrt4yCvGlr3UO0VZaluw==",
       "dev": true,
       "license": "MPL-2.0",
       "dependencies": {
         "@fluent/syntax": "0.19.0",
         "@fregante/relaxed-json": "2.0.0",
-        "@mdn/browser-compat-data": "7.2.4",
+        "@mdn/browser-compat-data": "7.3.0",
         "addons-moz-compare": "1.3.0",
         "addons-scanner-utils": "10.2.0",
         "ajv": "8.17.1",
         "chalk": "4.1.2",
-        "cheerio": "1.1.2",
+        "cheerio": "1.2.0",
         "columnify": "1.6.0",
         "common-tags": "1.8.2",
         "deepmerge": "4.3.1",
@@ -4119,7 +4119,7 @@
         "fast-json-patch": "3.1.1",
         "image-size": "2.0.2",
         "json-merge-patch": "1.0.2",
-        "pino": "10.2.1",
+        "pino": "10.3.0",
         "semver": "7.7.3",
         "source-map-support": "0.5.21",
         "upath": "2.0.1",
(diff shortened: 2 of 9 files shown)


tests/unit/ancestors.test.js
```diff
@@ -0,0 +1,335 @@
...
+    unregister(1, 0);
+    // Closes tab `foo.com`
+    unregister(0, 0);
+    // Closes tab `baz.com`
+    unregister(2, 0);
+
+    assert.deepEqual(tabs, []);
+  });
+
+  it('handles 100 tabs each one holding 200 frames', () => {
```
seia-soto, line 423: This test is responsible for the potential performance measurement and can be removed in the future. From my machine (M1 16GB), it takes about 14ms to process everything.

src/utils/ancestors.js
```diff
@@ -0,0 +1,129 @@
...
+      tabs.splice(tabIndex, 1);
+      return;
+    }
+
+    const frames = tabs[tabIndex].frames;
+    const parents = [frameId];
+
+    let frameIndex = 0;
+    while (parents.length) {
+      const parent = parents.shift();
```
seia-soto: We might prefer using `Array.prototype.pop` instead.

src/background/adblocker/index.js
```diff
@@ -348,7 +354,7 @@ async function injectCosmetics(details, config) {
     return;
   }
 
-  const { frameId, url, tabId } = details;
+  const { frameId, parentFrameId, url, tabId } = details;
```
seia-soto: This `parentFrameId` property is not available from `chrome.runtime.onMessage`.

src/utils/ancestors.js
```diff
@@ -0,0 +1,129 @@
...
+      // continue the outer loop.
+      while (--frameIndex > -1) {
+        if (frames[frameIndex].id === parentFrameId) {
+          chain.push(frames[frameIndex].details);
+          parentFrameId = frames[frameIndex].parent;
+
+          // If it reached the top-most frame, exit with the
+          // result.
+          if (parentFrameId === -1) {
+            return chain.reverse();
```
seia-soto: We technically don't need this `Array.prototype.reverse` call because adblocker library doesn't care about the ordering of the frame.

seia-soto: This only affects the test code.

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    // If it ends up detecting an incomplete hierarchy, we drop
+    // all the details to prevent further resource use.
+    tabs.splice(tabIndex, 1);
+
+    return [];
+  }
+
+  function unregister(tabId, frameId) {
```
chrmod: when this is called?

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+export const createAncestorsList = () => {
+  const tabs = [];
```
chrmod: looks like a potential memory leak

src/background/adblocker/index.js
```diff
@@ -372,13 +378,22 @@ async function injectCosmetics(details, config) {
 
   const engine = engines.get(engines.MAIN_ENGINE);
 
+  let ancestors = undefined;
+  if (typeof parentFrameId === 'number') {
```
seia-soto: The type of `parentFrameId` could be `number` or `undefined`.

seia-soto:
Depends on https://github.com/ghostery/ghostery-extension/pull/2975

src/background/adblocker/ancestors.js
seia-soto: We might want to run a periodic clean-up

src/background/adblocker/ancestors.js
seia-soto: Corresponding event listeners are added now.

seia-soto:
For subframes, we can try to import existing frameIds from Chrome extension API: rebuilding.

src/background/adblocker.js
```diff
@@ -487,6 +512,23 @@ chrome.runtime.onMessage.addListener((msg, sender) => {
   }
 });
 
+// Listen for tab changes to maintain ancestor chain.
+chrome.tabs.onRemoved.addListener((tabId) => {
```
smalluban: Move those listeners up to the `hierarchy` definition (after L86).

chrmod: or we move it to ancestors.js

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,161 @@
+/**
```
chrmod, line 1: I would prefer to create `background/adblocker` folder and put that file there.

src/background/adblocker/index.js
chrmod: maybe we can afford to get it from chrome.webNavigation?

seia-soto:
@chrmod implemented frame details syncing upon the service worker starts.

src/background/adblocker/index.js
seia-soto: I wonder if that's going to be fast enough or effective at the end. We will need to test it but not sure how yet.

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,188 @@
...
+      // Pick one `parent` and remove every frames having
+      // `parent` as parent frame ID or frame ID.
+      const parent = parents.pop();
+      for (frameIndex = 0; frameIndex < frames.length; frameIndex++) {
+        if (
+          frames[frameIndex].parent === parent ||
+          frames[frameIndex].id === parent
+        ) {
+          parents.push(frames[frameIndex].id);
+          frames.splice(frameIndex, 1);
```
philipp-classen, line 178: Here, you have to be careful since removing the element will shift the indices of the following items.

Iterating the list in reverse order is one way to avoid that problem.

src/background/adblocker/index.js
```diff
@@ -82,6 +83,59 @@ const contentScripts = (() => {
   };
 })();
 
+const hierarchy = createAncestorsList();
+
+// Start syncing tab information as the service worker started.
+chrome.tabs
```
chrmod: how about we replace `createAncestorsList` with a class. That way we could put `chrome.tabs.query` and listeners on the class so all can be encapsulated in the `ancerstors.js`?

src/background/adblocker/ancestors.js
seia-soto: Thanks, fixed in 20f75ebf3cc78ec1a1dab2fd9b60438ddf1d65d0

src/background/adblocker/index.js
seia-soto: I'd love to separate the logic from webextension APIs for the ease of testing and imports. Would it be better to create another file for that?

tests/unit/ancestors.test.js
```diff
@@ -0,0 +1,409 @@
...
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
```
philipp-classen: Nitpicking: I see, that legacy mode it used by the existing tests as well, so there is some argument for consistency. But we could migrate to strict assertions, which is best practice as I understand it:

import assert from 'node:assert/strict';

https://nodejs.org/api/assert.html#strict-assertion-mode

(I think, assert.deepStrictEqual -> assert.deepEqual and nothing else would change here.)

src/background/adblocker/index.js
philipp-classen: I would agree that having a class would be more natural. Even though, the existing ancestors.js file is structurally equivalent (before the "class" syntax was introduced, it was done like that).

But moving the listener registration into a class method that can be called, would make it more obvious that they belong together. In other words, making it an implementation detail of the class.

Basically, this part

```
const hierarchy = createAncestorsList();
// Start syncing tab information as the service worker started.
chrome.tabs// Using empty object will query all tabs in all windows.
  .query({})
  .then(async function (tabs) {
...
chrome.tabs.onReplaced.addListener((addedTabId, removedTabId) => {
  if (FIREFOX_CONTENT_SCRIPT_SCRIPTLETS.enabled) {
    return;
  }

  hierarchy.replace(removedTabId, addedTabId);
});
```

... could turn into something like (maybe we could also consider the names):

const hierarchy = new AncestorList();
hierarchy.initListeners();

And initListener would be responsible for setting up the event listeners, and if necessary, trigger also the async part (chrome.tabs.query). But it is important that it stays a synchronous function, so the event listener are registered on the first tick (as it is correctly down in the current version). In other words, an API like this would potentially break:

const hierarchy = new AncestorList();
await hierarchy.initListeners(); // bad: should stay synchronous

When it comes to testing, the requirement that event listeners have to be registered on the first tick can be restrictive (because it enforces that the initialization is hard-coded and cannot be swapped out easily). But I'm not aware of a good way to avoid that (meaning your current approach is fine from a testing perspective, as far as I can tell).

src/background/adblocker/index.js
```diff
@@ -82,6 +83,59 @@ const contentScripts = (() => {
...
+
+      hierarchy.sync(
+        tab.id,
+        (await chrome.webNavigation.getAllFrames({ tabId: tab.id })).map(
+          decorateFrame,
+        ),
+      );
+    }
+
+    await Promise.allSettled(tabs.map(syncAllFrames));
```
philipp-classen: Hard to think it through just by reading, but this part could be potentially introduce races. Given a list of tabs, the update of each individual tab runs asynchronously (meaning also that during the update, the list of tabs of the original snapshot might be outdated.

Eliminate the potential for such races is difficult. In short, once you use any async operation, it is potentially racy. As the APIs come from the browser and we have to work with what exists, it is sometimes unavoidable to use async calls.

One approach is to accept races and deal with inconsistencies gracefully (e.g. assume that tabs are already closed, and check against documentId if available). But it is hard to reason about it, of course, because it means operating always on a snapshot of the world, which is no longer matching the reality.

Mixing different API (e.g. chrome.tabs and chrome.webNavigation) will have additional potential for races, since the events can be interleaved in an unpredictable way. Again, mixing APIs may be unavoidable if the information that you need is not provided by a single API.

---

Since a lot of what I wrote is very likely unavoidable, my only concrete recommendation is to consider including the documentId in the bookkeeping. Unfortunately, it is unavailable on Firefox; but on Chrome it could help to detect inconsistencies (due to races).

tests/unit/ancestors.test.js
```diff
@@ -0,0 +1,692 @@
...
+      );
+      // Assume there's a pre-rendering frame as the user types
+      // into the omnibox or the address bar.
+      assert.deepStrictEqual(
+        hierarchy.ancestors(
+          { tabId: 0, frameId: 1_000, parentFrameId: 0, documentId: 'Z0' },
+          'foo.com',
+        ),
+        ['about:blank'],
+      );
```
seia-soto, line 613: Not sure if a pre-rendering frame is linked to the main frame on Safari. I need to double check this.

src/background/adblocker/index.js
seia-soto: I'm still not sure if I should lock the modification to the list structure to prevent that. Also, I'm afraid if any other events will fire at the moment.

Currently, this approach is dropping the specific structure (not actually the full but by tracking parent node) using `frameId` when `parent` is not found or the list of ancestors cannot be exist in the complete form.

tests/unit/mocks/browser.js
```diff
@@ -0,0 +1,229 @@
...
+};
+
+const NEWTAB_URL = {
+  [BROWSER_KIND.chromium]: 'chrome://newtab',
+  [BROWSER_KIND.firefox]: 'about:newtab',
+  // Safari shows `favorites://` from `location.href` but they
+  // will use `about:blank` when preloading top-hit from their
+  // address bar.
+  [BROWSER_KIND.safari]: 'about:blank',
+};
```
seia-soto: This is to describe the behaviour from the browser as much as possible. This is mostly TBD changes so we can skip for now.

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,359 @@
...
+      );
+    }
+  }
+
+  // This section needs to be separated to keep the codepath from
+  // testing code clean. Please, keep the event listeners as-is.
+  // We don't want to sacrifice the ease of understanding. Also,
+  // we need to resolve "flag" which will cause prop drilling.
+  handleWebextensionEvents(FIREFOX_CONTENT_SCRIPT_SCRIPTLETS) {
+    chrome.tabs.onRemoved(function (tabId) {
```
philipp-classen:
```suggestion
    chrome.tabs.onRemoved.addListener((tabId) => {
```

Note: "=>" is also important here, since "this.unregister" will otherwise fail.

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,359 @@
...
+      if (FIREFOX_CONTENT_SCRIPT_SCRIPTLETS.enabled === false) {
+        this.unregister(tabId, 0);
+      }
+    });
+    chrome.tabs.onReplaced(function (addedTabId, removedTabId) {
```
philipp-classen:
```suggestion
    chrome.tabs.onReplaced.addListener((addedTabId, removedTabId) => {
```

Again, "=>" is to make "this.replace" work.

seia-soto:
From testing with Chromium, I found that we need to test against main frame refreshes. It will nuke all subframes as well.

seia-soto:
I merged changes from #3021 thanks to @philipp-classen .

seia-soto:
Pending https://github.com/ghostery/ghostery-extension/pull/3030

seia-soto:
## Test instructions

1. Launch an extension using this branch: `git checkout subframe-scripting && npm ci` then `npm run start -- chromium`.
2. Enable extension and disable "Never-Consent". We're going to test this using consent filters.
3. Visit one of the website listed without any filters.
4. Apply given filter and refresh the website to compare the behaviour.
5. If you need to clean website data, open devtools then access "Application". You can find "Clear site data" with "including 3rd-party cookies" in that panel.

Please, note that you might not see the consent popup depending on the region. Not all the region enforces the cookie consent.

**Notice**

The filters are coming from uAssets, special thanks to uBlock Origin filter contributors.

### `https://www.weltkunst.de/`

```
weltkunst.de>>##+js(trusted-click-element, button[title="ALLE AKZEPTIEREN"], , 1000)
```

Expected: the filter should handle the consent popup automatically.

### `https://tidende.dk/`

```
tidende.dk>>##+js(trusted-click-element, button[title="TILLAD NØDVENDIGE"], , 1000)
```

Expected: the filter should handle the consent popup automatically.

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  async handleWebWorkerStart({ maxRetries = 25 } = {}) {
+    let allTabs;
+    for (let i = 0; i < maxRetries; i++) {
+      try {
+        allTabs = await chrome.tabs.query({});
+        break;
+      } catch (e) {
+        if (e.message.includes('(user may be dragging a tab)')) {
```
chrmod: this is very browser version specific and it does not look like a common problem. Lets remove it for now (and the retry logic)

seia-soto: removed.

src/background/adblocker/index.js
```diff
@@ -89,6 +90,11 @@ if (__PLATFORM__ === 'firefox') {
   );
 }
 
+const hierarchy = new FramesHierarchy();
+
+hierarchy.handleWebextensionEvents(FIREFOX_CONTENT_SCRIPT_SCRIPTLETS);
```
smalluban, line 98: Why is it not in the `constructor()` method of the `FramesHierarchy` class?

You can hide details, and only run

```js
const hierarchy = new FramesHierarchy(FIREFOX_CONTENT_SCRIPT_SCRIPTLETS);
```

BTW, I think it is a proper time to clean up `FIREFOX_CONTENT_SCRIPT_SCRIPTLETS` flag (in another PR) - @chrmod . This complicates new logic, but it is constantly `true` from 8 months (https://github.com/ghostery/broken-page-reports/commit/8068f0462b54e250958f5eecfa24016764851113).

smalluban (changes requested):

src/background/adblocker/index.js
philipp-classen: > Why is it not in the `constructor()` method of the `FramesHierarchy` class?

Could be matter of taste, but having the installation of the event listener outside of the class, is not uncommon (maybe "installEventListeners" would be a better name). But I can also see the setup happening in the constructor. Might be a bit surprising though that the constructor has side-effects then.

seia-soto: My opinion towards that structure is that we don't need to expect any side-effects from event listeners, keeping test code simple.

src/background/adblocker/ancestors.js
```diff
@@ -0,0 +1,265 @@
...
+        `Failed to get frames of the tab: tabId="${tab.id}"`,
+        error,
+      );
+      return;
+    }
+
+    this.sync(
+      tab.id,
+      frames.map((frame) => {
+        const parsed = parse(frame.url);
```
chrmod: I'm bit worried about that as we do it for all URL. Isn't it done already in the adblocker/index.js? dont we have a cached API for it? LRU cache could help. What do you think @philipp-classen @smalluban

smalluban: Good catch. We have utility here: https://github.com/ghostery/ghostery-extension/blob/main/src/utils/request.js (it must be exported and can be used).

seia-soto: resolved in bf992e9fc446cd3a6e26199c2828430872849f0b

PR #2962, opened 2026-01-07
feat(notifictions): add review notification

smalluban:
Fixes https://github.com/ghostery/journal/issues/632

We were planning to set the shown flag before displaying the notifications. The issue with this approach is that, when we try to display the notification, we check if another notification is already displayed. This is very unlikely, as our conditions for displaying notifications are very different. However, if we set the flag before processing the notification, we can set the flag for the case when the notification was not displayed. 

### Testing

Run locally or use the DEBUG version of the extension. Go to settings and enable the `review-notification` flag. Open any website in a new tab. You should see the review notification.


chrmod (changes requested):
Lets move all the notification files from background root into `notifications` subfolders

chrmod (approved):

PR #2964, opened 2026-01-07
feat(telemetry): install_complete event

chrmod:
1. New install_complete telemetry signal

  - Purpose: Measure day 0 retention by tracking when users complete onboarding
  - Trigger: Sent when user reaches the onboarding success page via the component's connect hook
  - Files: metrics.js (handler), success.js (sender)

  2. active signal now sends on install day

  - Previously, the active ping was skipped on the first day due to initialization logic
  - Now active is excluded from that logic (like engaged), so it sends immediately

  3. active signal delayed after install

  - Added a 10-minute cooldown: if install was sent less than 10 minutes ago, active ping is skipped
  - Prevents active from firing immediately alongside install on startup

### DIFF
--- a/src/background/telemetry/metrics.js
+++ b/src/background/telemetry/metrics.js
@@ -16,6 +16,8 @@ import { FLAG_MODES } from '@ghostery/config';
 import getDefaultLanguage from './language.js';
 import getBrowserInfo from '/utils/browser-info.js';
 
+const ACTIVE_PING_DELAY = 10 * 60 * 1000; // 10 mintues
+
 /**
  * Allows to run async operations one by one (FIFO, first-in first-out).
  * The execution of function will only be started once all previously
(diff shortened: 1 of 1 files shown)


src/background/telemetry/index.js
```diff
@@ -93,8 +95,9 @@ OptionsObserver.addListener(async function telemetry(
   if (terms) {
     setup.pending && (await setup.pending);
 
-    if (runner.isJustInstalled()) {
+    if (!runner.storage.install_all) {
```
smalluban: `runner.ping('install')` has included a check for it. If you want to use a condition here for an alarm, use `runner.isJustInstalled()` method. Now you are using the internal structure of the storage.

src/pages/onboarding/views/success.js
```diff
@@ -23,6 +23,11 @@ import pinExtensionOpera from '../assets/pin-extension-opera.jpg';
 let screenshotURL = '';
 let type = '';
 
+chrome.runtime.sendMessage({
```
smalluban: This is called in the module body, just after loading the file. I expect it is not what you wanted. 

You should use `connect` method of the `render` property:

```js
{
  ...,
  render: {
    value: (options) => html...,
    connect() {
      chrome.runtime.sendMessage(...);
    },
  },
}
```

PR #2986, opened 2026-01-14
feat(telemetry): attribution method

chrmod:
Introduces a new attribution tracking method that prioritizes reading attribution data from cookies on Ghostery domains, with fallback to UTM parameters from browser store pages.

- New attribution module (attribution.js): Detects install source via cookies or UTM parameters
- Refactored telemetry: Moved attribution detection to post-install for better reliability
- Dev tools UI: Added attribution info display (install date, source, campaign)
- E2E tests: Added test coverage for cookie-based attribution

### DIFF
--- a/tests/e2e/wdio.update.conf.js
+++ b/tests/e2e/wdio.update.conf.js
@@ -38,6 +38,11 @@ import { setupTestPage } from './page/server.js';
  */
 export const config = {
   ...wdio.config,
+  exclude: [
+    // The `attribution.spec.js` tests relates to the code running only on the first install
+    // and can't be run during the update process.
+    './spec/attribution.spec.js',
+  ],
   onPrepare: async (config, capabilities) => {
     if (wdio.argv.clean) {
       rmSync(wdio.WEB_EXT_PATH, { recursive: true, force: true });


src/background/telemetry/attribution.js
```diff
@@ -0,0 +1,78 @@
...
+  if (source && campaign) {
+    return { utm_source: source, utm_campaign: campaign };
+  }
+  return null;
+}
+
+async function getAttributionFromCookies() {
+  for (const domain of ATTRIBUTION_DOMAINS) {
+    try {
+      const cookies = await chrome.cookies.getAll({
```
smalluban: Why getAll if you set the `name`?

src/background/telemetry/index.js
```diff
@@ -99,6 +68,15 @@ OptionsObserver.addListener(async function telemetry(
     setup.pending && (await setup.pending);
 
     if (runner.isJustInstalled()) {
+      try {
+        const attribution = await detectAttribution();
```
smalluban, line 72: We should exclude Safari here.

src/pages/settings/components/devtools.js
```diff
@@ -242,6 +252,32 @@ export default {
               </div>
             </div>
           `}
+          ${telemetry &&
```
smalluban: Use:

```js
html.resolve(getTelemetryStorage().then(telementry => html` ... `))
```

Setting the component property has no benefit, as the value is not reactive (static resolved promise).

src/pages/settings/components/devtools.js
```diff
@@ -290,7 +326,7 @@ export default {
             <ui-text type="headline-s">Actions</ui-text>
             <div layout="row gap items:start">
               <ui-button onclick="${clearStorage}" layout="shrink:0" size="s">
-                <button>
+                <button data-qa="button:clear-storage">
```
smalluban: You must be aware that this button reloads the extension, so the current tab is closed within the action. This can affect e2e tests.

src/pages/settings/views/privacy.js
```diff
@@ -321,6 +321,7 @@ export default {
             <settings-devtools
               onshown="${html.set('devMode', true)}"
               visible="${devMode}"
+              data-qa="component:devtools"
```
smalluban: This is not needed, as e2e tests run in the `debug` mode, which opens the dev tools by default.

tests/e2e/spec/attribution.spec.js
```diff
@@ -0,0 +1,35 @@
...
+  getExtensionPageURL,
+} from '../utils.js';
+
+describe('Attribution', () => {
+  before(enableExtension);
+
+  it('captures attribution from ghostery.com cookie', async () => {
+    await browser.url(getExtensionPageURL('settings'));
+    const version = await getExtensionElement('component:devtools', 'ui-text');
+    for (let i = 0; i < 6; i++) {
```
smalluban: This is not needed. See another comment.

tests/e2e/utils.js
```diff
@@ -91,6 +91,15 @@ export async function reloadExtension() {
 }
 
 export async function enableExtension() {
+  /* attribution.spec */
+  await browser.url('https://www.ghostery.com/');
```
smalluban: Move this to `before()` inside of the `attribution.spec.js`. You can run custom function, which runs `enableExtension()` after this code.

src/pages/settings/components/devtools.js
smalluban: FYI: I plan to add support for `Promise` natively. I discovered that fallback view is rarely used (the `html.resolve` accepts more arguments, which allow for displaying of content before the promise is resolved).

tests/e2e/utils.js
smalluban: Do we have to visit the page to set the cookie? It is not a good idea to relay on `ghostery.com` for every test.

smalluban: I see that this must be added before the very first try to enable the extension. We can keep it here, but if possible, remove the step of visiting the Ghostery website.

src/background/telemetry/index.js
chrmod: why?

smalluban:
The tests might fail because of opening ghostery.com - as currently it is opened every time (regardless if it is already enabled), so the notification is shown too early. Changes highlighted in comments should fix the issue.

Please describe the PR :)

src/background/telemetry/index.js
smalluban: 1. On Safari, there is no Web Store from which to download an extension. Getting this attribution makes no sense. 
2. Your previous version did not have `storeId`, so it would not work - that was one of my intentions too.

tests/e2e/utils.js
```diff
@@ -91,6 +91,14 @@ export async function reloadExtension() {
 }
 
 export async function enableExtension() {
+  /* attribution.spec */
+  await browser.setCookies({
```
smalluban: You should do it only if the extension must be enabled. Move this to L105

src/background/telemetry/index.js
smalluban: Ok, I see that there is a change to open the App Store on the web, and then navigate to the native App Store. We can close this.

smalluban:
There is a logic problem with running your test in `the e2e-update` job. It won't work there, as the extension is already enabled in the production version, which does not use the `enableExtension()` function (by design). I think you can move the setting of the cookie to the `wdio.conf.js` file in L188

smalluban (approved):

PR #2997, opened 2026-01-16
fix(utils): use ghostery.com direct download URL

### LINKED ISSUE #2980: Use ghostery.com direct links for extension download
We should replace the `REVIEW_PAGE_URL` complex logic with a direct link to `ghostery.com`.

smalluban:
Fixes #2980

* panel: `utm_source=gbe&utm_campaign=panelreview`
* notification `utm_source=gbe&utm_campaign=review`

### DIFF
--- a/src/utils/urls.js
+++ b/src/utils/urls.js
@@ -21,8 +21,8 @@ export const WTM_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/whotracksme`;
 export const SUPPORT_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/support`;
 export const WHATS_NEW_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/blog/ghostery-extension-v10-5?embed=1&utm_campaign=whatsnew`;
 
-export const PANEL_STORE_PAGE_URL = `${HOME_PAGE_URL}downloads/extension?utm_source=gbe&utm_campaign=panel`;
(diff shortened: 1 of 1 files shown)


src/utils/urls.js
```diff
@@ -21,16 +21,8 @@ export const WTM_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/whotracksme`;
 export const SUPPORT_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/support`;
 export const WHATS_NEW_PAGE_URL = `https://www.${GHOSTERY_DOMAIN}/blog/ghostery-extension-v10-5?embed=1&utm_campaign=whatsnew`;
 
-export const REVIEW_PAGE_URL = (() => {
-  if (__PLATFORM__ === 'firefox') return 'https://mygho.st/ReviewFirefoxPanel';
```
chrmod, line 25: this does not work as Firefox has dedicated review page https://addons.mozilla.org/en-US/firefox/addon/ghostery/reviews/

So we may want to solve it by having a separate redirect on ghostery.com just for reviews. lets say `ghostery.com/extension-reviews?utm....`

chrmod (changes requested):

src/utils/urls.js
smalluban: Oh, I did not catch that. Thanks for that. I will prepare the change on ghostery.com first.

smalluban: Updated.

smalluban:
@chrmod URLs updated to use `downloads/review` from the website. Tested on staging, it works as expected.

Issue #3004, opened 2026-01-21
Missing keywords comment section in the redirect resources


seia-soto:
Some websites like `49ers.com` will try to detect ad-blocker by comparing the actual content of the advertisement script. See https://github.com/ghostery/broken-page-reports/pull/2105/changes#diff-7080bf5960be22b376a66e3a72dcbbec1e5a65107480f4e97b911f761c331330R2372 for the current mitigation.

Ghostery bringing uBlock Origin resources, it doesn't include "comments" from the redirect resources. In that specific case, I had to change the query.

```txt
49ers.com##+js(trusted-replace-argument, String.prototype.includes, 0, repl:/googleads/google/)
```

The comment section format is like below:

```
...

/*
keyword1,keyword2,...
*/
```

The fix is already ongoing the backend but this issue allows us to expose the issue and build new test against it.

PR #3005, opened 2026-01-21
test: google-ima.js

seia-soto:
fixes https://github.com/ghostery/ghostery-extension/issues/3004


seia-soto:
The added test is meant to be failed but it will pass as soon as we deploy the changes in the backend.

seia-soto:
Waiting for another staging build.

seia-soto:
Data dependency was pinned. Updated to new value.

PR #3021, opened 2026-01-27
Attempt to simplify

philipp-classen:
Changes intended to simplify, but to keep the original implementation semantically unchanged (with few exceptions, see comments).

### DIFF
--- a/src/background/adblocker/ancestors.js
+++ b/src/background/adblocker/ancestors.js
@@ -57,7 +57,10 @@ export class FramesHierarchy {
         targetFrame.documentId = documentId;
         targetFrame.details = details;
         break;
-      } else if (frames[frameIndex].documentId === documentId) {
+      } else if (
+        documentId.length &&
+        frames[frameIndex].documentId === documentId
+      ) {
         this.#handleFrameReplacement(
           frames,
           frameIndex,


src/background/adblocker/ancestors.js
```diff
@@ -47,174 +39,138 @@ export class FramesHierarchy {
...
+    // If frame not found, add it
     if (frameIndex === -1) {
-      framesLength =
-        frames.push({
-          id: frameId,
-          parent: parentFrameId,
-          documentId,
-          details,
-        }) - 1;
+      frames.push({
```
philipp-classen, line 78: Note: this is a semantic change. framesLength is implicitly frames.length.

The old code resulted in framesLength being "frames.length - 1" though. I'm assuming that is unintended.

src/background/adblocker/ancestors.js
```diff
@@ -237,138 +193,98 @@ export class FramesHierarchy {
...
         console.error(
-          `Failed to query tabs too many times! The whole tabs will not be synchronised but we will continue to build the hierarchy with cosmetic filter handlers.`,
-          tabsOrError,
+          'Failed to query tabs. The whole tabs will not be synchronised, but',
+          'we will continue to build the hierarchy with cosmetic filter handlers.',
+          e,
         );
-
-        break;
+        return;
```
philipp-classen, line 272: Here I assume that breaking the loop was unintended, because "tabsOrError" would be of type Error in that case (and would be used for "tabsOrError.map", which will fail).

src/background/adblocker/ancestors.js
```diff
@@ -47,174 +39,137 @@ export class FramesHierarchy {
...
-          frames[frameIndex].details = details;
 
-          break;
-        }
+        // Update frame details
+        targetFrame.parent = parentFrameId;
+        targetFrame.documentId = documentId;
+        targetFrame.details = details;
+        break;
+      } else if (frames[frameIndex].documentId === documentId) {
```
seia-soto: This might not be fail-safe if `documentId` is not given at all. I'm not sure if we can just fully drop requests without `documentId` tho.

seia-soto: This is based on the tests.

src/background/adblocker/ancestors.js
seia-soto: It's a micro-optimisation which we wouldn't need.

src/background/adblocker/ancestors.js
seia-soto: Good catch, thanks.

seia-soto (approved):
Thanks for the simplifications. I fixed the failing test (it's not even complicated change, just one line) and I'm merging this.

PR #3031, opened 2026-01-31
Fixed matching in findParentDomain

philipp-classen:
The existing logic checks matches hostnames against domains that are a postfix. That can lead to incorrect matches.

Example:
* Visit wein.de (verify: not paused)
* Visit ein.de and pause
* Visit wein.de again (verify: now it is incorrectly paused)
* Unpause wein.de and visit ein.de (verify: now ein.de has been incorrectly unpaused)

Also, improves the complexity to O(n) number of steps by using one linear scan instead of sorting first.

### DIFF
--- a/src/pages/panel/views/main.js
+++ b/src/pages/panel/views/main.js
@@ -17,7 +17,6 @@ import Options, {
   isGloballyPaused,
   revokeGlobalPause,
   getPausedDetails,
-  findParentDomain,
   MODE_DEFAULT,
   MODE_ZAP,
 } from '/store/options.js';
@@ -41,6 +40,7 @@ import ReportConfirm from './report-confirm.js';
 import TrackerDetails from './tracker-details.js';
 import TrackersReport from './trackers-report.js';
 import WhoTracksMe from './whotracksme.js';
+import { findParentDomain } from '/utils/domains.js';
 import { ZAP_AUTORELOAD_DISABLED_HOSTNAMES } from '/utils/urls.js';
 
 const PANEL_URL = chrome.runtime.getURL('/pages/panel/index.html');
--- a/src/store/options.js
+++ b/src/store/options.js
@@ -13,6 +13,7 @@ import { store } from 'hybrids';
 
 import { DEFAULT_REGIONS } from '/utils/regions.js';
 import { isOpera, isSafari } from '/utils/browser-info.js';
+import { findParentDomain } from '/utils/domains.js';
 
 import CustomFilters from './custom-filters.js';
 import ManagedConfig, { TRUSTED_DOMAINS_NONE_ID } from './managed-config.js';
@@ -282,31 +283,6 @@ async function manage(options) {
   }
 
   return options;
-}
-
-/*
- * Finds the best matching parent domain for a given hostname from a record
- * of domains. The best match is determined as the shortest domain that is
- * either identical to the hostname or is a parent domain of the hostname.
- *
- * Returns null if no match is found or if no hostname is provided.
- */
-export function findParentDomain(record, hostname = '') {
-  if (!hostname) return null;
-
-  let bestMatch = null;
-  let lengthToBeat = hostname.length + 1;
-
-  for (const domain of Object.keys(record)) {
-    if (domain.length < lengthToBeat && hostname.endsWith(domain)) {
-      const startPos = hostname.length - domain.length;
-      if (startPos == 0 || hostname[startPos - 1] === '.') {
-        bestMatch = domain;
-        lengthToBeat = domain.length;
-      }
-    }
-  }
-  return bestMatch;
 }
 
 export function isGloballyPaused(options) {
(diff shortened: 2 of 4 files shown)


tests/unit/setup/init-chrome.js
```diff
@@ -0,0 +1,21 @@
+if (typeof globalThis.chrome === 'undefined') {
```
seia-soto, line 1: I think chrome APIs is already mocked by other tests as well. Would it be better to extend from that instead, even if we're going to keep the current approach?

philipp-classen: The reason for the extra file is that it allows to run before everything else. Mocking inside a test is too late, because the imports already crashed.

philipp-classen:
> We can avoid the additional unit tests setup by moving all "unit testable" code into separate files in `src/utils`. This is what I did in #2988.
> 
> We already have `src/utils/request.js` where you can put the `findParentDomain()` method. Utils mostly does not use any other reference from the source code.

But is this an improvement? So, it means that we can never directly test that part of the extension, right? That we always have to move code around for testing purposes?

smalluban:
Unit tests by design should test a part of the code in isolation, like a single function. Some of our files include side effects (like most of the `src/background` files - also models, like `options.js` - they also rely on extension APIs). For testing those parts, we have e2e tests. 

If you want to create unit tests for a specific function, we should move that to the `src/utils` (as it becomes a utility function), so the source file does not import other parts of the extension.

philipp-classen:
> Unit tests by design should test a part of the code in isolation, like a single function. Some of our files include side effects (like most of the `src/background` files - also models, like `options.js` - they also rely on extension APIs). For testing those parts, we have e2e tests.
> 
> If you want to create unit tests for a specific function, we should move that to the `src/utils` (as it becomes a utility function), so the source file does not import other parts of the extension.

But just to rephrase: you are saying that this part of the code cannot be unit tested. And that we currently have to organize our code around that assumption, right?

To challenge it, this PR brings a proof-of-concept that we can write unit tests independent of the part where the code lives. For me, the question is rather if the approach has down-sides that will make it harder to maintain?

It does two things:
* It provides a mechanism to setup the environment (Node.js) to not crash when loading files with browser APIs
* It extends the file loading to handles absolute paths

smalluban:
Unit tests should load as less as possible. In theory only functions that are tested. I expect that your setup works today, but some change in the source code can suddenly break those - and it can be not related change (as it would be a chain of imports, etc). If the file would have side effect based on not mocked API, it will fail.

We have two ways of testing. The e2e tests are for loading whole source code. However, our setup is complex, like absolute paths, side-effects, using extension APIs, etc. 

Your setup also creates logic in `package.json` (in the command definition). It is not clear what happens. 

I still suggest to move tested function to file, which then is imported by the extension parts.

philipp-classen:
> Unit tests should load as less as possible. In theory only functions that are tested. I expect that your setup works today, but some change in the source code can suddenly break those - and it can be not related change (as it would be a chain of imports, etc). If the file would have side effect based on not mocked API, it will fail.
> 
> We have two ways of testing. The e2e tests are for loading whole source code. However, our setup is complex, like absolute paths, side-effects, using extension APIs, etc.
> 
> Your setup also creates logic in `package.json` (in the command definition). It is not clear what happens.
> 
> I still suggest to move tested function to file, which then is imported by the extension parts.

Only that we are using the same definitions, because I feel that where we have different understandings: there are unit tests, and there are is setup to run them (the integration in the build and CI system).

In this PR, the unit test is the file [tests/unit/options.test.js](https://github.com/ghostery/ghostery-extension/pull/3031/changes#diff-a7e5cb027eff62c060f85eccdc2e28cef6907b1d6dbbcb310dc7c27b0990d200). It has no outside logic, is isolated, and the same approach is adaptable to any other file that would need to be tested.

The other part is the integration in the build system. Our current setup is limiting, since it does not support unit tests for code in all places, the PR extends the setup to fill that gap. With it, it becomes straightforward to write a test.

So, to rephrase, the real concern is that the build system becomes overly complex, right?

On the other hand, it also simplifies since you can write unit tests without having extra knowledge (e.g. knowing that part of our code base cannot be unit tested, and that code needs to be moved to other parts). Note that this is also a complexity. By extending the build system, this aspect is gone.

tests/unit/setup/init-chrome.js
```diff
@@ -0,0 +1,21 @@
...
+  global.chrome = {
+    runtime: {
+      getManifest() {
+        return {};
+      },
+      onMessage: {
+        addListener() {},
+      },
+    },
+    storage: {
```
smalluban, line 11: If `/src/store/options.js` or its dependency will use `chrome.storage.local.get()` as a side effect, your unit test, which is unrelated, will break (this is an example, it can be any other extension API, which is not mocked by your setup).

Your unit tests are fine. I want to move the function in the source code to avoid loading files unrelated to the test.

PR #3035, opened 2026-02-03
fix(paused): unify DNR rules for paused and zapped mode

### LINKED ISSUE #3034: Unify paused logic for DNR
The paused DRN rules introduced for Zapped are different from the rules for the default mode. We should unify them (I expect move to the new version from zap mode).

smalluban:
Fixes #3034. 

Tested on Safari 26.2 (macOS, iPadOS).


smalluban:
It is not required; I tested on iPad with 18.7.2, and it works as before. We can bump the version, but it is not related to this PR.

chrmod (approved):

PR #3051, opened 2026-02-06
fix(adblocker): clean up `FIREFOX_CONTENT_SCRIPT_SCRIPTLETS` flag

smalluban:
The flag is set to 100% for 8 months (https://github.com/ghostery/broken-page-reports/commit/8068f0462b54e250958f5eecfa24016764851113). It also complicates the introduced logic (https://github.com/ghostery/ghostery-extension/pull/2961), which relies on the feature.

### DIFF
--- a/src/background/adblocker/ancestors.js
+++ b/src/background/adblocker/ancestors.js
@@ -249,4 +249,14 @@ export class FramesHierarchy {
       (await chrome.tabs.query({})).map((tab) => this.#handleTab(tab)),
     );
   }
+
+  handleWebextensionEvents() {
+    chrome.tabs.onRemoved.addListener((tabId) => {
+      this.unregister(tabId, 0);
+    });
+
+    chrome.tabs.onReplaced.addListener((addedTabId, removedTabId) => {
+      this.replace(addedTabId, removedTabId);
+    });
+  }
 }
--- a/src/background/adblocker/index.js
+++ b/src/background/adblocker/index.js
@@ -81,13 +81,6 @@ const contentScripts = (() => {
   };
 })();
 
-let SUBFRAME_SCRIPTING = { enabled: false };
-
-SUBFRAME_SCRIPTING = resolveFlag(FLAG_SUBFRAME_SCRIPTING);
-
-const hierarchy = new FramesHierarchy();
-hierarchy.handleWebWorkerStart();
-
 function getEnabledEngines(config) {
   if (config.terms) {
     const list = ENGINES.filter(({ key }) => config[key]).map(
@@ -334,6 +327,16 @@ function injectStyles(styles, details) {
     .catch((e) => console.warn('[adblocker] failed to inject CSS', e));
 }
 
+const SUBFRAME_SCRIPTING = resolveFlag(FLAG_SUBFRAME_SCRIPTING);
+
+let framesHierarchy;
+if (__PLATFORM__ !== 'firefox') {
+  framesHierarchy = new FramesHierarchy();
+
+  framesHierarchy.handleWebWorkerStart();
+  framesHierarchy.handleWebextensionEvents();
+}
+
 async function injectCosmetics(details, config) {
   const { bootstrap: isBootstrap = false, scriptletsOnly } = config;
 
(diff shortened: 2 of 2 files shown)


src/background/adblocker/index.js
```diff
@@ -383,7 +369,7 @@ async function injectCosmetics(details, config) {
 
   let ancestors = undefined;
   if (SUBFRAME_SCRIPTING.enabled && typeof parentFrameId === 'number') {
-    if (FIREFOX_CONTENT_SCRIPT_SCRIPTLETS.enabled) {
```
smalluban, line 386: @seia-soto Flags are only temporary and should not be used as a platform check - you should always use `__PLATFORM__` variable to run code only on Firefox (in this case, the condition should contain both).
The code on `main` would run Firefox-related code on Chromium if we removed the platform check from the flag (https://github.com/ghostery/broken-page-reports/pull/2124)

tests/e2e/spec/adblocker.spec.js
```diff
@@ -197,7 +197,7 @@ describe('Adblocker Capabilities', function () {
       for (const [id, filter] of networkingFilters) {
         it(filter, async function () {
           // The networking test only have one timing candidate
-          expect(reports.networking[0].results[id]).toBe(true);
+          await expect(reports.networking[0].results[id]).toBe(true);
```
chrmod, line 200: why we need those? which part is async?

smalluban: `expect` is async by default. This is an unrelated bug, but I added those for the sake of correctness

FYI: https://webdriver.io/docs/api/expect-webdriverio/

src/background/adblocker/ancestors.js
```diff
@@ -249,17 +249,4 @@ export class FramesHierarchy {
...
-  handleWebextensionEvents(FIREFOX_CONTENT_SCRIPT_SCRIPTLETS) {
-    chrome.tabs.onRemoved.addListener((tabId) => {
-      if (FIREFOX_CONTENT_SCRIPT_SCRIPTLETS.enabled === false) {
-        this.unregister(tabId, 0);
-      }
-    });
-    chrome.tabs.onReplaced.addListener((addedTabId, removedTabId) => {
-      if (FIREFOX_CONTENT_SCRIPT_SCRIPTLETS.enabled === false) {
-        this.replace(addedTabId, removedTabId);
-      }
```
seia-soto, line 262: I believe we should preserve these event listeners. I don't see the additions.

tests/e2e/spec/adblocker.spec.js
```diff
@@ -172,7 +172,7 @@ describe('Adblocker Capabilities', function () {
     describe('Styling', function () {
       for (const [id, filter] of stylingFilters) {
         it(filter, async function () {
-          expect(
+          await expect(
```
seia-soto, line 175: Are these `await`s necessarily required? I think it might be better to use `function` instead of `async function` here.

seia-soto: I saw https://github.com/ghostery/ghostery-extension/pull/3051/changes#r2781866615, resolving this.

seia-soto (commented):
I see event listeners dropped but not added again.

src/background/adblocker/ancestors.js
philipp-classen: I think you are right. I also missed that. If I understand correctly, we should have the "tabs.onRemoved" and "tabs.onReplaced" for all platforms, right?

Something like that if I'm not mistaken:
```
  handleWebextensionEvents() {
    chrome.tabs.onRemoved.addListener((tabId) => {
      this.unregister(tabId, 0);
    });
    chrome.tabs.onReplaced.addListener((addedTabId, removedTabId) => {
      this.replace(addedTabId, removedTabId);
    });
  }
```

And re-add the call to `handleWebextensionEvents`.

```
hierarchy.handleWebextensionEvents();
```

tests/e2e/spec/adblocker.spec.js
smalluban: Yes, expect returns a promise.

src/background/adblocker/ancestors.js
smalluban: If the `FIREFOX_CONTENT_SCRIPT_SCRIPTLETS` is constantly true, there are empty functions...

smalluban: This must be another example of wrong usage of the flag - Do you mean that `this.unregister()` and `this.replace()` should run on Chromium?

seia-soto: @smalluban These are not for Firefox. Yes, they should run on Chromium. I recommend to add a Firefox check there instead.

PR #3059, opened 2026-02-11
fix(download): improve files download and backup export on Safari

### LINKED ISSUE #3058: iOS: csv stemming from tracker report needs name
Image

smalluban:
Fixes #3058

* Centralized File Downloads: Replaced the file-saver dependency with a new internal utility files.js to handle file downloads across the extension, standardizing logic for reports and settings backups.
* Platform-Specific Improvements: The new utility centralizes workarounds for iOS/iPadOS (converting Blobs to Data URLs) and WebKit browsers, ensuring consistent download behavior across different platforms.
* Safari Download Workaround: Introduced a dedicated "Download Helper" page (download) and updated the build script to include it for non-Firefox targets, solving issues where downloads wouldn't trigger correctly from popups in Safari.

### DIFF
--- a/src/pages/logger/views/main.js
+++ b/src/pages/logger/views/main.js
@@ -48,7 +48,11 @@ async function downloadReport(host) {
   ]);
 
   await download({
-    data: stringify(report, { header: true, columns: REPORT_COLUMNS, escape_formulas: true }),
+    data: stringify(report, {
+      header: true,
+      columns: REPORT_COLUMNS,
+      escape_formulas: true,
+    }),
     filename: `ghostery-logger-report.csv`,
     type: 'text/csv;charset=utf-8;',
   });
--- a/src/pages/panel/views/trackers-report.js
+++ b/src/pages/panel/views/trackers-report.js
@@ -42,7 +42,11 @@ async function downloadReport(host, event) {
   }
 
   await download({
-    data: stringify(report, { header: true, columns: REPORT_COLUMNS }),
+    data: stringify(report, {
+      header: true,
+      columns: REPORT_COLUMNS,
+      escape_formulas: true,
+    }),
     filename: `ghostery-website-report-${host.stats.hostname}.csv`,
     type: 'text/csv;charset=utf-8;',
     // Safari does not support downloading files from the popup window,


smalluban:
@AdamGhst Please test out the changes. We currently have 3 places where you can download a file:
1. Settings backup (Settings page -> My Ghostery)
2. Trackers report (panel -> click on number of blocked trackers)
3. Logger (download action on the toolbar) - skip Safari here, as it is "supported", but currently it does not fetch data correctly.

On iOS, we cannot download a file directly - for backup, it is just "displayed", for a report, it shows a prompt to download the "Unknown.csv" file. You should focus on desktop browsers on Win/Mac to be sure that this PR does not break anything.

src/pages/logger/views/main.js
```diff
@@ -46,8 +47,11 @@ function downloadReport(host) {
     log.organization ?? '',
   ]);
 
-  const data = stringify(report, { header: true, columns: REPORT_COLUMNS });
-  saveAs(new Blob([data], { type: 'text/csv' }), 'Ghostery Logger Report.csv');
+  await download({
+    data: stringify(report, { header: true, columns: REPORT_COLUMNS }),
+    filename: `ghostery-logger-report-${new Date().toISOString()}.csv`,
```
philipp-classen: This will lead to filenames including a `:`.

Might be an issue on Windows (not sure). Allowed, but uncommon in Linux.

I would recommend to replace ":" by something safe like "-".

smalluban: I must have made a suggested change from the editor when I was updating names. I don't mind removing the date, as if it is not in proper format, it makes no sense to put it. The files have their own creation and modification dates (on the filesystem).

src/pages/logger/views/main.js
```diff
@@ -46,8 +47,11 @@ function downloadReport(host) {
     log.organization ?? '',
   ]);
 
-  const data = stringify(report, { header: true, columns: REPORT_COLUMNS });
-  saveAs(new Blob([data], { type: 'text/csv' }), 'Ghostery Logger Report.csv');
+  await download({
+    data: stringify(report, { header: true, columns: REPORT_COLUMNS }),
```
philipp-classen: Just noticed that cvs-stringify requires to opt-in to sanitization. I would recommend to do that always by default:

```suggestion
    data: stringify(report, { header: true, columns: REPORT_COLUMNS, escape_formulas: true }),
```

smalluban: Good catch!

PR #3060, opened 2026-02-12
Handle request reporter initialization errors gracefully

philipp-classen:
It should not happen that the request reporter fails during construction. However, given that it is an optional library, failing gracefully and disabling it without crashing looks safer.

If it crashes, it also prevents the url reporter from starting up. By disabling it, this cascading effort is mitigated.

### DIFF
--- a/src/background/reporting/webrequest-reporter.js
+++ b/src/background/reporting/webrequest-reporter.js
@@ -38,41 +38,54 @@ if (chrome.webRequest) {
     remoteConfig = remote;
   });
 
-  webRequestReporter = new RequestReporter(config.request, {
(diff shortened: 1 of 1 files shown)


philipp-classen:
> Is it not the case that `webRequestReporter` fails, but only when it initializes?
> 
> https://github.com/ghostery/ghostery-extension/blob/main/src/background/reporting/index.js#L42

No, it failed already in the constructor (even though this should normally not happen), and that prevents the construction of the urlReporter component.

The handling of "init" is correct as is. It does not delay the initialization of urlReport, nor does it prevent it from starting.

IMHO, we should normally not have to work around for errors in the construction, but here, I think it still makes sense, as the component is optional and disabling it is a possibility. But, of course, in general the construction should be never risk failing.

Issue #2779, opened 2025-10-06
Custom filters: on MV3 generichides are not passed to adblocker engine


chrmod:
Custom filter like `@@github.io$generichide,badfilter` are network filters, but they still should be added to adblocker engine as they influence cosmetic filters matching

seia-soto:
The question would be even if we pass to adblocker engine, will it effectively apply ghide and shide for the DNR API? I believe we will require more measure than just passing values to adblocker engine.

seia-soto:
We're now shipping all network filters for logger and hide exceptions.

seia-soto:
We still don't pass those filters to the engine.

PR #3081, opened 2026-02-17
fix: preserve $elemhide, $ghide, and $shide

seia-soto:
fixes https://github.com/ghostery/ghostery-extension/issues/2779


seia-soto:
We need to pass all network filters to the engine for the logger actually.

seia-soto:
closing in favour of https://github.com/ghostery/ghostery-extension/pull/3085

PR #3085, opened 2026-02-18
fix(custom-filters): pass all network filters regardless of platform

### LINKED ISSUE #2779: Custom filters: on MV3 generichides are not passed to adblocker engine

seia-soto:
This fixes custom filters not being reflected to the logger interface and the corrected behaviour with $elemhide, $generichide, $specifichide exceptions.

fixes #2779 fixes https://github.com/ghostery/ghostery-extension/issues/3086

### DIFF
--- a/tests/e2e/spec/adblocker.spec.js
+++ b/tests/e2e/spec/adblocker.spec.js
@@ -32,6 +32,11 @@ async function enableAllPrivacyToggle() {
   await setPrivacyToggle('anti-tracking', true);
 }
 
+async function disableCustomFilters() {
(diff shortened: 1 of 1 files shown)


smalluban:
Can we add an e2e test for it? Or is it hard to create such a case?

smalluban:
@seia-soto ping

seia-soto:
@smalluban this is now ready.

PR #3087, opened 2026-02-18
fix: replace `__PLATFORM__` environment variable

smalluban:
This PR replaces `__PLATFORM__` environment variable with `__FIREFOX__` and `__CHROMIUM__`. The readability of conditions has been improved.

This PR does not remove any condition. It only replaces variables. We can check whether there are places where using env for conditional code makes sense, but it's usually used when code must run only on a selected platform.

### DIFF
--- a/src/background/adblocker/index.js
+++ b/src/background/adblocker/index.js
@@ -108,7 +108,7 @@ function pause(ms) {
 
 export async function reloadMainEngine() {
   // Delay the reload to avoid UI freezes in Firefox and Safari
-  if (__PLATFORM__ === 'firefox' || isWebkit()) await pause(1000);
+  if (__FIREFOX__ || isWebkit()) await pause(1000);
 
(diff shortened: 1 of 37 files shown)


src/background/adblocker/index.js
```diff
@@ -216,7 +216,7 @@ const INJECTION_TARGET_DOCUMENT_ID = resolveFlag(FLAG_INJECTION_TARGET_DOCUMENT_
 function resolveInjectionTarget(details) {
   const target = { tabId: details.tabId };
 
-  if (__PLATFORM__ !== 'firefox' && INJECTION_TARGET_DOCUMENT_ID.enabled && details.documentId) {
+  if (__CHROMIUM__ && INJECTION_TARGET_DOCUMENT_ID.enabled && details.documentId) {
```
smalluban, line 219: Potential candidate for removing the condition. I looked through the changes, and I haven't found much more.

PR #3095, opened 2026-02-20
fix(adblocker): update filters when engines change & add missing documentId

smalluban:
This PR fixes two minor issues:
* Adds a trigger for updating filters when enabled engines change. It is required, as we only update enabled engines. If the user disables one of them, and then re-enable, those might be out of date until next scheduled update.
* Adds missing `documentId` option calculated in `onMessage` callback from the content scripts

### DIFF
--- a/src/background/adblocker/index.js
+++ b/src/background/adblocker/index.js
@@ -186,7 +186,7 @@ export const setup = asyncSetup('adblocker', [
     const lastEnabledEngines = lastValue && getEnabledEngines(lastValue);
 
     // Enabled engines changed (they might contain outdated filters)
-    const engingesChanged =
+    const enginesChanged =
       lastEnabledEngines &&
       (enabledEngines.length !== lastEnabledEngines.length ||
         enabledEngines.some((id, i) => id !== lastEnabledEngines[i]));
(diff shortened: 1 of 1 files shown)


src/background/adblocker/index.js
```diff
@@ -185,19 +185,23 @@ export const setup = asyncSetup('adblocker', [
...
+    if (!(await engines.init(engines.MAIN_ENGINE)) || engingesChanged) {
       await reloadMainEngine();
     }
 
-    // Update engines if filters are outdated (older than 1 hour)
-    if (options.filtersUpdatedAt < Date.now() - UPDATE_ENGINES_DELAY) {
+    // Update engine filters:
+    // * when engines changed, so there might be re-enabled engines with outdated filters
+    // * when filters are outdated (older than 1 hour)
+    if (engingesChanged || options.filtersUpdatedAt < Date.now() - UPDATE_ENGINES_DELAY) {
```
smalluban: As requests to the CDN are cached at the HTTP level, even with toggling the feature in sequence, the update process will not fetch data and skip the update (if it was already done within the timeframe).

src/background/adblocker/index.js
```diff
@@ -185,19 +185,23 @@ export const setup = asyncSetup('adblocker', [
...
-    if (
-      // Reload/mismatched main engine
-      !(await engines.init(engines.MAIN_ENGINE)) ||
-      // Enabled engines changed
-      (lastEnabledEngines &&
-        (enabledEngines.length !== lastEnabledEngines.length ||
-          enabledEngines.some((id, i) => id !== lastEnabledEngines[i])))
-    ) {
+    // Enabled engines changed (they might contain outdated filters)
+    const engingesChanged =
```
philipp-classen: typo in variable name

PR #3096, opened 2026-02-20
fix(utils): add fallback to request url for source url

smalluban:
On Firefox, requests to images on YT throw an error:
Zrzut ekranu 2026-02-20 o 09 44 48

The error is related to how we compute the `sourceHostname`. Take a look at the request details:

Zrzut ekranu 2026-02-20 o 09 44 37

It must be an XHR unconnected to the main frame, as the `originUrl` and `documentUrl` are `undefined`. As the last resort, we must fallback to `details.url`.

The PR fixes the error message (and this request processing) in the SW logs.

### DIFF
--- a/src/background/adblocker/index.js
+++ b/src/background/adblocker/index.js
@@ -528,7 +528,7 @@ function isTrusted(request, type) {
 }
 
 if (__FIREFOX__) {
-  function resolveRequest(details) {
+  function getMatchableRequest(details) {
     // Extension context request
     if (
       (details.tabId === -1 && details.url.startsWith('moz-extension://')) ||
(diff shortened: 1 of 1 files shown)


philipp-classen:
The example request is interesting. But it looks like that the URL in this example is a third-party (or comes from a service worker). The question is if it is correct to check if the third-party domain is paused?

src/utils/request.js
```diff
@@ -43,7 +43,7 @@ function resolveSourceURL(details) {
     return initiator;
   }
 
-  return details.originUrl || details.documentUrl || '';
+  return details.originUrl || details.documentUrl || details.url;
```
chrmod: we cannot do that as this breaks "partiness" - a 3rd party request may become a first party.

smalluban:
@chrmod I moved the fix to event listeners. I've found that logic, which allows requests to be processed, is not the same for `onBeforeRequest` and `onHeadersReceived`. I've created the `resolveRequest()` function to unify the logic and avoid differences in the future.

PR #3097, opened 2026-02-23
feat(adblocker): DNR preprocessors

### LINKED ISSUE #3057: Preprocessors for DNR
We want to deliver preprocessors for DNR rules to align the behaviour on multiple platforms more precisely. Multiple approaches can be made involving both backend and extension.

The last consensus is below:

1. We will create a separate file per each DNR ruleset to ship additional metadata along with the DNR rules. This is to completely be free with the DNR specification; some platforms might restrict us from using additional fields. The file is in the following JSON format: `Record<number & { __: 'filter id' }, { preprocessor: string & { __: 'condition string' } }>`.
2. Those files will be loaded using glob import in the extension side. (see https://vite.dev/guide/features#glob-import)
3. We will use those metadata to disable ineligible rules depending on the platform and environment. The string evaluation will be done using adblocker library (see the details in bottom of this issue). We will do this every time we enable static rulesets.

---

Regardless of the specific details, we have some potential blockers listed below for the future reference:

1. `IFilter.prototype.toString` method cannot be perfect

Adblocker library mutates the filter representation internally to speed up matching and optimise the structure. This is currently seen as the limitation of the implementation. However, we can potentially set the `rawFilter` property of `NetworkFilter` to let `toString` work even without the `debug` flag of the engine.

This has been investigated by https://github.com/ghostery/adblocker/issues/3694.

2. Evaluation of preprocessor condition

We do have a function to evaluate the preprocessor condition in the adblocker library, but it's not simply exposed to the module entry point. If we want to evaluate the condition string outside of the library, we need to expose the function in the index file.

refs https://github.com/ghostery/adblocker/blob/master/packages/adblocker/src/preprocessor.ts#L103

seia-soto:
fixes #3057

### DIFF
--- a/tests/e2e/spec/custom-filters.spec.js
+++ b/tests/e2e/spec/custom-filters.spec.js
@@ -108,7 +108,9 @@ describe('Custom Filters', function () {
 
     if (browser.isChromium) {
       await expect(getExtensionElement(`icon:tracker:www.example.net:blocked`)).toBeDisplayed();
+      await expect(getExtensionElement(`icon:tracker:www.example.org:blocked`)).not.toBeDisplayed();
     } else if (browser.isFirefox) {
+      await expect(getExtensionElement(`icon:tracker:www.example.net:blocked`)).not.toBeDisplayed();
       await expect(getExtensionElement(`icon:tracker:www.example.org:blocked`)).toBeDisplayed();
     }
   });


src/background/dnr.js
```diff
@@ -10,18 +10,86 @@
...
+import { getBrowser, isMobile } from '/utils/browser-info.js';
 
 import { UPDATE_ENGINES_DELAY } from './adblocker/index.js';
 import { updateRedirectProtectionRules } from './redirect-protection.js';
 
 if (__CHROMIUM__) {
+  const DNR_METADATA = (function () {
+    // We need to depend on `eager` option since dynamic imports
+    // are not allowed in web workers scope.
+    const modules = import.meta.glob('/rule_resources/*.metadata.json', { eager: true });
```
seia-soto: This generates multiple static `import` statements in the resulting code then maps them into an object.

```js
import filenameIntoUniqueName from '../relpath';
import filenameIntoUniqueName2 from '../relpath';
import filenameIntoUniqueName3 from '../relpath';

const modules = {'./filepath': filenameIntoUniqueName, ...};
```

seia-soto: refs https://vite.dev/guide/features#glob-import

chrmod: wouldn't it be better to fetch only the ones you need with `fetch`? 

or alternatively, build time, you can generate a .js file with all of the metadata merged. So its just a static import. If the file is small that's probably the fastest for the runtime

src/background/dnr.js
```diff
@@ -10,18 +10,86 @@
...
+
+    const env = new Map();
+    env.set('ext_ghostery', true);
+    env.set('ext_ublock', true);
+    env.set('ext_ubol', true);
+    env.set('env_ghostery', true);
+    env.set('env_chromium', browser.name !== 'safari');
+    env.set('env_edge', browser.name === 'edge');
+    env.set('env_firefox', false);
+    env.set('env_mobile', isMobile());
```
chrmod: can this change dynamically for iPad?

src/background/dnr.js
```diff
@@ -10,18 +10,86 @@
...
 import { updateRedirectProtectionRules } from './redirect-protection.js';
 
 if (__CHROMIUM__) {
+  const DNR_METADATA = (function () {
+    // We need to depend on `eager` option since dynamic imports
+    // are not allowed in web workers scope.
+    const modules = import.meta.glob('/rule_resources/*.metadata.json', { eager: true });
+    const browser = getBrowser();
+
+    const env = new Map();
```
chrmod: shouldn't we use https://github.com/ghostery/ghostery-extension/blob/4f38ae428fb1357eefd7ed9094d058f03fda5862/src/utils/engines.js#L38 instead?

chrmod (changes requested):
what this should react to `env_experimental`?

src/background/dnr.js
smalluban: It will be a JS file in the dist.

Using `import.meta.glob`, we move those files into the dist and generate the most efficient version (vite decides) to import JSON data. It is cached. 
Those files are almost empty.
Also, the module import is synchronous (the code awaits it), so we can use those as already fetched.

src/background/dnr.js
```diff
@@ -10,18 +10,60 @@
...
 if (__CHROMIUM__) {
+  const DNR_METADATA = (function () {
+    // We need to depend on `eager` option since dynamic imports
+    // are not allowed in web workers scope.
+    const modules = import.meta.glob('/rule_resources/*.metadata.json', { eager: true });
+
+    /**
+     * @returns {Record<string, { preprocessor: string; }>}
+     */
+    function getMetadata(rulesetId) {
```
smalluban: Move those helper functions from here to the `utils/dnr.js`

src/background/dnr.js
```diff
@@ -10,18 +10,60 @@
...
+        return;
+      }
+      return metadata.default;
+    }
+
+    /**
+     * @param {string} rulesetId
+     * @returns {number[]}
+     */
+    function getDisabledRuleIds(rulesetId) {
```
smalluban: `getEnvDisabledRuleIds` ? We need some context in the function name.

src/background/dnr.js
```diff
@@ -10,18 +10,60 @@
...
+      const env = getEnv();
+      const metadata = getMetadata(rulesetId);
+      if (typeof metadata === 'undefined') {
+        return [];
+      }
+      const disabledRuleIds = [];
+      for (const [ruleId, constraints] of Object.entries(metadata)) {
+        if (!evaluatePreprocessor(constraints.preprocessor, env)) {
```
smalluban: This should be a helper function of `utils/engines.js` - it is not scrict call to the imported function. It uses `getEnv()` (which I am not sure is needed - if the `evaluatePreprocessor` does not change `env`, it can be skipped. We control this function...)

src/background/dnr.js
seia-soto: We're now reusing env from engine utils.

src/utils/dnr.js
```diff
@@ -85,3 +87,38 @@ export function getRedirectProtectionRules(rules) {
 
   return result;
 }
+
+export const getExcludedRuleIdsByPreprocessors = (function () {
+  // We need to depend on `eager` option since dynamic imports
+  // are not allowed in web workers scope.
+  const modules = import.meta.glob('/rule_resources/*.metadata.json', { eager: true });
```
smalluban: Why can this not be just in the main module body? If we can, we should avoid dynamic function generation. Those helper functions can be defined at the top level.

src/utils/engines.js
```diff
@@ -452,4 +461,8 @@ export function remove(name) {
   });
 }
 
+export function evaluatePreprocessorCondition(condition) {
+  return evaluatePreprocessor(condition, ENV);
```
smalluban: This `getEnv` is not required by anything else, move `structuredClone()` call into this function and remove `getEnv()` function.

src/utils/dnr.js
seia-soto: We're not calling the function dynamically. That part will be fully replaced by Vite on build time into an object. See "eager" part in  https://vite.dev/guide/features#glob-import for more information.

src/utils/engines.js
seia-soto: `getEnv` is dropped but I don't think we need `structuredClone` as we don't expect the env modification elsewhere than via the engine.

src/utils/env.js
```diff
@@ -0,0 +1,24 @@
+import { evaluatePreprocessor } from '@ghostery/adblocker';
+
+export const ENV = new Map([
```
seia-soto: @smalluban The `ENV` definition is moved to separate file to prevent call to hybridjs. It's localization feature automatically queries `navigator.language` and it ends up unhandled error from Node.JS. I wonder if you just prefer to patch the `navigator.language`.

```
file:///Users/aa/local/repositories/ghostery-extension/node_modules/hybrids/src/localize.js:22
    const codeWithoutRegion = code.split("-")[0];
                                   ^

TypeError: Cannot read properties of undefined (reading 'split')
    at file:///Users/aa/local/repositories/ghostery-extension/node_modules/hybrids/src/localize.js:22:36
```

We really don't reference the hybrid related functions from unit test but `utils/engines.js` imported by `utils/dnr.js` has an import to `hybrid`.

src/utils/dnr.js
```diff
@@ -85,3 +87,46 @@ export function getRedirectProtectionRules(rules) {
...
 }
+
+export const getExcludedRuleIdsByPreprocessors = (function () {
+  // This simple check is to support unit test without Vite-only
+  // features.
+  if (typeof import.meta.glob === 'undefined') {
+    return function () {
+      return [];
+    };
+  }
```
seia-soto: @smalluban Here, we probably don't want to compile test with Vite before running it. However, this check runs on initialisation step only once and it's very cheap.

src/utils/dnr.js
smalluban: Now it is on the module body as I asked for. It is fine now.

src/utils/dnr.js
smalluban: Look at the current unit test setup (files running before tests). It should be possible to mock `import.meta.glob` from there. The source code must be clean.

src/utils/env.js
smalluban: Importing `hybrids` in nodejs environment is supported. If the `utils/engines.js` does not have any side-effects which break the tests, this should not be moved into a separate file. The `ENV` is strictly related to `utils/engines.js`.

src/utils/dnr.js
smalluban: No, actually it is not. I am not sure how I looked at it. Please define `getExcludedRuleIdsByPreprocessors` as ordinary function on the module level, and modules should be defined on the module top level.

src/utils/engines.js
smalluban: I still see `getEnv()` method above.

src/utils/env.js
seia-soto: Loading hybrid on Node actually breaks the things. I'm not sure things are working internally but I believe that it could be because of some polyfill injected from the test setup.

seia-soto: You can check https://github.com/ghostery/ghostery-extension/actions/runs/22387400828/job/64801026352

src/utils/dnr.js
seia-soto: `import.meta` is defined per module. Therefore, we cannot avoid making the change on the source code. I mocked it via using NodeJS preload hook.

src/utils/engines.js
seia-soto: It's probably outdated diff. I don't see `getEnv` keyword from the source code search.

package.json
```diff
@@ -12,7 +12,7 @@
     "lint:src": "eslint",
     "lint:build": "npm run build firefox -- --silent && addons-linter dist",
     "test": "npm run test:unit && npm run wdio && npm run wdio:update",
-    "test:unit": "node --require ./tests/unit/setup/init-chrome.js --import ./tests/unit/setup/register-import.js --test",
+    "test:unit": "node --import ./tests/unit/setup/preload.js --test",
```
seia-soto: I'm simplifying this preload calls. `--require` is for CJS module that we wouldn't prefer in this package. Note that `--require` still has a priority according to Node.JS docs.

src/utils/env.js
smalluban: You should look at the source code of hybrids. It is prepared to run on nodejs, but with mocked browser APIs. I should mention that (and thought about that). The error comes from accessing non-existing `globalThis.navigator.languages` - you should just mock this, and hybrids will work.

However, we should avoid loading UI modules as much as we can in unit tests. In this case, I see we have unit tests for `utils/dnr.js` - so we should move your new functions (which rely on `utils/engines.js` toa  separate file, not moving out the `ENV` from the engines). It might be a good choice to put some helper functions inside the `utils/engines.js` as those are related to the engine directly.

seia-soto: I think that we could be some logic from `engines.js`, especially about the idb related functions. Not sure if it's going to be a big refactor and what's going to be an ideal change.

I just mocked the browser API.

seia-soto: I tried not to separate the env from the engine but having every import in `engines.js` makes everything complicated. First, it's not only tied to the extension and test code but also strongly tied to the build step. From there, we cannot resolve the path without the custom loader for the modules.

```
aa@HoJeongs-MacBook-Pro ghostery-extension % npm run build

> ghostery-extension@10.5.31 build
> node scripts/build.js

Reading manifest.chromium.json...
node:internal/modules/esm/resolve:275
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/store/resources.js' imported from /Users/aa/local/repositories/ghostery-extension/src/utils/engines.js
    at finalizeResolution (node:internal/modules/esm/resolve:275:11)
    at moduleResolve (node:internal/modules/esm/resolve:866:10)
    at defaultResolve (node:internal/modules/esm/resolve:992:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:719:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:736:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:765:52)
    at #resolve (node:internal/modules/esm/loader:701:17)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:621:35)
    at ModuleJob.syncLink (node:internal/modules/esm/module_job:160:33)
    at ModuleJob.link (node:internal/modules/esm/module_job:245:17) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///store/resources.js'
}
```

I recommend just separting envs for now, then fully separate the utils required by both the external components and extension.

seia-soto:
@smalluban I wonder if proceeding in that direction is the simplest. Please confirm that we put loader to every module that loading the module including "hybrid", "import.meta.glob", or use of absolute path: "/...". We have so many places to patch even including build step: scripts/download-dnr-rulesets.js (requires utils/dnr.js). We can just create another utils/env.js and isolate the env or put loader to everywhere.

Putting loader and patching the source code on the way makes us to consider source-mapping in the future as well.

src/utils/env.js
smalluban: There is a unit test file `tests/unit/redirect-protection.test.js` which imports `utils/dnr.js`
> import { getRedirectProtectionRules } from '../../src/utils/dnr.js';

If you move your new methods from `utils/dnr.js` into a separate file, the `utils/engines.js` and whole chain of files won't be loaded. We do not plan to unit test `utils/engines.js` - we have e2e tests for that. We cannot avoid importing things that may break when running in an isolated NodeJS process. If so, we need to move out those parts from files, which are imported by unit tests (not the opposite).

If it is better not to put those in `utils/engines.js`, you can create `utils/preprocessors.js`. Then the whole problem with unit testing will be gone.

src/utils/preprocessor.js
```diff
@@ -0,0 +1,53 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { evaluatePreprocessor } from '@ghostery/adblocker';
+
+export const ENV = new Map([
```
smalluban: This still should be inside of `utils/engines.js`. This file's contents are no longer important in unit tests; you can import `utils/engines.js` here.

tests/unit/setup/init-chrome.js
```diff
@@ -22,6 +22,7 @@ Object.defineProperty(global, 'navigator', {
     userAgent:
       'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36',
```
smalluban, line 23: Why did you move this file into `scripts`? All test-related files should be in the `tests/` path.

Also, please move out all of the tests-related changes to a separate PR (those changes are good changes), as you don't have to fix anything with the current structure.

seia-soto: It was for the build script as well at the moment. It's now decided to use a separate file to avoid imports, so it can be extracted as a separate pull request.

src/utils/preprocessor.js
seia-soto: Importing `engines.js` means that we're going to bring `hybrids` and I don't want it as it increases the complexity. In that case, we still need to import `engines.js` from `preprocessors.js` losing the sense of having a separate file.

Still removing a requirement for mocking `import.meta.glob` is making the situation a lot better. The changes are now in the suggested direction again.

smalluban: Which unit test will load `utils/engines.js` ?

smalluban:
@seia-soto, please update the branch and set the PR as "ready for review".

src/background/dnr.js
smalluban: @seia-soto What about that?

seia-soto: I think it depends if we can detect the desktop mode on Safari. I don't think we need to target them differently from the view of the filter.

src/utils/preprocessor.js
seia-soto: Sorry for the confusion there. It looks like I confused with the old diff when writing that. The dependency tree is now clear.

src/background/dnr.js
smalluban: I doubt that UBlock is dynamically changing the "env" on iPads. The `isMobile()` returns `false` on iPads, and I would expect that. Most of the sites even force desktop view on iPads.

For me, this is fine to be resolved.

smalluban:
@seia-soto This is a good question from @chrmod
> what this should react to env_experimental?

Have you considered that?

seia-soto:
@smalluban I think it's fine since we regenerate the list of filters to be disabled on every configuration update.

src/background/dnr.js
```diff
@@ -175,6 +192,19 @@ if (__CHROMIUM__) {
       if (!enabledRulesetIds.includes(id)) {
         enableRulesetIds.push(id);
       }
+
+      try {
+        const disableRuleIds = getExcludedRuleIdsByPreprocessors(id);
+        await chrome.declarativeNetRequest.updateStaticRules({
```
smalluban: We should update disabled rules only once per version (not each time SW starts). I think it would be much better to only process rulesets, which must be enabled -> L216 - process only rulesests from  `enableRulesetIds` (this is how we avoid re-enabling rulesets for each SW restart, but we must update when extension is updated)

smalluban: I think it must work that way, but can we disable rules from not yet enabled rulesets? This is how it currently in your code.

src/background/dnr.js
```diff
@@ -175,6 +192,19 @@ if (__CHROMIUM__) {
       if (!enabledRulesetIds.includes(id)) {
         enableRulesetIds.push(id);
       }
+
+      try {
+        const disableRuleIds = getExcludedRuleIdsByPreprocessors(id);
```
smalluban: I expect the static ruleset if reloaded with extension update, will have a clean state of disabled rules. We should only run `updateStaticRules()` if the `disableRuleIds` is not empty (look at another comment of refactoring the function, but still it should run the Chrome APIs only if it is required)

scripts/download-dnr-rulesets.js
```diff
@@ -42,32 +53,25 @@ for (const [name, target] of Object.entries(RULESETS)) {
...
       .then(filterMaxPriorityRules);
 
     writeFileSync(outputPath, JSON.stringify(dnr, null, 2));
+
+    if (list.dnr.metadataUrl) {
+      const metadata = await fetch(list.dnr.metadataUrl).then(
+        generateJSON('DNR ruleset metadata', name),
+      );
+
+      writeFileSync(metadataPath, JSON.stringify(metadata, null, 2));
```
smalluban: We can optimize here, and write file only if metadata is not empty. We still have to try to load it (to be sure that error is not coming from fetching data from server), but we can skip empty files to avoid too much imports of empty files (as your `getExcludedRuleIdsByPreprocessors()` is already prepared for `undefined` as metadata).

src/utils/preprocessor.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {string} rulesetId
+ * @returns {number[]}
+ */
+export function getExcludedRuleIdsByPreprocessors(rulesetId) {
+  const metadata = DNR_METADATA[`/rule_resources/dnr-${rulesetId}.metadata.json`];
+  const disabledRuleIds = [];
+  if (typeof metadata === 'undefined') {
+    return disabledRuleIds;
+  }
+  for (const [ruleId, constraints] of Object.entries(metadata)) {
```
smalluban: `metadata` is a module, which has `default` property, which maps to your metadata... Did you test the feature?

src/utils/preprocessor.js
```diff
@@ -0,0 +1,36 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { evaluatePreprocessor } from '@ghostery/adblocker';
+
+import { ENV } from './engines.js';
+
+// We need to depend on `eager` option since dynamic imports are
+// not allowed in web workers scope.
+const DNR_METADATA = import.meta.glob('/rule_resources/*.metadata.json', { eager: true });
```
smalluban: Add `import: 'default'` to options, to avoid a nested object.

src/utils/preprocessor.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @param {string} rulesetId
+ * @returns {number[]}
+ */
+export function getExcludedRuleIdsByPreprocessors(rulesetId) {
+  const metadata = DNR_METADATA[`/rule_resources/dnr-${rulesetId}.metadata.json`];
+  const disabledRuleIds = [];
+  if (typeof metadata === 'undefined') {
```
smalluban: you can simplify it to just `if (metadata) { return Object.entries().reduce(); }; return []`

src/utils/preprocessor.js
```diff
@@ -0,0 +1,36 @@
...
+
+// We need to depend on `eager` option since dynamic imports are
+// not allowed in web workers scope.
+const DNR_METADATA = import.meta.glob('/rule_resources/*.metadata.json', { eager: true });
+
+/**
+ * @param {string} rulesetId
+ * @returns {number[]}
+ */
+export function getExcludedRuleIdsByPreprocessors(rulesetId) {
```
smalluban: This function is used only to update static rulesets. It should be more complex utility, which takes `id` of the ruleset and calls `chrome.declarativeNetRequest.updateStaticRules()` inside. It also can be moved to `background/dnr.js` - it is only used in background process.

src/background/dnr.js
seia-soto: I believe `disableRuleIds` can be empty in many cases and I'm proceeding with `chrome.runtime.onInstalled` with `details.reason === 'update'` check for now.

src/utils/preprocessor.js
seia-soto: `metadata` is not a module because we're using `import.meta.glob` with `eager: true`. I think the missing `import: 'default'` makes the difference, and yes I tested the feature.

seia-soto: I just re-read https://vite.dev/guide/features#named-imports, and I think it won't change anything than the generated code. Still `import: 'default'` with JSON modules will make the default export as an object.

According to the example,

```
// code produced by vite:
import { default as __vite_glob_0_0 } from './dir/bar.js'
import { default as __vite_glob_0_1 } from './dir/foo.js'
const modules = {
  './dir/bar.js': __vite_glob_0_0,
  './dir/foo.js': __vite_glob_0_1,
}
```

it's automatically picking-up the default export. You probably might had different values when using with other file types or scripts which might have an actual JavaScript function.

seia-soto: Another example in `src/pages/settings/assets/index.js`:

```
export default Object.fromEntries(
  Object.entries(import.meta.glob('./*.svg', { eager: true, import: 'default' })).map(
    ([path, url]) => [path.replace('./', '').replace('.svg', ''), url],
  ),
);
```

seia-soto: Closed https://github.com/ghostery/ghostery-extension/pull/3097#discussion_r2890504320 as pointing to the same issue.

seia-soto: Having a default export as a default import with `import: 'default'` is expected with Vite but it doesn't import as default by default for JSON files without additional transformation.

seia-soto:
Waiting for the new staging build since this cannot be tested properly.

src/background/dnr.js
seia-soto: I proceeded with the way of using `chrome.runtime.onInstalled` with `reason === 'update'` check for now. I also believe that `options.terms` will actually trigger the observer and it makes sense.

seia-soto: For some reasons, the extension doesn't work well with the current staging build and I requested another staging build now.

src/background/dnr.js
smalluban: Don't use `chrome.runtime.onInstalled` - this won't trigger on Safari.

src/utils/preprocessor.js
smalluban: In the tested version from yesterday, the result dist has:
>import * as dnrTracking_metadata from '../rule_resources/dnr-tracking.metadata.json.js';

The `import: 'default'` is an option to automatically use `default` export of module, rahter than return the whole module.

This is just after enabling the extension:
Zrzut ekranu 2026-03-6 o 09 23 02

That version cannot work (42aa977f417ebb4f5756d2ab3505afe77fd35f18). This error is obvious, so I asked if you tested that version.

tests/unit/setup/init-chrome.js
```diff
@@ -22,6 +22,7 @@ Object.defineProperty(global, 'navigator', {
...
     brave: undefined,
+    languages: ['en-US'],
```
smalluban: I asked to move the unit test changes to a separate PR, as this one no longer requires those changes (no new unit tests or loading new files).

src/utils/preprocessor.js
smalluban: You still have `utils/preprocessor.js` - move this to `background/dnr.js`

scripts/download-dnr-rulesets.js
```diff
@@ -42,32 +53,27 @@ for (const [name, target] of Object.entries(RULESETS)) {
...
       .then(filterMaxPriorityRules);
 
     writeFileSync(outputPath, JSON.stringify(dnr, null, 2));
+
+    if (list.dnr.metadataUrl) {
+      const metadata = await fetch(list.dnr.metadataUrl).then(
+        generateJSON('DNR ruleset metadata', name),
+      );
+
+      if (Object.keys(metadata) === 0) {
```
smalluban: Are you sure that `Object.keys()` can return a number and we want to save file only if `Object.keys(metadata) === 0` ?

src/background/dnr.js
```diff
@@ -193,6 +211,27 @@ if (__CHROMIUM__) {
       } catch (e) {
         console.error(`[dnr] Error while updating static rulesets:`, e);
       }
+
+      for (const id of enableRulesetIds) {
+        try {
+          // The below will run when the extension is installed as
+          // well with the change of `options.terms`.
+          const disableRuleIds = await disableExcludedRulesByPreprocessor(id);
```
smalluban: The logging (if needed at all at this detail level) should be done inside the function.

src/background/dnr.js
smalluban: I did not take into account `env_experimental`, which is fully dynamic - it makes it more complicated, as we need to reload disabled rules not only once, but every time the `experimentalFilters` option changes. 

So we don't need `onInstalled` - the update of disabled rules should run each time the main condition passes.

src/utils/preprocessor.js
seia-soto: Please, confirm because my local build shows as the below:

```
import __vite_glob_0_0 from '../rule_resources/dnr-ads.metadata.json.js';
import __vite_glob_0_1 from '../rule_resources/dnr-annoyances.metadata.json.js';
import __vite_glob_0_2 from '../rule_resources/dnr-fixes.metadata.json.js';
import __vite_glob_0_3 from '../rule_resources/dnr-lang-lt.metadata.json.js';
import __vite_glob_0_4 from '../rule_resources/dnr-lang-ru.metadata.json.js';
import __vite_glob_0_5 from '../rule_resources/dnr-tracking.metadata.json.js';
import '../npm/tldts-experimental/npm/tldts-core/dist/es6/src/options.js';
import '../npm/@ghostery/adblocker/dist/esm/data-view.js';
import '../npm/@ghostery/adblocker/dist/esm/fetch.js';
import '../npm/@ghostery/adblocker/dist/esm/filters/cosmetic.js';
import '../npm/@ghostery/adblocker/dist/esm/lists.js';
import '../npm/@ghostery/adblocker/dist/esm/request.js';
import '../npm/@remusao/small/dist/esm/index.js';
import '../npm/@ghostery/adblocker/dist/esm/filters/network.js';
import { evaluate } from '../npm/@ghostery/adblocker/dist/esm/preprocessor.js';
import { ENV } from './engines.js';

// We need to depend on `eager` option since dynamic imports are
// not allowed in web workers scope.
const DNR_METADATA = /* #__PURE__ */ Object.assign({"/rule_resources/dnr-ads.metadata.json": __vite_glob_0_0,"/rule_resources/dnr-annoyances.metadata.json": __vite_glob_0_1,"/rule_resources/dnr-fixes.metadata.json": __vite_glob_0_2,"/rule_resources/dnr-lang-lt.metadata.json": __vite_glob_0_3,"/rule_resources/dnr-lang-ru.metadata.json": __vite_glob_0_4,"/rule_resources/dnr-tracking.metadata.json": __vite_glob_0_5
```

chrmod:
@smalluban @seia-soto lets ignore env_experimental in this PR - its about time we remove Experimental Filters https://github.com/ghostery/broken-page-reports/issues/2182

src/utils/preprocessor.js
seia-soto: I don't think that the output on that version generated `import`s but rather generated inlines instead. However, this is now updated with `import: 'default'` properly.

src/background/dnr.js
seia-soto: How would you compare the version difference? I don't think it's optimal to reuse the "what's new" logic here.

src/utils/preprocessor.js
seia-soto: We can't. It will bring imports from `engines.js` again as mentioned earlier about the dependency chain.

The function rather merged into `disableExcludedRulesByPreprocessor` in `preprocessor.js` now.

smalluban:
@chrmod, I will prepare a PR to remove experimental filters support. It can be done first on the extension side.

src/utils/preprocessor.js
smalluban: Are you sure? Either the `utils/engines.js` and `utils/preprocessor.js` are not imported by any unit tests.

src/background/dnr.js
smalluban: As @chrmod wrote, we are killing experimental filters; please clean up the code.

src/utils/preprocessor.js
seia-soto: However, `utils/dnr.js` is used by `redirect-protection.test.js` and `build-redirect-protection-rules.js`, also `download-dnr-rulesets.js`. If we move that function into `dnr.js`, it will bring import of `ENV` from `engines.js` and it also brings all the chaos again.

smalluban: I want to move that function to `background/dnr.js` - where it is only used...

seia-soto: If you meant the function: `getExcludedRuleIdsByPreprocessors` from the diff shown in this thread. The function is merged into `disableExcludedRulesByPreprocessor` since it's the only use case.

smalluban: Yes, there is only one function in the current `utils/preprocessor.js`.

smalluban: It's fine now.

smalluban:
@seia-soto I see good progress. Please also move the unit tests changes to a separate PR

src/background/dnr.js
```diff
@@ -17,18 +17,55 @@ import Resources from '/store/resources.js';
...
+  async function disableExcludedRulesByPreprocessor(rulesetId) {
+    const metadata = DNR_METADATA[`/rule_resources/dnr-${rulesetId}.metadata.json`];
+    if (!metadata) {
+      return [];
+    }
+    const disableRuleIds = Object.entries(metadata).reduce(function (
+      disabledRuleIds,
+      [ruleId, constraints],
+    ) {
+      if (!evaluatePreprocessor(constraints.preprocessor, ENV)) {
```
smalluban: Use `evaluatePreprocessorCondition` from `utils/engines.js`, so you don't need to import the `adblocker` package.

seia-soto:
@smalluban Apart from the E2E test, I drafted the initial version of the preprocessor with custom filters support. Please, review its direction first. I'm now continuing with E2E tests.

src/background/custom-filters.js
```diff
@@ -24,75 +30,146 @@ import { setup, reloadMainEngine } from './adblocker/index.js';
...
+    return conditions[0];
   }
+  return conditions.map((condition) => `(${condition})`).join('&&');
+}
 
-  const parsedScript = cosmeticFilter.parseScript();
+/**
+ * @returns {Record<'cosmeticFilters' | 'networkFilters' | 'errors', string[]>}
+ */
+function normalizeFilters(text = '', { trustedScriptlets }) {
```
smalluban, line 79: How do all of these changes relate to the preprocessors? If not, please move it to a separate PR.

I expect that change in custom filters should contain one or two lines - just adding a filter to generated DNR rules (to the filters before they are passed to DNR converter).

seia-soto: Those changes definitely include some clean-ups, but I rather prefer injecting the lines into the adblocker library which is capable of executing the preprocessor condition. One thing is that preprocessors will be treated as a comment line unless we inject them as a filter list into the converter because of the syntax characteristics.

seia-soto: We might be able to just pass all the lines instead to the converter but I am not sure how to inject the ENV into the converter in this case.

seia-soto:
@smalluban Please, review again as the Firefox E2E fixed. One thing I would like to mention is that sharing the engine config for the validation might be the best. It create another room for config confusion.

tests/e2e/spec/custom-filters.spec.js
```diff
@@ -96,6 +101,19 @@ describe('Custom Filters', function () {
...
   });
 
+  it('applies preprocessor to the network filter', async function () {
+    await browser.url(PAGE_URL);
+
+    await openPanel();
+    await getExtensionElement('button:detailed-view').click();
+
+    if (browser.isFirefox) {
+      await expect(getExtensionElement(`icon:tracker:www.example.net:blocked`)).toBeDisplayed();
```
smalluban, line 110: Add also `.not.toBeDisplayed()` for the not active filter

scripts/download-dnr-rulesets.js
```diff
@@ -30,8 +30,19 @@ const RULESETS = {
   }, {}),
 };
 
+function generateJSON(type, name) {
```
chrmod, line 33: that does not generate JSON. If you feel a generic helper like this is truly needed, then maybe turn it into `fetchJSON(url)`, so it handles the whole fetching end-to-end ?

src/background/dnr.js
```diff
@@ -27,8 +28,43 @@ if (__CHROMIUM__) {
     .getManifest()
     .declarative_net_request.rule_resources.filter(({ enabled }) => !enabled)
     .map(({ id }) => id);
+  const DNR_METADATA = import.meta.glob('/rule_resources/*.metadata.json', {
```
chrmod, line 31: I dont strongly dislike built system leaking into the codebase. We know all the rulesets already, no need to resolve to the Vite magic.

chrmod: that would also allow this Map to have more natural keys like just `rulesetId`

src/background/dnr.js
```diff
@@ -27,8 +28,43 @@ if (__CHROMIUM__) {
...
+    }, []);
+    try {
+      await chrome.declarativeNetRequest.updateStaticRules({
+        rulesetId: rulesetId,
+        disableRuleIds,
+      });
+    } catch (e) {
+      console.error(`[dnr] Failed to apply preprocessors:`, e);
+    }
+    console.info(
```
chrmod, line 63: maybe `console.debug`?

src/background/dnr.js
```diff
@@ -27,8 +28,43 @@ if (__CHROMIUM__) {
...
+        disabledRuleIds.push(Number(ruleId));
+      }
+      return disabledRuleIds;
+    }, []);
+    try {
+      await chrome.declarativeNetRequest.updateStaticRules({
+        rulesetId: rulesetId,
+        disableRuleIds,
+      });
+    } catch (e) {
```
chrmod, line 60: is this try/catch really needed, the DNR already throws, we dont handle it in any special way, so maybe we should simply let it crash

src/background/dnr.js
```diff
@@ -27,8 +28,43 @@ if (__CHROMIUM__) {
...
+  /**
+   * @param {string} rulesetId
+   * @returns {Promise<number[]>}
+   */
+  async function disableExcludedRulesByPreprocessor(rulesetId) {
+    const metadata = DNR_METADATA[`/rule_resources/dnr-${rulesetId}.metadata.json`];
+    if (!metadata) {
+      return [];
+    }
+    const disableRuleIds = Object.entries(metadata).reduce(function (
```
chrmod, line 46: I like to add new lines to separate preparations from logic

src/background/dnr.js
```diff
@@ -110,6 +146,15 @@ if (__CHROMIUM__) {
...
                 .then(filterMaxPriorityRules),
             );
+            const metadata = new Set(
+              await fetch(list.dnr.metadataUrl)
+                .then((res) =>
+                  res.ok
+                    ? res.json()
+                    : Promise.reject(new Error(`Failed to fetch DNR metadata: ${res.statusText}`)),
+                )
+                .then(filterMaxPriorityRules),
```
chrmod, line 156: why is this needed?

src/background/dnr.js
```diff
@@ -110,6 +146,15 @@ if (__CHROMIUM__) {
                 )
                 .then(filterMaxPriorityRules),
             );
+            const metadata = new Set(
+              await fetch(list.dnr.metadataUrl)
+                .then((res) =>
+                  res.ok
+                    ? res.json()
+                    : Promise.reject(new Error(`Failed to fetch DNR metadata: ${res.statusText}`)),
```
chrmod, line 154: what should happen when that fails? shouldn't we get an empty Set instead of crashing whole procedure? it that case we still want console.error

src/utils/engines.js
```diff
@@ -74,6 +78,10 @@ function deserializeEngine(engineBytes) {
   return engine;
 }
 
+export function evaluatePreprocessorCondition(condition) {
```
chrmod, line 81: I would rename it something that explains the result. This function tells if the condition matches the environment, so that affected filters should apply or not.

src/background/custom-filters.js
```diff
@@ -24,75 +30,154 @@ import { setup, reloadMainEngine } from './adblocker/index.js';
...
+    loadCosmeticFilters: false,
+    loadPreprocessors: true,
+    // The filters will be dropped later in case of unsupported
+    // platform.
+    enableHtmlFiltering: true,
+    // Disable unnecessary internals since this engine is only
+    // for the validation.
+    enableOptimizations: false,
+    enableCompression: false,
+    debug: true,
```
chrmod, line 155: this should be the config we use for `main` engine

src/background/custom-filters.js
```diff
@@ -24,75 +30,154 @@ import { setup, reloadMainEngine } from './adblocker/index.js';
...
-      networkFilters: new Set(),
-      cosmeticFilters: new Set(),
-      errors: [],
-    },
-  );
+      encodeScriptInject(filter, script.args);
+    }
+
+    cosmeticFilters.push(filter.toString());
+  }
```
chrmod, line 142: All the above should be move after `FiltersEngine.parse` - we should work with `.getFilters().cosmeticFilters` instead parsing the list by ourselves

PR #3098, opened 2026-02-23
chore(npm): audit fix & dedupe

smalluban:


smalluban:
There are some packages from the wdio group in the changelog, which make some tests fail (touching notification iframes). I can't fix those tests using this config. As this PR is not a "must have", I am closing this PR for now. We can try to dedupe and fix audit issues after the next batch of npm updates.

PR #3122, opened 2026-02-27
Build: shorten long file paths to avoid windows limits

chrmod:
Shorten long file paths in build output to avoid Windows path length limits

With preserveModules: true, some dependencies (e.g. `@adguard/agtree`) produce output paths that can exceed Windows' MAX_PATH limit. The build already renames node_modules to npm, but deeply nested packages with long filenames still cause issues.

This adds a generic path shortening mechanism in `sanitizeFileName`: when a relative path exceeds 110 characters, the longest segments are iteratively replaced with a deterministic short hash (e.g. `ubo-scriptlet-injection-body-deserializer.js` → `u18p9.js`). Shortened files get a `/* original: ... */` banner comment for traceability.

If two different paths ever shorten to the same name, the build fails with an explicit error naming both, so hash collisions cannot silently produce misleading output. The shortened-path map is reset at the start of each rebuild so watch mode stays consistent.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -352,6 +352,7 @@ function mapPaths(paths) {
   }, {});
 }
 
+const shortenedPaths = new Map();
 const buildPromise = build({
   ...config,
   build: {
@@ -380,10 +381,50 @@ const buildPromise = build({
             .replace('node_modules', 'npm')
             .replace('_virtual', 'virtual');
 
-          const path = name.replace(pwd, '');
-          if (path.length > 110 && !argv['no-filename-limit']) {
+          const PATH_LIMIT = 110;
+          const hasPwd = name.startsWith(pwd);
+          let relPath = hasPwd ? name.slice(pwd.length) : name;
+
+          if (relPath.length > PATH_LIMIT) {
+            const segments = relPath.split('/');
+
+            while (segments.join('/').length > PATH_LIMIT) {
+              let maxLen = 0;
+              let maxIdx = -1;
+
+              for (let i = 0; i < segments.length; i++) {
+                const dotIdx = segments[i].lastIndexOf('.');
+                const base = dotIdx > 0 ? segments[i].slice(0, dotIdx) : segments[i];
+                if (base.length > maxLen) {
+                  maxLen = base.length;
+                  maxIdx = i;
+                }
+              }
+
+              if (maxLen <= 6) break;
+
+              const seg = segments[maxIdx];
+              const dotIdx = seg.lastIndexOf('.');
+              const ext = dotIdx > 0 ? seg.slice(dotIdx) : '';
+              const base = dotIdx > 0 ? seg.slice(0, dotIdx) : seg;
+
+              let hash = 5381;
+              for (let i = 0; i < base.length; i++) {
+                hash = ((hash << 5) + hash + base.charCodeAt(i)) >>> 0;
+              }
+
+              segments[maxIdx] = base[0] + hash.toString(36).slice(0, 4) + ext;
+            }
+
+            const shortened = segments.join('/');
+            shortenedPaths.set(shortened, relPath);
+            relPath = shortened;
+            name = hasPwd ? pwd + relPath : relPath;
+          }
+
+          if (relPath.length > PATH_LIMIT && !argv['no-filename-limit']) {
             throw new Error(
-              `Filename too long: ${path} (${path.length}) (pass --no-filename-limit to disable; for instance, "npm run build -- --no-filename-limit")`,
+              `Filename too long: ${relPath} (${relPath.length}) (pass --no-filename-limit to disable; for instance, "npm run build -- --no-filename-limit")`,
             );
           }
 
@@ -393,6 +434,18 @@ const buildPromise = build({
     },
   },
   plugins: [
+    {
+      name: 'shortened-path-banner',
+      generateBundle(_, bundle) {
+        for (const [fileName, chunk] of Object.entries(bundle)) {
+          if (chunk.type !== 'chunk') continue;
+          const original = shortenedPaths.get('/' + fileName);
+          if (original) {
+            chunk.code = `/* original: ${original} */\n` + chunk.code;
+          }
+        }
+      },
+    },
     // Keep offscreen documents from @whotracksme/reporting
     {
       name: 'copy-reporting-assets',
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -353,7 +353,7 @@ function mapPaths(paths) {
 }
 
 const PATH_LIMIT = 110;
-const SHORTENED_BASE_LENGTH = 5;
+const MIN_SHORTEN_BASE = 6;
 
 function splitSegment(segment) {
   const dotIdx = segment.lastIndexOf('.');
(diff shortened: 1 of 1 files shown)


smalluban (approved):

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
+              for (let i = 0; i < base.length; i++) {
+                hash = ((hash << 5) + hash + base.charCodeAt(i)) >>> 0;
+              }
+
+              segments[maxIdx] = base[0] + hash.toString(36).slice(0, 4) + ext;
+            }
+
+            const shortened = segments.join('/');
+            shortenedPaths.set(shortened, relPath);
+            relPath = shortened;
```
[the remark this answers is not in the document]
chrmod: The hash is deterministic (same input always produces the same output), so two different segment names would need to collide within the same build's file set. With 4 base36 chars (~1.7M possible values) and the first character of the original name preserved, the practical collision risk for our dependency tree is negligible. If it ever does happen, the build would still succeed — Rollup resolves output file names independently of `sanitizeFileName`.

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
+                const dotIdx = segments[i].lastIndexOf('.');
+                const base = dotIdx > 0 ? segments[i].slice(0, dotIdx) : segments[i];
+                if (base.length > maxLen) {
+                  maxLen = base.length;
+                  maxIdx = i;
+                }
+              }
+
+              if (maxLen <= 6) break;
+
```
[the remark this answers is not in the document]
chrmod: This is intentional — the check at line 425 throws an explicit error if the path still exceeds the limit after shortening. So it won't silently produce overlong paths. In practice, paths that are still too long after hashing segments down to 6 chars would indicate an unusually deep nesting structure that we'd want to investigate rather than silently mangle further.

scripts/build.js
```diff
@@ -393,6 +434,18 @@ const buildPromise = build({
     },
   },
   plugins: [
+    {
+      name: 'shortened-path-banner',
```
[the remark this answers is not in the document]
chrmod: This build script is only used for production builds (not watch mode), so stale entries aren't a concern here. Even if it were, `sanitizeFileName` is deterministic — the same input always maps to the same shortened output — so old entries would just be overwritten with the same values.

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
+              }
+
+              if (maxLen <= 6) break;
+
+              const seg = segments[maxIdx];
+              const dotIdx = seg.lastIndexOf('.');
+              const ext = dotIdx > 0 ? seg.slice(dotIdx) : '';
+              const base = dotIdx > 0 ? seg.slice(0, dotIdx) : seg;
+
+              let hash = 5381;
```
philipp-classen: This looks like the DJBX33A algorithm. Would be good to mention that (and best extract as a general hash function to hide the details here).

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
+              for (let i = 0; i < base.length; i++) {
+                hash = ((hash << 5) + hash + base.charCodeAt(i)) >>> 0;
+              }
+
+              segments[maxIdx] = base[0] + hash.toString(36).slice(0, 4) + ext;
```
philipp-classen: This seems to be doing a Base36 encoding (0-9a-z)

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
             .replace('_virtual', 'virtual');
 
-          const path = name.replace(pwd, '');
-          if (path.length > 110 && !argv['no-filename-limit']) {
+          const PATH_LIMIT = 110;
+          const hasPwd = name.startsWith(pwd);
+          let relPath = hasPwd ? name.slice(pwd.length) : name;
+
+          if (relPath.length > PATH_LIMIT) {
+            const segments = relPath.split('/');
```
philipp-classen: Would help to give the reader some intuition what is going on.

It is better commented in the pull request description, as I now see. But note that later most reader will not look at the pull request any more.

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
+
+            while (segments.join('/').length > PATH_LIMIT) {
+              let maxLen = 0;
+              let maxIdx = -1;
+
+              for (let i = 0; i < segments.length; i++) {
```
philipp-classen: This seems to look for the longest folder name or filename (without the path).

scripts/build.js
```diff
@@ -380,10 +381,50 @@ const buildPromise = build({
...
+                  maxLen = base.length;
+                  maxIdx = i;
+                }
+              }
+
+              if (maxLen <= 6) break;
+
+              const seg = segments[maxIdx];
+              const dotIdx = seg.lastIndexOf('.');
+              const ext = dotIdx > 0 ? seg.slice(dotIdx) : '';
```
philipp-classen: This I read as preserving the file extension.

scripts/build.js
philipp-classen: Doing some collision checks would be prudent. If you take 4 Base36 values, you have 36^4 possibilities. That is not so much: 36^4 = 1,679,616

Collisions become more likely than not, once you cross the square root. Here, sqrt(36^4) = 36^2 = 1296 elements.

When I tried it in an example, I also saw the first clashes in that range happening (after 1140 elements, but if you are unlikely, it can happen earlier). One way would be at to detect collision and then failing. Otherwise, this may introduce very nasty bugs.

philipp-classen:
Just as a general idea: because the hashing introduces the chance of collisions. And this there are strong incentives to keep the filenames short, but cannot easily solve it by using larger hashes.

But perhaps it would be an option to avoid hashing at all, but just create name deterministically. "a", "b", "c", ...", "aa", "ab", "abc", ... Keep them in a mapping and do also a check that the new name of the segmentic does not clash with something that not existed before.

But the advantages that I would see is that you should get even small filenames (because you safely start with a single letter.) And you eliminate the risk of collisions.

(If the renaming could be done in a postprocessing step, you would not have to do it on-the-fly. That would simplify parts, because you have already have seen all names and can directly start with the longest ones, which is the one that you want to optimize for. That would lead to optimal mappings, but may be difficult to integrate in the existing build. You would need two-steps: the standard build and then the renaming as an independent postprocess step.)

scripts/build.js
```diff
@@ -375,10 +381,50 @@ const buildPromise = build({
...
+              const base = dotIdx > 0 ? seg.slice(0, dotIdx) : seg;
+
+              let hash = 5381;
```
Tristan-deAlwis: SipHash24 would reduce the likelihood of collisions

philipp-classen:
@chrmod @smalluban I would approve, but the conflict is nasty to resolve since we were based on the February main. What I did is to create here a follow-up PR: https://github.com/ghostery/ghostery-extension/pull/3463

It should be identical but based on the current main. I did not want to force-push, thus the other PR.

philipp-classen:
Merged now as https://github.com/ghostery/ghostery-extension/pull/3463 (identical, but reapplied to resolved conflicts)

Issue #3130, opened 2026-03-03
Better redirect protection target rules identification


seia-soto:
I suggest changing the current redirect protection rule classification into the below.

```diff
diff --git a/src/utils/dnr.js b/src/utils/dnr.js
index ea768ed2..bd1350d1 100644
--- a/src/utils/dnr.js
+++ b/src/utils/dnr.js
@@ -65,7 +65,13 @@ export function getRedirectProtectionRules(rules) {
   const result = [];
 
   for (const rule of rules) {
-    if (rule.action?.type === 'block' && rule.condition?.resourceTypes?.includes('main_frame')) {
+    if (
+      rule.action?.type === 'block' &&
+      rule.condition?.domainType !== 'thirdParty' &&
+      !rule.condition?.excludedResourceTypes?.includes('main_frame') &&
+      (rule.condition?.resourceTypes?.length === 0 ||
+        rule.condition?.resourceTypes?.includes('main_frame'))
+    ) {
       result.push({
         ...rule,
         priority: rule.priority + 1,
```

Currently, we're missing many network filters that need to be handled with redirect protection such as simple form: `||foo.com^` or `||foo.com^$~3p`.

seia-soto:
Also, this will actually add more weight on the expectation towards filter maintainers to write filters with more flags frequently by default. When the request is needed to be fully blocked, they shouldn't put a flag. However, if they want a hostname to be blocked from third-party, they need to put `3p` which actually makes intentions more verbose.

seia-soto:
I also wonder how the exceptions are handled.

smalluban:
https://github.com/ghostery/ghostery-extension/pull/3131#issuecomment-4155493023

PR #3131, opened 2026-03-05
Make redirect protection rule generation more inclusive

seia-soto:
fixes https://github.com/ghostery/ghostery-extension/issues/3130

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -179,6 +179,17 @@ const config = {
     __STAGING__: JSON.stringify(argv.staging),
   },
   build: {
+    rolldownOptions: {
+      ...(argv.watch && {
+        onLog(level, log, handler) {
+          // Suppress FILE_NAME_CONFLICT warnings caused by the watch mode
+          // when files overrite each other during the sequential builds.
+          if (log.code === 'FILE_NAME_CONFLICT') return;
+
+          handler(level, log);
+        },
+      }),
+    },
     outDir: options.outDir,
     assetsDir: '',
     emptyOutDir: false,
@@ -186,7 +197,13 @@ const config = {
     modulePreload: {
       polyfill: false,
     },
-    watch: argv.watch ? {} : null,
+    watch: argv.watch
+      ? {
+          // VS Code editor can trigger multiple save events,
+          // so we add a small delay to prevent multiple builds in a row
+          buildDelay: 200,
+        }
+      : null,
   },
 };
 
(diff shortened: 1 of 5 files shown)


smalluban:
We tested together with @seia-soto that filters in the form of `||domain.com^` produce DNR rule with not set `resourceTypes`, and those generated rules don't block the main frame. It looks like the type of the resource must be set explicitly with `main_frame` to block it.

@seia-soto Please confirm the above with the DNR documentation. If so, the bug is on the converter side (the filter should convert to 2 rules - one with resourceTypes set), as UBlock Origin uses this type of filter to block navigation to the site.

image

FYI @chrmod

seia-soto:
@smalluban After investigating in, it's now pretty clear that the behaviour was related to the DNR converter or other components. Including Chrome 84's IDL definition, it states the following: if neither resourceTypes nor excludedResourceTypes is specified, all resource types except main_frame are matched. Then I couldn't find anything from the converter.

smalluban:
Can you try to find a solution on the DNR converter side?

seia-soto:
I believe that we don't have any issues right now. We ship the bypass rule to mitigate the issue from our backend and now we're filtering it on the build time. I need to check if it still causes the same issue because users were seeing "foo.com is blocked, ... disable extension ..." message as a fallback before the introduction of the redirect protection.

If everything is alright, we can even remove that custom rule from our backend and simplify the build logic as well. The converter itself doesn't generate a rule with max priority.

smalluban:
If Ublock Origin shows the redirect protection page for filters like `||example.com^`, indeed there is a bug in dnr converter - as those filters should produce a rule with `resourceTypes` set to `main_frame` too... And they don't do it right now, don't they?

seia-soto:
They don’t do it right now as suspected.

smalluban:
>They don’t do it right now as suspected.

Does it mean that they don't block those main_frame requests anymore? @seia-soto, what do you recommend? Do you close the PR, change it... Do we make a change in the converter?

seia-soto:
I believe we still need to show redirect protection screen for newly eligible filters. I'm on it.

seia-soto:
The status here is

1. I'm not sure which exactly caused it but Chrome states that the DNR rule will not block main frame unless specified in `resourcesType` constraint. It's also stated in Chrome v84 IDL.
2. We still want to make the redirect protection more inclusive to match the behaviour.

smalluban:
The problem is on the converter side, not the condition. As your point 1 states - Chrome does not block main frame requests until they are specified - we should do the same. However, it looks like filters like `||example.com^` should be converted to 2 rules (or one, if possible), which also includes `resourcesType` set with `main_frame`.

seia-soto:
Deciding this out of scope on extension side, and closing. If we need it, we're moving this to the urlfilter2dnr.

PR #3132, opened 2026-03-05
fix(telemetry): `modeTouched` and async ping execution fixes

smalluban:
This PR fixes two issues with the telemetry system:
* The update of the `modeTouched` flag moved to the settings page, to the change listener of modes options. The condiation, which was in `telemetry/index.js` looks fine at first, but when user sees "choose mode" screen and clicks "zap", there is a change from "default" to "zap", so the options observer runs the callback, and the condition is "true" (`lastOptions?.mode && lastOptions.mode !== mode`). As a result, `modeTouched` is set to `true` (this is unintentional, but still logical). The `install_complete` was executed after this code runs, so the `zap` query parameter was always `3`. With this change, the `modeTouched` is connected to the right moment/place - to the change of the setting (as we wanted from the business logic perspective)
* The sequencer must wrap the whole body of `_record*` methods, as the `_recordActive` relies on `storage.install_all`, which is only set after a successful request with `install` ping. I tried a few methods to fix it, but it looks like wrapping the whole public API - the `ping()` method is the simplest solution. It ensures that `_record*` methods are executed only after the previous ping finishes.

REBASE when merging.

### DIFF
--- a/src/background/telemetry/metrics.js
+++ b/src/background/telemetry/metrics.js
@@ -110,7 +110,7 @@ export default class Metrics {
     this.log = log;
     this.saveStorage = saveStorage;
     this.storage = storage || {};
-    this._sendReqLock = new SeqExecutor();
+    this._seqExecutor = new SeqExecutor();
   }
 
   /**
@@ -126,26 +126,28 @@ export default class Metrics {
    * @param {string} type    type of the telemetry ping
    */
   ping(type) {
-    switch (type) {
-      // Key Performance Metrics
-      case 'install':
-        this._recordInstall();
-        break;
-      case 'active':
-        this._recordActive();
-        break;
-      case 'engaged':
-        this._recordEngaged();
-        break;
-      case 'install_complete':
-        this._recordInstallComplete();
-        break;
+    return this._seqExecutor.run(async () => {
+      switch (type) {
+        // Key Performance Metrics
+        case 'install':
+          await this._recordInstall();
+          break;
+        case 'active':
+          await this._recordActive();
+          break;
+        case 'engaged':
+          await this._recordEngaged();
+          break;
+        case 'install_complete':
+          await this._recordInstallComplete();
+          break;
 
-      // Uncaught Pings
-      default:
-        this.log(`metrics ping() error: ping name ${type} not found`);
-        break;
-    }
+        // Uncaught Pings
+        default:
+          this.log(`metrics ping() error: ping name ${type} not found`);
+          break;
+      }
+    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/background/telemetry/metrics.js
```diff
@@ -250,97 +252,95 @@ export default class Metrics {
...
-
-        const ONE_YEAR = 1000 * 60 * 60 * 24 * 365;
-        if (this.storage[key] > now + ONE_YEAR) {
-          this.log(
-            `ping: resetting metrics (type=${type}, frequency=${frequency}) [clock jump detected]`,
-          );
-          this.storage[key] = now;
-          storageDirty = true;
-        }
+    const now = Date.now();
```
smalluban, line 255: Here we only remove the `_sendReqLock` wrapper. The logic is untouched.

src/background/telemetry/metrics.js
```diff
@@ -487,7 +488,7 @@ export default class Metrics {
     this.storage.engaged_daily_count = engaged_daily_count;
     this.storage.engaged_daily_velocity = engaged_daily_velocity;
 
-    this.saveStorage(this.storage)
+    return this.saveStorage(this.storage)
```
chrmod, line 491: why introducing this change?

src/background/telemetry/metrics.js
```diff
@@ -456,7 +457,7 @@ export default class Metrics {
     if (this.storage.install_complete_all) {
       return;
     }
-    this._sendReq('install_complete').catch((err) => {
+    return this._sendReq('install_complete').catch((err) => {
```
chrmod, line 460: why introducing this change?

src/background/telemetry/metrics.js
```diff
@@ -441,7 +442,7 @@ export default class Metrics {
     }
 
     this.storage.active_daily_velocity = active_daily_velocity;
-    this.saveStorage(this.storage)
+    return this.saveStorage(this.storage)
```
chrmod, line 445: why introducing this change?

src/background/telemetry/metrics.js
```diff
@@ -415,7 +415,8 @@ export default class Metrics {
     if (this.storage.install_all) {
       return;
     }
-    this._sendReq('install').catch((err) => {
+
+    return this._sendReq('install').catch((err) => {
```
chrmod, line 419: why introducing this change?

src/pages/settings/views/my-ghostery.js
```diff
@@ -38,6 +38,13 @@ async function importSettings(host, event) {
   }
 }
 
+function updateMode(value) {
+  return async (host) => {
+    await store.set(host.options, { mode: value });
+    chrome.runtime.sendMessage({ action: 'telemetry:modeTouched' });
```
chrmod, line 44: little inconvenient that we need a special handler

src/background/telemetry/metrics.js
```diff
@@ -110,7 +110,7 @@ export default class Metrics {
     this.log = log;
     this.saveStorage = saveStorage;
     this.storage = storage || {};
-    this._sendReqLock = new SeqExecutor();
+    this._seqExecutor = new SeqExecutor();
```
chrmod, line 113: why nameing?

src/background/telemetry/metrics.js
```diff
@@ -250,97 +252,95 @@ export default class Metrics {
    * @param {array} 		[frequencies = ['all']] 	array of ping frequencies
    */
   async _sendReq(type, frequencies = ['all']) {
-    return this._sendReqLock.run(async () => {
```
chrmod, line 253: to reduce the indentation diff, we could replace it with IIFE

src/pages/settings/views/my-ghostery.js
philipp-classen, line 44: Should this also return the result (or await it)?

src/background/telemetry/metrics.js
smalluban: I explained in the description:
> The sequencer must wrap the whole body of _record* methods, as the _recordActive relies on storage.install_all, which is only set after a successful request with install ping. 

The ping method uses now `await this._record*()`. We need to return this promise chain.

src/background/telemetry/metrics.js
smalluban: It is no longer only about the request.

src/pages/settings/views/my-ghostery.js
smalluban: The UI does not use the result of this method. It does not matter.

src/background/telemetry/metrics.js
smalluban: Now or later, the IIFE is pointless, so you cannot avoid PR with just an indentation difference.

src/pages/settings/views/my-ghostery.js
smalluban: We want to update an instance of `Metrics`, as it might be just running. This is how we implemented it.

src/background/telemetry/metrics.js
smalluban: https://github.com/ghostery/ghostery-extension/pull/3132#discussion_r2890186696

src/background/telemetry/metrics.js
smalluban: https://github.com/ghostery/ghostery-extension/pull/3132#discussion_r2890186696

src/background/telemetry/metrics.js
smalluban: https://github.com/ghostery/ghostery-extension/pull/3132#discussion_r2890186696

src/background/telemetry/metrics.js
chrmod: it is not about PR but about diffs and blames

PR #3233, opened 2026-04-08
chore(build): minimize DNR rulesets via requestDomains folding

chrmod:
Merge bare-domain `urlFilter` rules into `requestDomains` arrays and split large rulesets for Safari compilation time limits.

## What it does

Rules whose `urlFilter` is `||<fqdn>^` are grouped by action, priority, and remaining condition properties, then collapsed into rules with a `requestDomains` array holding the union of domains (1,000 domains per rule). Groups with only a single domain keep their original rule unchanged.

For large rulesets (`ads` and `tracking`), the output is split into multiple files capped at ~10,000 effective Safari rules each:

**ads** (73,877 → 8,508 rules, 12.46 MB → 2.56 MB):
| File | DNR rules | Effective Safari rules |
|---|---|---|
| `dnr-ads-domains-0.json` … `-5.json` | 10 each | ~10,000 each |
| `dnr-ads-domains-6.json` | 66 | ~5,957 |
| `dnr-ads-0.json` | 5,000 | ~5,000 |
| `dnr-ads-1.json` | 3,381 | ~3,381 |

**tracking** (56,297 → 9,735 rules, 8.94 MB → 2.12 MB):
| File | DNR rules | Effective Safari rules |
|---|---|---|
| `dnr-tracking-domains-0.json` … `-3.json` | 10 each | ~10,000 each |
| `dnr-tracking-domains-4.json` | 31 | ~6,633 |
| `dnr-tracking-0.json` | 5,000 | ~5,000 |
| `dnr-tracking-1.json` | 4,664 | ~4,664 |

The build discovers split files automatically and expands them into the manifest. The runtime enables all sub-rulesets (e.g. `ads-0`, `ads-domains-0`, …) when the parent engine (`ads`) is toggled.

## Why this is safe

`||domain.tld^` in Chrome's DNR `urlFilter` and `requestDomains: ["domain.tld"]` have identical matching semantics:

- **Left boundary**: `||` anchors to the start of the host or after a `.` (subdomain boundary). `requestDomains` does suffix matching with dot-boundary enforcement via Chromium's `url::DomainIs()`. Same result.
- **Right boundary**: `^` matches any separator character. After a hostname in any valid URL, the next character (`/`, `:`, `?`, `#`, or end-of-string) is always a separator — so `^` is always satisfied. `requestDomains` doesn't need this check because it matches the parsed host component directly.
- **Iframes / nested iframes**: Both match against the **request URL** (the URL being fetched), not the initiator or frame hierarchy. Frame nesting is irrelevant.
- **All resource types**: Both apply identically to `main_frame`, `sub_frame`, `xmlhttprequest`, `image`, etc.

### Metadata / preprocessor constraints

Rules with preprocessor metadata entries are **excluded from merging** and kept as their original rule objects. This preserves the ability to disable them per-rule via `chrome.declarativeNetRequest.updateStaticRules`. Metadata files are split across output files and re-keyed to match renumbered rule IDs.

## Safari / WebKit compatibility

WebKit does not evaluate `requestDomains` as a single array lookup at match time. Instead, in [`-[_WKWebExtensionDeclarativeNetRequestRule ruleInWebKitFormat]`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm) (around line 832), it loops over each entry and emits a separate content blocker rule per domain:

```objc
if (requestDomains && !_condition[regexFilterKey]) {
    for (NSString *requestDomain in requestDomains) {
        NSString *combined = [self _combineRequestDomain:requestDomain withURLFilter:...];
        // → one content blocker rule per domain
    }
}
```

This means WebKit effectively reverses the folding at compile time. The optimization is therefore **size-neutral** for Safari's 150,000 compiled-rule hard limit ([`ContentExtensionParser.cpp`](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/contentextensions/ContentExtensionParser.cpp), `constexpr size_t maxRuleCount = 150000`). Splitting into multiple static rulesets (capped at ~10k effective rules each) keeps per-ruleset compilation within Safari's time budget.

## What we intentionally skip

- **IP address patterns** (~143 rules like `||192.168.1.1^`): Chrome's docs specify `requestDomains` expects "a canonical domain." While `url::DomainIs()` would likely handle IPs via raw suffix matching, this is undocumented behavior — not worth the risk for so few rules.
- **Single-label domains** (1 rule: `||undefined^`): A hostname with no TLD. `requestDomains` behavior for single-label names is unspecified.

## Theoretical edge case (not a practical concern)

`||` does not enforce a right-side **label** boundary on its own — `||examp` would match `example.com`. However, for bare-domain patterns `||domain.tld^`, the `^` separator after the TLD prevents partial-label matching (`.` is not a separator for `^`). So `||tracker.com^` does not match `tracker.company.com`, and neither does `requestDomains: ["tracker.com"]` (dot-boundary enforcement). Both agree.


smalluban (approved):
LGTM

seia-soto (changes requested):
Making sure this file is not overwriting the `rule_resources` under `src` directory will make the further development comfier by not replacing the source DNR rulesets.

scripts/build.js
```diff
@@ -90,6 +90,11 @@ if (argv.target === 'chromium') {
   execSync('node scripts/build-redirect-protection-rules.js', {
     stdio: silent ? '' : 'inherit',
   });
+
+  // Merge bare-domain urlFilter rules into requestDomains arrays (in-place, idempotent)
+  execSync('node scripts/minimize-dnr-rulesets.js', {
```
seia-soto, line 95: I recommend to run this "after" `scripts/download-redirect-resources.js`, so we can apply it even with freshly downloaded rulesets.

seia-soto:
This has to be revalidated with the actual source citation because mainly I couldn't actually find the source of the timeout relevance with dNR compilation. Also, it seems like the size reduction is basically overlooked because there would be more overhead with encoding and JSON formatting at the end by splitting into multiple files.

It's quite meaningful that compressing into one rule grouped with `requestDomains` especially for the dNR rule count limitation. However, I don't see a straightforward reason for splitting into multiple chunks or ruleset parts like `ads-0.json`.

smalluban:
@chrmod Can we split these two optimizations into separate PRs? The original one (to combine multiple rules into one) should be safe. @seia-soto ?

The PR description doesn't even mention splitting into chunks.

seia-soto:
I'll create a separate pull request for requestDomains grouping.

PR #3235, opened 2026-04-08
fix(storage): clear storage.session when reaching quota limit

smalluban:
This PR adds a check for session storage quota on startup to prevent issues with quota limits.

The WTM reporting package uses `storage.session` directly, so we can't create an abstract layer to protect access to the storage.

Also, the condition to fallback to `storage.local` was added to support Safari 15 (https://github.com/ghostery/ghostery-extension/pull/1340). All currently supported platforms provide `storage.session`.

### DIFF
--- a/src/background/storage.js
+++ b/src/background/storage.js
@@ -1,0 +1,45 @@
+/**
+ * Ghostery Browser Extension
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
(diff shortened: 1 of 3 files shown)


src/background/storage.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const bytesInUse = await chrome.storage.session.getBytesInUse();
+    const usage = bytesInUse / quotaBytes;
+
+    if (usage >= QUOTA_THRESHOLD) {
+      console.warn(
+        `[storage] Session storage usage at ${(usage * 100).toFixed(1)}% (${bytesInUse}/${quotaBytes} bytes). Clearing session storage.`,
+      );
+
+      await chrome.storage.session.clear();
```
chrmod, line 32: lets start without cleaning just reporting. Be sure to send report not more than once a day

chrmod (changes requested):

smalluban:
For now, we only want to add a check, which was added with #3276

PR #3254, opened 2026-04-13
Paused domains estimation support

philipp-classen:
urlReporter:
- Provide access to the state of paused domains
- Expose a transaction API to the wrapped IndexDB

Provides the integration for https://github.com/whotracksme/webextension-packages/pull/193

### DIFF
--- a/src/background/reporting/url-reporter.js
+++ b/src/background/reporting/url-reporter.js
@@ -12,13 +12,33 @@
 import { UrlReporter } from '@whotracksme/reporting/reporting';
 
 import config from './config.js';
+import { getPausedDetails } from '/store/options.js';
+import * as OptionsObserver from '/utils/options-observer.js';
 import communication from './communication.js';
 import prefixedIndexedDBKeyValueStore from './storage-indexeddb.js';
 import StorageLocal from './storage-chrome-local.js';
+
+let options = {};
+OptionsObserver.addListener(function urlReporting(value) {
+  options = value;
+});
 
 export default new UrlReporter({
   config: config.url,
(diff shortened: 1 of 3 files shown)


src/background/reporting/url-reporter.js
```diff
@@ -12,13 +12,33 @@
...
   storage: new StorageLocal('reporting'),
   connectDatabase: prefixedIndexedDBKeyValueStore('reporting'),
   communication,
+
+  pauseState: {
+    getFilteringMode: () => options.mode,
+    isHostnamePaused: (hostname) => !!getPausedDetails(options, hostname),
+    connectHostnamePausingEvents: (notify) => {
+      chrome.runtime.onMessage.addListener((msg) => {
+        if (msg.action === 'updateHostnamePause') {
```
smalluban: Use `reporting:` prefix to action name. Where it fits, we add them.

smalluban (approved):

src/pages/panel/views/main.js
```diff
@@ -90,6 +90,14 @@ async function togglePause(host, event) {
     });
   }
 
+  if (stats.hostname) {
```
smalluban: I am not sure if you need this condition. The pause button is only displayed in UI if the `stats.hostname` is set. The code above uses `stats.hostname` and it does not check if it is filled. 

This is the code, which secures it: https://github.com/ghostery/ghostery-extension/blob/main/src/background/stats.js#L324

philipp-classen: Yes, looks unnecessary. I removed it.

PR #3265, opened 2026-04-17
fix(build): bump target version for Safari to 26.0

smalluban:
This PR bumps target version for Safari to 26.0. The macOS target can stay at `14.6` as the macOS 14 is the oldest version, which supports Safari 26.


chrmod (changes requested):
I think it's too early. iOS 26 adoption is still around 80%. lets wait until mid summer and revisit

PR #3271, opened 2026-04-20
chore(deprecate): `filterMaxPriorityRules`

seia-soto:
Deprecating this method as we no longer handle main frame exception via backend. Please, make sure the backend is shipping rulesets without the exception rule with the max priority.

### DIFF
--- a/src/background/dnr.js
+++ b/src/background/dnr.js
@@ -14,7 +14,7 @@ import { store } from 'hybrids';
 import { ENGINES, isGloballyPaused } from '/store/options.js';
 import Resources from '/store/resources.js';
 
-import { FIXES_ID_RANGE, getDynamicRulesIds, filterMaxPriorityRules } from '/utils/dnr.js';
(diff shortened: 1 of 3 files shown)


src/background/dnr.js
```diff
@@ -14,7 +14,7 @@ import { store } from 'hybrids';
 import { ENGINES, isGloballyPaused } from '/store/options.js';
 import Resources from '/store/resources.js';
 
-import { FIXES_ID_RANGE, getDynamicRulesIds, filterMaxPriorityRules } from '/utils/dnr.js';
+import { FIXES_ID_RANGE, getDynamicRulesIds } from '/utils/dnr.js';
```
smalluban, line 17: Is `filterMaxPriorityRules()` used anywhere now? If not, please remove this function from the codebase.

seia-soto: The function is removed by the first commit. The search of "filterMaxPriorityRules" returns zero result on my side. Let me know if I missed something.

smalluban (approved):
I must have seen changes of one commit - everything is correct.

PR #3272, opened 2026-04-20
feat(ci): concurrent dnr rulesets download

seia-soto:
Quick refactor of downloading dnr rulesets because CDN is slower in my network and we have decent amount of rulesets.

Run the following to test: `rm -r src/rule_resources && node scripts/download-dnr-rulesets.js`.


scripts/download-dnr-rulesets.js
```diff
@@ -39,36 +39,120 @@ function handleResponse(res) {
...
+  if (dnr) {
     writeFileSync(outputPath, JSON.stringify(dnr, null, 2));
+  }
 
-    if (list.dnr.metadataUrl) {
-      const metadata = await fetch(list.dnr.metadataUrl).then(handleResponse);
+  if (metadata && Object.keys(metadata).length !== 0) {
+    writeFileSync(metadataPath, JSON.stringify(metadata, null, 2));
+  } else if (existsSync(metadataPath)) {
+    rmSync(metadataPath);
```
smalluban: Are you sure that this case might happen? If the DNR ruleset exists, the whole logic for that ruleset is skipped (L118). It means that the above code runs only if there is no file.

smalluban (changes requested):

scripts/download-dnr-rulesets.js
seia-soto: L118 is keeping the original behaviour. We only download the ruleset if there's nothing. The code above (L66) is to remove the stale metadata file. If the metadata file was saved by the previous script calls and the latest list query shows that there's no metadata anymore, the file should be deleted. If we don't delete it, the legacy metadata file will be queried on the runtime by the extension and causes uninteded behaviour.

seia-soto: If we only consider the case when the directory is fully cleaned by `--clean` flag, we don't need to do that.

smalluban: I am sure that if the ruleset exists, still with your code you omit download and processing:
```
if (existsSync(outputPath)) {
    return prems;
  }
```

So, there is no way that you download metadata that already exists...

seia-soto: That code path happens if

1. `outputPath` doesn't exist
2. legacy metadata file exist
3. up-to-date list response with no metadata (or empty metadata)

In this case, the legacy metadata file will survive without the code. It cannot happen with the clean state. I wonder if we should handle this or not.

smalluban: There is no way that `outputPath` does not exist but the metadata does. We use `--clean` flag to remove all of the resources. 
You should skip this case. Clean up the code.

seia-soto: Removed.

smalluban (approved):
The changes look fine - please merge your changes with the current main (there is a change in grouping DNRs logic), and we can go with it

PR #3277, opened 2026-04-21
fix(adblocker): split logic into separate files

smalluban:
The upcoming GPC feature requires new logic in the adblocker module. As the codebase grows rapidly, keeping all code in one file is hard to read and reason about.

This PR splits the logic into a number of files, keeping the logic of the module untouched.


src/background/index.js
```diff
@@ -12,8 +12,10 @@
 import './onboarding.js';
 import './config.js';
 
+import './adblocker/network.js';
+import './adblocker/cosmetics.js';
```
chrmod: I would keep `./adblocker/index.js` just for sake of background/index.js having only one import

smalluban: I don't mind bringing it back, but I think the adblocker features are separate (eg. `network.js` is only for Firefox). Also, as we import directly from `adblocker/engines.js`, the `index.js` file looked redundant to me.

chrmod (approved):

PR #3280, opened 2026-04-21
feat(safari): strip DNR rules WebKit cannot compile

chrmod:
## Summary

- Safari's WebKit URL filter parser silently rejects `urlFilter` / `regexFilter` patterns it does not support (disjunctions, character classes, arbitrary atom repetitions, etc.). Those rules ship to users but never match, producing gaps in ads/tracking/annoyance protection specific to Safari.
- The Xcode build now runs the [ghostery/WebKit `validate-dnr-rules`](https://github.com/ghostery/WebKit/pull/4) tool against every ruleset produced in `dist/rule_resources/` — `dnr-ads`, `dnr-tracking`, `dnr-fixes`, `dnr-annoyances`, `dnr-redirect-protection`, and all `dnr-lang-*` regional lists — and drops any rules WebKit would reject.
- The validator binary is fetched lazily from the latest `ghostery/WebKit` release by a `postinstall` hook and kept out of the repo. Unsupported host platforms (e.g. Windows, Linux arm64) skip the download silently so regular `npm install` is unaffected; Xcode Cloud's macOS arm64 runners pick up the correct asset.

On the current rulesets this removes 342 rules across 29 of 31 files — top offenders are `dnr-ads` (130), `dnr-redirect-protection` (80), `dnr-tracking` (18), `dnr-lang-pl` (19), `dnr-lang-tr` (16), `dnr-lang-ja` (12).

## Test plan

- Install Ghostery from a TestFlight build produced by this branch on Safari (macOS and iOS) and confirm the extension loads without ruleset-compilation errors in the system log.
- Browse a handful of sites that rely on rules previously rejected by Safari (e.g. a page matched by one of the removed `dnr-ads` regex rules) and verify the page still renders correctly — the rule being dropped should not regress user-visible behaviour, since it was already a no-op in Safari.
- Toggle regional language lists in the extension settings on Safari and confirm they still enable/disable without console errors.


package.json
```diff
@@ -4,6 +4,7 @@
   "version": "10.5.39",
   "type": "module",
   "scripts": {
+    "postinstall": "node scripts/download-validate-dnr-rules.js",
```
smalluban: Why postinstall? It should be only a part of the Xcode build - it is only used by `filter-invalid-dnr-rules.js`, so it could also be a part of that script.

smalluban (changes requested):

smalluban (approved):

PR #3296, opened 2026-04-27
chore(ci): add translations update GitHub Action

smalluban:
## Summary
Add a GitHub Actions workflow that runs on pushes to `main`, executes `npm run locales`, and opens/updates a `chore-translations` PR branch when locale files change.

If we set up integration with Transifex correctly, once the PR is open, Transifex will pull the updated source locale file and push it back to the `chore-translations` branch when we translate it.

## Details
- Triggers on push to `main`
- Runs locale generation
- Detects changes in English locale output
- Uses `peter-evans/create-pull-request` to create or update PRs on `chore-translations`


.github/workflows/update-translations.yml
```diff
@@ -0,0 +1,56 @@
...
+
+          if [ -f src/_locales/en/messages.json ] && ! git diff --quiet -- src/_locales/en/messages.json; then
+            changed=true
+          fi
+
+          if [ "$changed" = false ] && [ -f src/locales/en/messages.json ] && ! git diff --quiet -- src/locales/en/messages.json; then
+            changed=true
+          fi
+
+          echo "changed=$changed" >> "$GITHUB_OUTPUT"
```
seia-soto: Please, use:

```
git diff --quiet -- src/_locales/en/messages.json src/locales/en/messages.json; [ $? -eq 1 ] && echo "changed=true" >> "$GITHUB_OUTPUT"
```

seia-soto (changes requested):

.github/workflows/update-translations.yml
smalluban: I simplified the command, but I checked it with copilot:

```
Notes on the reviewer's exact snippet vs. what I applied:

I kept if ! git diff --quiet ...; then instead of git diff ...; [ $? -eq 1 ] because GitHub Actions runs the script under bash -eo pipefail. With the reviewer's form, exit code 1 from git diff would trip errexit and abort the step before the [ $? -eq 1 ] test runs. Using if ! is exempt from errexit and behaves correctly.
Dropped the second path (src/locales/...) entirely since it doesn't exist in the repo.
```

seia-soto: I overlooked that part, it should be OR cond in the sh like: `git diff --quiet -- path/to/file || echo "changed=true" >> "$GITHUB_OUTPUT"`. However, with this version and that version as well, I concern both don't catch the real error codes (1>). Might be nice to handle exit code 1 (has change) explicitly.

seia-soto (approved):

PR #3297, opened 2026-04-28
fix(build): filter unsupported preprocessor envs from DNR metadata

smalluban:
Filters DNR rule metadata entries whose preprocessor expressions reference only environment identifiers the extension does not support. Such entries always evaluate to a constant (true or false) at runtime, so the rules they describe cannot be meaningfully toggled.

Changes:
- Add `scripts/utils/filter-metadata.js` defining the supported envs (mirrored from `ENV` in `src/utils/engines.js`) and a `filterMetadata` helper that drops entries whose result is constant across all combinations of supported envs.
- Apply the filter in `scripts/download-dnr-rulesets.js`; skip writing the metadata file when the filtered result is empty.
- Extend supported envs in `src/utils/engines.js`: `env_safari` (via `isSafari()`) and `env_mv3` (always `true`, since the extension is MV3-only).
- Move `group-dnr-rulesets.js` under `scripts/utils/`.


src/utils/engines.js
```diff
@@ -38,6 +39,7 @@ const engines = new Map();
 export const ENV = new Map([
   ['ext_ghostery', true],
   ['ext_ublock', true],
+  ['env_mv3', true],
```
chrmod: this should read the manifest

scripts/utils/filter-metadata.js
```diff
@@ -0,0 +1,85 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { evaluatePreprocessor } from '@ghostery/adblocker';
+
+// Keep in sync with the ENV map in src/utils/engines.js. These are the only
+// preprocessor identifiers the extension knows how to evaluate at runtime;
+// any other identifier is treated as `false` by `evaluatePreprocessor`.
+const SUPPORTED_ENVS = [
```
chrmod: can we somehow have one source of truth? I don't want to risk we add/remove support for a new env and forget we have to update it here.

chrmod (changes requested):

src/utils/engines.js
smalluban: My mistake, as I assumed this was only for DNRs, but obviously it does not. I will fix it.

src/utils/preprocessors.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Environment variables for preprocessor evaluation. These define which
+ * environment identifiers the extension supports and their current values.
+ * This map is the source of truth for all supported preprocessor envs.
+ */
+export const ENV = new Map([
+  ['ext_ghostery', true],
+  ['ext_ublock', true],
+  ['env_mv3', globalThis.chrome?.runtime.getManifest().manifest_version === 3],
+  ['ext_ubol', __FIREFOX__],
```
chrmod, line 27: shouldn't that be `__CHROMIUM__`?

chrmod: `ext_ublock` should be set for `MV2` only
and `ext_ubol` should be set for `MV3`

chrmod (changes requested):

chrmod (approved):

PR #3298, opened 2026-04-28
fix(redirect-protection): align Firefox main_frame matching with DNR

chrmod:
## Summary

On Firefox, type-less filters (e.g. `||tracker.com^` with no `$document` modifier) were triggering the redirect-protection warning page on top-level navigations. This diverges from Chrome MV3, where DNR's safety default excludes `main_frame` for rules that don't specify `resourceTypes`.

The behavior originated in `@ghostery/adblocker`, whose `FROM_ANY` mask matches `main_frame` for type-less filters. The Firefox webRequest handler in `src/background/adblocker/network.js` then engaged the redirect-protection branch on every such match.

We now restrict the main_frame branch to filters with an explicit `$document` modifier (`!filter.fromAny()`), matching DNR's behavior and the filter authors' intent. As a side effect, the silent pass-through caused by returning `{ redirectUrl: undefined }` is gone — main_frame handling lives entirely inside the document-only branch.

The `Type-less filter does not affect main_frame` e2e suite covers both toggle states.

## Test plan

- [ ] On Firefox, navigate to a page that matches a custom filter without `$document` (e.g. add `||example.com^` to custom filters and visit `https://example.com`). Expect: page loads normally, no warning page.
- [ ] On Firefox, with a `$document` custom filter (`||example.com^$document`) and redirect protection on, visit the URL. Expect: warning page appears (existing behavior).
- [ ] On Firefox, with a `$document` custom filter and redirect protection off, visit the URL. Expect: page loads normally (existing behavior).
- [ ] On Chromium, repeat all three scenarios. Expect: identical behavior — DNR was already correct here.

### DIFF
--- a/src/background/adblocker/network.js
+++ b/src/background/adblocker/network.js
@@ -90,11 +90,7 @@ if (__FIREFOX__) {
           // a resource type — we mirror that here.
           if (match === true && !filter?.fromAny()) {
             const options = store.get(Options);
-            const redirectUrl = getRedirectProtectionUrl(
-              details.url,
-              request.hostname,
-              options,
-            );
+            const redirectUrl = getRedirectProtectionUrl(details.url, request.hostname, options);
 
             return { redirectUrl };
           }
--- a/src/background/adblocker/network.js
+++ b/src/background/adblocker/network.js
@@ -80,13 +80,24 @@ if (__FIREFOX__) {
 
       if (isMatchableRequest(details, request)) {
         const engine = engines.get(engines.MAIN_ENGINE);
-        const { redirect, match } = engine.match(request);
+        const { redirect, match, filter } = engine.match(request);
 
-        if (match === true && details.type === 'main_frame') {
-          const options = store.get(Options);
-          const redirectUrl = getRedirectProtectionUrl(details.url, request.hostname, options);
+        if (details.type === 'main_frame') {
+          // For main_frame, only consider matches from filters with an
+          // explicit $document modifier. Type-less filters match main_frame
+          // via @ghostery/adblocker's FROM_ANY mask, but Chrome MV3 DNR's
+          // safety default excludes main_frame for filters that don't specify
+          // a resource type — we mirror that here.
+          if (match === true && !filter?.fromAny()) {
+            const options = store.get(Options);
+            const redirectUrl = getRedirectProtectionUrl(
+              details.url,
+              request.hostname,
+              options,
+            );
 
-          return { redirectUrl };
+            return { redirectUrl };
+          }
         } else if (redirect !== undefined) {
           request.blocked = true;
           // There's a possibility that redirecting to file URL can expose
--- a/src/background/adblocker/network.js
+++ b/src/background/adblocker/network.js
@@ -83,12 +83,10 @@ if (__FIREFOX__) {
         const { redirect, match, filter } = engine.match(request);
 
         if (details.type === 'main_frame') {
-          // For main_frame, only consider matches from filters with an
-          // explicit $document modifier. Type-less filters match main_frame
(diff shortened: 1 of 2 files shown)


smalluban (approved):

src/background/adblocker/network.js
```diff
@@ -80,13 +80,20 @@ if (__FIREFOX__) {
...
-          return { redirectUrl };
+        const { redirect, match, filter } = engine.match(request);
+
+        if (details.type === 'main_frame') {
+          // For main_frame, only consider matches from filters with an
+          // explicit $document modifier. Type-less filters match main_frame
+          // via @ghostery/adblocker's FROM_ANY mask, but Chrome MV3 DNR's
+          // safety default excludes main_frame for filters that don't specify
+          // a resource type — we mirror that here.
+          if (match === true && !filter?.fromAny()) {
```
philipp-classen: Getting confused with the comments and the code.

!filter?.fromAny()
<=>
!filter || !filter.fromAny()

but !filter.fromAny() is not the same as filter.fromDocument

seia-soto (changes requested):
Why do we correct Firefox behaviour rather correct MV3 behaviour for the redirect protection which strictly follows uBlock Origin behaviour? `$document` modifier mentioned in the description overlooks its role and the behaviour depends on the platform: uBlock Origin or AdGuard. The thing is that we're using AdGuard converter and it's contributing to the greatest mismatch. In previous pull request: https://github.com/ghostery/ghostery-extension/pull/3131 which tried to adjust the manifest version 3 behaviour, we concluded that the change should be made in the converter side because we want to make the extension to be a platform that processes the given dNR rules in specific way, meaning the converter should be patched to include "main_frame" rather remove from the target resource types. To make things clear, it should be written as: manifest version 3 behavoiur differs from manifest version 2 behaviour. It's still "decision" how we do treat rules but I guess this pull request lacks the context. I rather think that redirect protection page should be "shown" rather hidden on those cases because one of its role is to prevent user from making any contact with malicious or high profiling website.

seia-soto:
Both uBlock Origin and AdGuard specify that `$document` is for targeting main frame but they have the difference in exception handling and default behaviour. uBlock Origin **does** include main frame by default. If adblocker library is intended to resemble their behaviour, the approach in this pull request is wrong. Also, an exception with `$document` in AdGuard side means "unblock everything" as per their documentation; see below.

> It is a type option (like image or script) that specifies the main frame (a.k.a. the root document) of a web page. This option is automatically enabled in filters indicating only the host part of the URL (see "HOSTS files" section), causing web pages that match the filter to get subjected to "Strict blocking".
>
> https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#document

> The rule corresponds to the main frame document requests, i.e. HTML documents that are loaded in the browser tab. It does not match iframes, there is a $subdocument modifier for these.
> 
> https://adguard.com/kb/general/ad-filtering/create-own-filters/#document-modifier

> @@||example.com^$document completely disables filtering on all pages at example.com and all subdomains.
>
> https://adguard.com/kb/general/ad-filtering/create-own-filters/#document-modifier

seia-soto:
refs https://github.com/ghostery/ghostery-extension/issues/3130

chrmod:
The problem is more nuanced. uBO, will block ghostery.com/private-search on filter `||ghostery.com^`, but wont on filter `||ghostery.com/private-search`. However it will block it on filter `||ghostery.com/private-search$document`

chrmod:
This PR brings platform parity and I believe it should be merge for that reason. 

However, we must discuss other aspects like DNR converter and uBO compatibility in separate threads.

philipp-classen (approved):

PR #3301, opened 2026-04-28
fix(preprocessors): align ext_ublock/ext_ubol envs

smalluban:
Align preprocessor env flags so extension identifiers map correctly to MV2/MV3 behavior.

- Set `ext_ublock` to `true`
- Set `ext_ubol` for MV3 only.

Review comment:
- https://github.com/ghostery/ghostery-extension/pull/3297#discussion_r3153841039

### DIFF
--- a/src/utils/preprocessors.js
+++ b/src/utils/preprocessors.js
@@ -22,16 +22,22 @@ function checkUserAgent(pattern) {
  */
 export const ENV = new Map([
   ['ext_ghostery', true],
-  ['ext_ublock', true],
-  ['env_mv3', __CHROMIUM__],
-  ['ext_ubol', __FIREFOX__],
-  ['cap_html_filtering', __FIREFOX__],
-  // TODO: Can be removed once $replace support is sufficiently distributed
-  ['cap_replace_modifier', __FIREFOX__],
-  ['cap_user_stylesheet', true],
+
+  // Browser related envs
   ['env_firefox', __FIREFOX__],
   ['env_chromium', __CHROMIUM__],
   ['env_edge', __CHROMIUM__ && isEdge()],
   ['env_safari', __CHROMIUM__ && isSafari()],
   ['env_mobile', checkUserAgent('Mobile')],
+
+  // Product related envs
+  ['env_mv3', __CHROMIUM__],
+  ['ext_ublock', __FIREFOX__],
+  ['ext_ubol', __CHROMIUM__],
+
+  // Capability related envs
+  ['cap_html_filtering', __FIREFOX__],
+  // TODO: Can be removed once $replace support is sufficiently distributed
+  ['cap_replace_modifier', __FIREFOX__],
+  ['cap_user_stylesheet', true],
 ]);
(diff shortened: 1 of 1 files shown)


src/utils/preprocessors.js
```diff
@@ -22,16 +22,22 @@ function checkUserAgent(pattern) {
...
   ['env_mobile', checkUserAgent('Mobile')],
+
+  // Product related envs
+  ['env_mv3', __CHROMIUM__],
+  ['ext_ublock', __FIREFOX__],
+  ['ext_ubol', __CHROMIUM__],
+
+  // Capability related envs
+  ['cap_html_filtering', __FIREFOX__],
+  // TODO: Can be removed once $replace support is sufficiently distributed
```
seia-soto, line 40: This TODO comment might be misleading because they'll be resolved as false by default.

seia-soto (approved):
I'd keep the order of the envs as in the uBlock Origin wiki page instead for quicker cross-checking in the future. Otherwise, looks good.

src/utils/preprocessors.js
```diff
@@ -22,16 +22,22 @@ function checkUserAgent(pattern) {
...
+  // Browser related envs
   ['env_firefox', __FIREFOX__],
   ['env_chromium', __CHROMIUM__],
   ['env_edge', __CHROMIUM__ && isEdge()],
   ['env_safari', __CHROMIUM__ && isSafari()],
   ['env_mobile', checkUserAgent('Mobile')],
+
+  // Product related envs
+  ['env_mv3', __CHROMIUM__],
+  ['ext_ublock', __FIREFOX__],
```
seia-soto: This has to be true, see https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#if-condition

seia-soto (changes requested):

src/utils/preprocessors.js
smalluban: This is not a change in this PR - only lines have changed.

src/utils/preprocessors.js
smalluban: It is `true` for Ublock. Should it be for MV3 chromium? @chrmod

src/utils/preprocessors.js
seia-soto: I suggest to remove that line as removing the entry will mislead in the future.

smalluban: Do you mean that `['cap_replace_modifier', __FIREFOX__],` must stay?

src/utils/preprocessors.js
chrmod: @seia-soto question is what this should say on Ghostery for Chrome

chrmod: @seia-soto is correct - this should be true

src/utils/preprocessors.js
seia-soto: My opinion is over the comment mostly but also yes to the comment. If somebody is loading old filter list and as long as the flag is listed somewhere, it would be a problem.

Okay with as-is.

seia-soto (approved):

PR #3302, opened 2026-04-28
fix(ci): properly filter always true conditions

seia-soto:
Correction on https://github.com/ghostery/ghostery-extension/pull/3297

Previous pull request was dropped condition without properly looking at operators like "!" and logical gates whenever all identifier tokens are unsupported. This creates invalid drops of conditions from the metadata file even when it's always false. This may lead to duplicated scripting or blocking issue on specific websites when the list or user custom filter has two filters in different adblocker syntax (one in ubo and another in adguard) then both gets accepted in any way (when they're correctly parsed).

We can extended this pull request to drop filter ids when the condition is always false later. But it's more risky step than having a filter turned on.

### DIFF
--- a/scripts/utils/filter-metadata.js
+++ b/scripts/utils/filter-metadata.js
@@ -17,40 +17,20 @@ globalThis.__CHROMIUM__ ??= false;
 
 const { ENV } = await import('../../src/utils/preprocessors.js');
 
-const SUPPORTED_ENVS = [...ENV.keys()];
-const IDENTIFIER_RE = /[a-zA-Z0-9_]+/g;
-
-function extractSupportedIdentifiers(expression) {
-  const ids = expression.match(IDENTIFIER_RE) || [];
-  return [...new Set(ids.filter((id) => SUPPORTED_ENVS.includes(id)))];
-}
-
-/**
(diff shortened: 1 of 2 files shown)


seia-soto:
I rather suggest closing this pull request and just reverting all changes in case if you're not sure about the results. This is clear advance in our product as it optimises the calls to the Chrome API but it introduces significant logical complexity while solving less than 100 entries, which takes less than 1ms on modern machine. It's better to test more reliably before shipping as there's reason to rush these changes.

smalluban:
Thanks @seia-soto, for spotting the issue and trying to fix it. However, your PR breaks the single source of truth for envs. We want to keep it in one place (in the source code). Your solution is also more aggressive (as the result is smaller, as you set, for example, `env_chromium` as always true, which for DNRs is correct, but it must be manually set ). It might be an advantage, but I intended to clean out metadata rules, where we use only those that are not supported at all.

The original PR can be simply improved to cover the case. I used your test file to prove it - I just had to add one rule, which is not cleared by my code, as it only filters out rules that are not in the list.

https://github.com/ghostery/ghostery-extension/pull/3303

PR #3303, opened 2026-04-29
fix(build): filter DNR metadata entries whose preprocessor is always true

smalluban:
## Summary

Fixes the DNR metadata filtering so it only drops entries whose `preprocessor` always evaluates to `true` across every combination of supported envs. Previously the logic also dropped entries whose result was constantly `false`, and kept entries that varied based on the env — the inverse of what we want.

Entries that are constantly `false` or that depend on supported envs are now preserved, since they still carry meaningful runtime information.

## Changes

- `scripts/utils/filter-metadata.js`: replace `dependsOnSupportedEnvs` with `alwaysTrueForSupportedEnvs`; `filterMetadata` now skips only entries that are unconditionally `true`.
- `tests/unit/scripts/download-dnr-rulesets.js`: add coverage for the new behavior.

## Note on #3302

[#3302](https://github.com/ghostery/ghostery-extension/pull/3302) attempts to fix the same issue, but it does so by hardcoding the list of supported envs inside the build script. That breaks the single source of truth in [src/utils/preprocessors.js](src/utils/preprocessors.js), where the `ENV` map is defined:

> This map is the source of truth for all supported preprocessor envs.

This PR keeps that invariant — `scripts/utils/filter-metadata.js` continues to derive `SUPPORTED_ENVS` from `ENV.keys()`:

```js
const { ENV } = await import('../../src/utils/preprocessors.js');
const SUPPORTED_ENVS = [...ENV.keys()];
```

so any future change to the supported env list only needs to happen in one place.

### DIFF
--- a/scripts/utils/filter-metadata.js
+++ b/scripts/utils/filter-metadata.js
@@ -26,38 +26,35 @@ function extractSupportedIdentifiers(expression) {
 }
 
 /**
- * Returns true if the preprocessor expression can yield different results
- * depending on the value of supported envs. If the result is constant
- * regardless of the environment (always true or always false), returns false.
+ * Returns true if the preprocessor expression always evaluates to `true`
+ * across every combination of supported envs. Expressions that depend on
+ * supported envs (yielding both true and false) or that are constantly
+ * false return `false`.
  */
-function dependsOnSupportedEnvs(expression) {
+function alwaysTrueForSupportedEnvs(expression) {
   const ids = extractSupportedIdentifiers(expression);
   if (ids.length === 0) {
-    return false;
+    return evaluatePreprocessor(expression, new Map()) === true;
   }
 
   const combinations = 1 << ids.length;
-  let firstResult;
   for (let mask = 0; mask < combinations; mask++) {
     const env = new Map();
     ids.forEach((key, i) => {
       env.set(key, !!(mask & (1 << i)));
     });
-    const result = evaluatePreprocessor(expression, env);
-    if (mask === 0) {
-      firstResult = result;
-    } else if (result !== firstResult) {
-      return true;
+    if (evaluatePreprocessor(expression, env) !== true) {
+      return false;
     }
   }
-  return false;
+  return true;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tests/unit/scripts/download-dnr-rulesets.js
```diff
@@ -0,0 +1,40 @@
...
+      notAdguardOrSafari: { preprocessor: '!adguard||ext_safari' },
+      adguardOrSafari: { preprocessor: 'adguard||ext_safari' },
+      chromiumAndNotChromium: { preprocessor: 'env_chromium&&!env_chromium' },
+      chromiumOrNotChromium: { preprocessor: 'env_chromium||!env_chromium' },
+      chromiumAndGhostery: { preprocessor: 'env_chromium&&ext_ghostery' },
+      notEdge: { preprocessor: '!env_edge' },
+      notAdguardAndNotEdge: { preprocessor: '!adguard&&!env_edge' },
+    };
+
+    assert.deepEqual(filterMetadata(metadata), {
```
chrmod: I don't understand this assertion, it is bit too meta. Can we split it into many assertions that each has assertion message that describes what it does?

tests/unit/scripts/download-dnr-rulesets.js
```diff
@@ -0,0 +1,40 @@
...
+
+import { filterMetadata } from '../../../scripts/utils/filter-metadata.js';
+
+describe('Download DNR Rulesets', () => {
+  it('filterMetadata', () => {
+    const metadata = {
+      adguard: { preprocessor: 'adguard' },
+      notAdguard: { preprocessor: '!adguard' },
+      notAdguardOrSafari: { preprocessor: '!adguard||ext_safari' },
+      adguardOrSafari: { preprocessor: 'adguard||ext_safari' },
```
chrmod: how this setup deals with ext_safari which should be derived from UA so in the runtime?

chrmod (changes requested):

chrmod (approved):

PR #3313, opened 2026-05-04
fix(e2e): poll subdocument scriptlet result to avoid race condition

smalluban:
On Chromium, scriptlets are injected via `chrome.scripting.executeScript` triggered asynchronously from `webNavigation.onCommitted` in the background service worker. For a tiny document like `subdocument.html`, the parent frame's `load` event can fire before the scriptlet has run, causing a single synchronous check of `iframe.contentWindow.subdocument` to always return `false`.

The fix replaces the one-shot check with a polling loop (up to 10 retries × 50 ms = 500 ms) that waits for the scriptlet to set the property before concluding the test. The parent-frame guard (`window.subdocument !== true`) is preserved to ensure the subframe constraint (`>>`) is still validated correctly.

### DIFF
--- a/tests/e2e/page/adblocker/index.html
+++ b/tests/e2e/page/adblocker/index.html
@@ -137,13 +137,23 @@
           l++;
           const iframe = document.createElement("iframe");
           iframe.addEventListener("load", function () {
-            if (
-              iframe.contentWindow.subdocument === true
-              && window.subdocument !== true
-            ) {
-              results["subdocument"] = true;
+            // On Chromium, scriptlet injection via chrome.scripting.executeScript
+            // is triggered asynchronously from webNavigation.onCommitted in the
+            // background service worker. For a tiny document like subdocument.html,
+            // the load event in the parent can fire before the scriptlet runs.
+            // Poll with retries to account for this race condition.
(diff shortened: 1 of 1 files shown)


seia-soto (changes requested):
I think polling the result breaks the intent of the test. I suggest reusing the iframe element for the later testing (saving to the `window.subdocument_iframe` or somewhere; then read in specific timing - 100ms, 500ms, 1000ms). I can make a change tomorrow myself if it's not urgent.

smalluban:
This pulls the existence of the result. If scriptlets don't run, the result is still the same, as then tests will fail. 

The PR fixes flaky tests runs on CI, as lately 50% of runs on Chrome fails on the test with checking scriptlets from this suite.

seia-soto:
The timing exists to see if the scripting or injections worked in specific timeframes if the earliest didn’t work. We shouldn’t take injection happening too late as a success since it means the ad-blocking will fail with a timing issue.

I can rather say the existing test is wrong because it injects new frame in every test not just checking the value just like other tests.

smalluban:
Ok. If you think you can create a more reliable fix, we can wait for it.

PR #3317, opened 2026-05-04
fix(e2e): check value in subdocument at specific timeframe

seia-soto:
continues https://github.com/ghostery/ghostery-extension/pull/3313


seia-soto:
The root cause was the subdocument testing having invalid async operation and affecting the whole test condition, which prevents the strict condition of "suite.collection.expected !== suite.collection.reports.length" in the wdio test file.

smalluban (approved):

PR #3329, opened 2026-05-08
chore(github): add create-pr prompt for contributors

smalluban:
Contributors working in this repo have no shared Copilot prompt for opening pull requests, leading to inconsistent PR titles, missing descriptions, and manually typed commit patterns.

Adds a `.github/prompts/create-pr.prompt.md` prompt that contributors can invoke as `/create-pr` in Copilot chat. It guides the agent through checking the branch state, committing any pending changes with the correct conventional-commit pattern, pushing, and creating a well-structured PR with a motivation summary and change description.

### DIFF
--- a/.github/prompts/create-pr.prompt.md
+++ b/.github/prompts/create-pr.prompt.md
@@ -1,0 +1,26 @@
+---
+description: 'Create a GitHub Pull Request from the current branch. Use when: opening a PR, creating a pull request, submitting code for review.'
+argument-hint: 'Optionally specify a title, base branch, or draft'
+---
+
+Create a GitHub Pull Request for the current branch following these steps:
+
+1. **Check current branch**: If on `main`, ask the user for a branch name and create it before proceeding (`git checkout -b <branch-name>`).
+
+2. **Check for uncommitted changes**: If there are staged or unstaged changes, help the user commit them. Use the commit message pattern:
+   - Bug fix: `fix(MODULE_NAME): DESCRIPTION`
+   - New feature: `feat(MODULE_NAME): DESCRIPTION`
+   - Chore (scripts, dependencies, tooling — no extension source changes): `chore(MODULE_NAME): DESCRIPTION`
+     Ask the user what type of change it is, and what the module name and description are.
+
+3. **Check for unpushed commits**: Push the branch to the remote if needed (`git push` or `git push --set-upstream origin <branch>`).
+
+4. **Prepare PR details**:
+   - **Title**: Derive from the branch name and recent commits if not provided.
+   - **Body**: Write a structured description in two parts (no headers, plain prose):
+     1. (1–2 sentences): Explain the motivation or problem that prompted this change. Use git history, recent session context, or file change context to infer it — look at related commits, issue references, or prior conversation context about the changed files.
+     2. (2–3 sentences): Describe the changes applied — what was added, fixed, or refactored and how it addresses the need.
+
+5. **Create the PR** using the `github-pull-request_create_pull_request` tool.
+
+6. **Report the result**: Share the PR number and URL as a markdown link, and mention the base branch it targets.


seia-soto (commented):
First of all, I lack some understandings with copilot features as I don't use their features and I use vscodium fork instead of vscode. I probably haven't seen this feature. From my own searches, this prompt path is directly read by ide; vscode, vs, jetbrains - https://docs.github.com/en/copilot/tutorials/customization-library/prompt-files/your-first-prompt-file and identical to `AGENTS.md` or `CLAUDE.md` files but just in different path.

Unfortunately, the AI tooling that even our team uses are pretty much scattered and I don't expect we can have a single file for every tooling. The most standard `AGENTS.md` goes with almost everything except for Claude Code but it seems like also supported by copilot: https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-an-agentsmd-file. `CLAUDE.md` is also recognised by many tools: https://opencode.ai/docs/rules/#precedence and https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-a-claudemd-file.

What I suggest is to make `CLAUDE.md` or `AGENTS.md` as an entrypoint then expand the paths from there. If the model and prompting are smart enough to understand the structure, that's mostly enough.

Prompt-wise, I don't think we need to include step 1, 2, 3, 5, and 6. If we can just derive those functionality as a separate script, it will reduce the token usage and hallucination. Despite I don't like git hooks due to complexity it might bring with dependencies, it can still do the same thing. Then AI can use it as a tool (`gh` cli is already widely used for AIs afaik).

My personal setup (opencode) is pretty far from this and `.github/prompts` path is too unique for the community. Also, I guess that steps in prior of generating pr details should be caught by automated code-reviews or other systems.

philipp-classen:
I'm also not using Copilot myself, so I cannot provide good feedback. I would not mind to merge anything in .github/prompts though.

Reading AGENTS.md, I'm always a bit hesitant to put too much information there, since it by design it will always be read. For some tasks, this may be even harmful, since the extra information may confuse the model.

Btw, regarding Claude not following conventions, a hack is to do it like llama.cpp and just tell it to read the AGENTS.md file:
https://github.com/ggml-org/llama.cpp/blob/master/CLAUDE.md

smalluban:
We can use a more general location. I expect this one is also read by many model extensions for editors, or their own applications with agent mode. 

This file is a general instruction for generating PR. It is pointless to create a custom script for it, as the model-based prompt is more generic - it will scan commits, context, your sessions, etc to detect what has been done, etc. On that basis, it can also ask additional questions. 

I found it useful, as it follows the patterns we use - on commit message, branch name, PR title, and description. Do you have some thoughts about the prompt content? Do you want to have this process differently? It is not mandatory and can be used if you need it or like it.

Also, it is not meant for the community, but for the contributors. If you don't find it useful, I can still happily use my locally based prompts and close this PR.

seia-soto:
Making a custom script is not pointless direction because checking even the git status is too much delegation to AI. What I'm rather saying is that make a simple script that will simply fail and let AI to run it instead. This will decrease the token usage in the longer term significantly. Giving a proper tool to AI instead of delegating everything is really important imo. Building such script is not even a complicated thing, or instruct to use specific command that emits simple result. It will make the whole process much more simpler and lead to the faster thinking time.

seia-soto:
I'm leaning same to @philipp-classen unless we have some policies over AI tooling (despite I guess that having a policy is very hard in the current situation). If it works well for you, you can just keep it and use it as you want.

If we see it as an individual IDE/editor configuration, I would just recommend to keep it local or store somewhere else rather than the repository. Unless it's required and recommended setup, usually those content never goes to the repository (just like .vscode dir).

smalluban:
I understand your point of view. I don't want to exploit the discussion. I will close the PR and keep the prompts local.

PR #3330, opened 2026-05-08
Update @ghostery/adblocker packages to 2.17.0

seia-soto:
There's no API changes, but deserialisation performance improvement with different crc32 method.

refs https://github.com/ghostery/adblocker/pull/5648
refs https://github.com/ghostery/ghostery-extension/issues/3175


seia-soto:
v2.17.0 introduces binary level merge and this requires a separate implementation. There is no backward-incompatible change.

smalluban (approved):

PR #3345, opened 2026-05-13
ci(release): build and upload automation extension zips

chrmod:
Adds automation variants of all three extension zips (`ghostery-automation-{firefox,chromium,safari}-<version>.zip`) for use with Selenium/Playwright. Skips the onboarding gate so test drivers can interact with the extension on first load, and identifies as "Ghostery (Automation)" in `chrome://extensions` to avoid confusion with production builds.

Stacked on #3341 — merge that first.


smalluban (approved):
LGTM, but I don't fully understand where we need it.

scripts/build.js
```diff
@@ -56,9 +56,32 @@ const argv = process.argv.slice(2).reduce(
     silent: false,
     clean: false,
     watch: false,
+    automation: false,
   },
 );
 
+// --- Automation patch ---
+function applyAutomationPatch(distDir) {
```
smalluban, line 64: What do you think about the env variable `__AUTOMATION__` and then set `disableOnboarding` directly in the code? The current solution should be resilient to code changes, but it is still a string replacement method, and it might fail silently.

chrmod: Not sure what's best yet. I'm still experimenting with the setup.

PR #3350, opened 2026-05-15
feat(distractions): add feature with basic options

### LINKED ISSUE #2356: Block annoyances
A meta issue to track all users request to block different kinds of annoyances.

smalluban:
Fixes https://github.com/ghostery/ghostery-extension/issues/2356

Adds a new "Distractions" feature so users can hide intrusive on-page prompts (such as Google Sign-In One Tap and Reddit's mobile "open in app" interstitial) without having to write custom filters.

Introduces a dedicated `distractions` settings view with per-source toggles, a new `distractions` engine that compiles the selected filters via `parseFilters` and applies them through `reloadMainEngine` (plus dynamic DNR rules on Chromium), and an `onInstalled`/options-observer pipeline that rebuilds the engine only when the bundled filters change (tracked via an `xxh32` checksum stored on the engine). Adds an e2e spec that verifies the Google Sign-In client script is loaded when the toggle is off and blocked when it is on.

Zrzut ekranu 2026-05-21 o 13 58 05


src/store/options.js
```diff
@@ -67,6 +67,12 @@ const Options = {
     trustedScriptlets: false,
   },
 
+  // Distractions
+  distractions: {
+    google: false,
```
chrmod: wonder if that structure is best - what if we will have hundreds of domains here?

smalluban: You're right, i'll move to `store.record()` - the structure will be the same, but it will support any key.

mdsandu (changes requested):
Search box: update copy to "Search distractions or websites..."

We should make descriptions answer the question: What distraction will disappear? and normalize wording across all rows with Verb + thing + benefit, and keep some naming consistency 

1. Sign in with Google
Hide Google sign in prompts across the web.
Works on: google.com

Short video feeds
Hide endless short video feeds that can pull you into distraction loops.
Works on: youtube.com, instagram.com, facebook.com

Browser promotions
Hide browser prompts that interrupt your browsing.
Works on: [domain names comma-separated]

Social media clutter
Hide social widgets and sharing panels that add noise to pages.
Works on: [domain names comma-separated]

mdsandu:
based on the mistake in 1, change to

Sign in with Google
Hide Google sign in prompts across web.
Applies to: Websites with Google sign in

exchange "Works on:" to "Applies to:" everywhere

chrmod (approved):

mdsandu:
@smalluban Applies to: **W**ebsites with Google sign in - W in Websites is uppercase here

PR #3358, opened 2026-05-21
build: add --automation flag for automation harnesses

chrmod:


chrmod:
Superseded by #3345 — moving the `--automation` build flag onto that branch since it covers the same ground.

PR #3368, opened 2026-05-23
feat(managed): cache in local storage

### LINKED ISSUE #3367: Ad-blocking hangs with Chrome Policy Service initialisation
Ghostery extension hangs due to Chrome Policy Service initialization. Policy Service is a subsystem of Chromium for Enterprise customers but it doesn't make sense to hold the non-Enterprise users.

From https://github.com/ghostery/ghostery-extension/issues/3175#issuecomment-4518263567, I could confirm `chrome.storage.managed.get` takes a few seconds (3-8 seconds range for me; Apple M1).

```
diff --git a/src/store/managed-config.js b/src/store/managed-config.js
index c6441a6b..3111c84e 100644
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -38,6 +38,7 @@ const ManagedConfig = {
     if (__CHROMIUM__ && (isOpera() || isWebkit())) return {};

     try {
+      console.time('managedConfig-storeConnect-chrome.storage.managed');
       // Firefox uses filesystem configuration on every platform
       // and throws if storage.managed is not available,
       // so we can skip fallback for it
@@ -46,7 +47,9 @@ const ManagedConfig = {
         if (__DEBUG__) return {};
         throw e;
       });
+      console.timeEnd('managedConfig-storeConnect-chrome.storage.managed');
```

Image

Also, the similar log could be found from `chrome://policy/logs` (or see https://github.com/ghostery/ghostery-extension/issues/3175#issuecomment-4522929548). The first log comes at 14:09:22 PDT but the actual initialisation complete log appears after ~6 seconds.

```
2026-05-22, 14:09:22 PDT	VERBOSE	chrome_browser_cloud_management_controller.cc:121	DM token = empty
2026-05-22, 14:09:22 PDT	VERBOSE	chrome_browser_cloud_management_controller.cc:127	Enrollment token =
...
2026-05-22, 14:09:28 PDT	VERBOSE	policy_service_impl.cc:568	PolicyService is initialized for domain: 1
2026-05-22, 14:09:28 PDT	VERBOSE	policy_service_impl.cc:568	PolicyService is initialized for domain: 3
...
```

One thing unconfirmed but observed and assumed is that if extension calls `chrome.storage.managed.get` before its initialisation, it may slow down the process heavily: 3s to ~6s. Not sure if this is just because of the system load.

seia-soto:
fixes #3367

It modifies managed config initialisation process not to be held by `chrome.storage.managed.get` call, which takes more than 6 seconds depends on the environment. Despite it only happens on fresh start of Google Chrome, 6 seconds is enough for most of users to start browsing and it shouldn't be the case for users without managed policy.

Exact reason for Chrome's policy service hanging so long in the initialisation is not known and it doesn't seem to be controllable from the extension level.

As requested in https://github.com/ghostery/ghostery-extension/pull/3368#pullrequestreview-4361983635, this version doesn't support options update on the fly. It still tries to receive data from `chrome.storage.managed.onChanged` as having up-to-date data is important.

<details>

<summary>Log screenshot when Chrome Policy Service is already initialised</summary>

image

</details>

<details>

<summary>Log screenshot when Chrome Policy Service is not ready</summary>

**Make sure Google Chrome is fully shut downed before testing this**

image

</details>

### DIFF
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -37,42 +37,43 @@ const ManagedConfig = {
   [store.connect]: async () => {
     if (__CHROMIUM__ && (isOpera() || isWebkit())) return {};
 
-    try {
-      // Try to get cached local version to mitigate slow Chrome
-      // managed storage initialization.
-      // We endure and expect the stale cache from local storage cache
-      // as service worker will restart frequently.
-      let { managedConfig } = await chrome.storage.local.get('managedConfig');
+    // Try to get cached local version to mitigate slow Chrome
+    // managed storage initialization.
+    // We endure and expect the stale cache from local storage cache
+    // as service worker will restart frequently.
+    let { managedConfig } = await chrome.storage.local.get('managedConfig');
 
-      // We will skip managed config in debug mode
-      // Also this local storage overriding in e2e tests
-      if (__DEBUG__) {
-        managedConfig ??= {};
-      } else {
-        // Firefox throws if storage.managed is not found unlike Chrome
-        // returning empty object
-        const managedConfigFromBackend = chrome.storage.managed
-          .get()
-          .then(function (managedConfig) {
-            chrome.storage.local.set({ managedConfig });
-            return managedConfig;
-          });
+    // We will skip managed config in debug mode
+    // Also this local storage overriding in e2e tests
+    if (__DEBUG__) {
+      managedConfig ??= {};
+    } else {
+      // Firefox throws if storage.managed is not found unlike Chrome
+      // returning empty object
+      const managedConfigFromBackend = chrome.storage.managed
+        .get()
+        .then(function (managedConfig) {
+          chrome.storage.local.set({ managedConfig });
+          return managedConfig;
+        })
+        .catch(function (e) {
+          // `e.message` equals `Managed storage manifest not found` here
+          if (__FIREFOX__ && e?.message.includes('not found')) return {};
+          throw e;
+        });
 
-        managedConfig ??= await managedConfigFromBackend;
-      }
-
-      // Translate `customFilters` storage.managed key (an array) to model structure
-      if (managedConfig.customFilters) {
-        managedConfig.customFilters = {
-          enabled: true,
-          filters: managedConfig.customFilters,
-        };
-      }
-
-      return managedConfig;
-    } catch {
-      return {};
+      managedConfig ??= await managedConfigFromBackend;
     }
+
+    // Translate `customFilters` storage.managed key (an array) to model structure
+    if (managedConfig.customFilters) {
+      managedConfig.customFilters = {
+        enabled: true,
+        filters: managedConfig.customFilters,
+      };
+    }
+
+    return managedConfig;
   },
 };
 
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -56,10 +56,10 @@ const ManagedConfig = {
           chrome.storage.local.set({ managedConfig });
           return managedConfig;
         })
-        .catch(function (e) {
-          // `e.message` equals `Managed storage manifest not found` here
-          if (__FIREFOX__ && e?.message.includes('not found')) return {};
-          throw e;
+        .catch(function () {
+          // We intend to catch all errors and make the function safe
+          // Returning empty object here will handle every case
+          return {};
         });
 
       managedConfig ??= await managedConfigFromBackend;
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -45,18 +45,11 @@ const ManagedConfig = {
       let { managedConfig } = await chrome.storage.local.get('managedConfig');
 
       // Firefox throws if storage.managed is not available unlike Chrome
-      let managedConfigFromBackend = chrome.storage.managed
-        .get()
-        .catch(function (e) {
-          // Allow fallback in debug mode (for e2e tests)
-          if (__DEBUG__) return {};
-          throw e;
-        })
-        .then(function (managedConfig) {
-          // Prevent local storage overriding in debug mode (for e2e tests)
-          if (!__DEBUG__) chrome.storage.local.set({ managedConfig });
-          return managedConfig;
-        });
+      let managedConfigFromBackend = chrome.storage.managed.get().then(function (managedConfig) {
+        // Prevent local storage overriding in debug mode (for e2e tests)
+        if (!__DEBUG__) chrome.storage.local.set({ managedConfig });
+        return managedConfig;
+      });
 
       managedConfig ??= await managedConfigFromBackend;
 
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -38,41 +38,36 @@ const ManagedConfig = {
     if (__CHROMIUM__ && (isOpera() || isWebkit())) return {};
 
     try {
-      // Firefox uses filesystem configuration on every platform
-      // and throws if storage.managed is not available,
-      // so we can skip fallback for it
-      let managedConfig = await chrome.storage.managed.get().catch((e) => {
-        // Allow fallback in debug mode (for e2e tests)
-        if (__DEBUG__) return {};
-        throw e;
+      // Try to get cached local version to mitigate slow Chrome
+      // managed storage initialization.
+      // We endure and expect the stale cache from local storage cache
+      // as service worker will restart frequently.
+      let { managedConfig } = await chrome.storage.local.get('managedConfig');
+
+      // Allow fallback in debug mode (for e2e tests)
+      if (__DEBUG__) {
+        return managedConfig || {};
+      }
+
+      // Firefox throws if storage.managed is not available unlike Chrome
+      let managedConfigFromBackend = chrome.storage.managed.get().then(function (managedConfig) {
+        chrome.storage.local.set({ managedConfig });
+        return managedConfig;
       });
 
-      // Chromium-based browsers just return an empty object,
-      // and it might be available later (especially on the browser restart),
-      // so we try to read it with fallback from local storage
-      if (__CHROMIUM__ || __DEBUG__) {
-        if (Object.keys(managedConfig).length) {
-          // Save local version for fallback usage
-          // Don't await - we don't need to block on this
-          chrome.storage.local.set({ managedConfig });
-        } else {
-          // Try to get local version as fallback
-          const { managedConfig: fallbackConfig = {} } =
-            await chrome.storage.local.get('managedConfig');
+      if (!managedConfig) {
+        managedConfig = await managedConfigFromBackend;
 
-          managedConfig = fallbackConfig;
+        // Translate `customFilters` storage.managed key (an array) to model structure
+        if (managedConfig.customFilters) {
+          managedConfig.customFilters = {
+            enabled: true,
+            filters: managedConfig.customFilters,
+          };
         }
       }
 
-      // Translate `customFilters` storage.managed key (an array) to model structure
-      if (managedConfig.customFilters) {
-        managedConfig.customFilters = {
-          enabled: true,
-          filters: managedConfig.customFilters,
-        };
-      }
-
-      return managedConfig || {};
+      return managedConfig;
     } catch {
       return {};
     }
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -44,14 +44,22 @@ const ManagedConfig = {
       // as service worker will restart frequently.
       let { managedConfig } = await chrome.storage.local.get('managedConfig');
 
-      // Firefox throws if storage.managed is not available unlike Chrome
-      let managedConfigFromBackend = chrome.storage.managed.get().then(function (managedConfig) {
-        // Prevent local storage overriding in debug mode (for e2e tests)
-        if (!__DEBUG__) chrome.storage.local.set({ managedConfig });
-        return managedConfig;
-      });
+      // We will skip managed config in debug mode
+      // Also this local storage overriding in e2e tests
+      if (__DEBUG__) {
+        managedConfig ??= {};
+      } else {
+        // Firefox throws if storage.managed is not found unlike Chrome
+        // returning empty object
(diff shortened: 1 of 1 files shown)


smalluban (changes requested):
The cold start issue is real and worth solving, but I have a few thoughts:

Separate file: I'd prefer to keep the model's connection logic in managed-config.js (store) — the circular reference concern doesn't justify the added indirection.

Simpler approach: The adaptor abstraction is over-engineered. We can flip priority in the existing [store.connect]: read managedConfig from chrome.storage.local first, return it immediately if available, then fetch from chrome.storage.managed in the background and update the cache.

Critical issue of your proposal: On first cold start (fresh install or first run after upgrading), managedConfigCache is empty, so the adaptor resolves with {}. A managed user with disableOnboarding: true will see onboarding — the real config arrives too late. The current code avoids this by blocking on chrome.storage.managed.get(). Any replacement must preserve that safety guarantee: if managedConfig exists in local storage, it must resolve before onboarding/settings checks run.

seia-soto:
The critical flow is now fixed with E2E exception of skipping local storage overriding. Also, I stopped deriving "managedConfigCache" item in local storage as now I get what "fallback config from local storage" means. I thought that it was actually used to apply managed config via local storage but it's mostly used for fallback when the managed storage fails. This is what I understood and inferred from the comment but I'm not fully sure.

Especially I couldn't understand the exception under previous `chrome.storage.managed.get().catch` block. I need more context about it because there's no way it can fail under E2E testing afaik. The current changes drop those exception handling.

```
      let managedConfig = await chrome.storage.managed.get().catch((e) => {
        // Allow fallback in debug mode (for e2e tests)
        if (__DEBUG__) return {};
        throw e;
      });
```

Regarding renaming of `manage` function into `applyManagedConfigToOptions`, I would like to keep this change despite there's no reason to change. It's confusing to call function just with name "manage".

Other than those, the function now retrieves local storage then managed storage with a help of `populateManagedConfigCache`, which is a function taking optional `config` (from managed storage). The `populateManagedConfigCache` function will perform managed storage lookup (unless given) then populate the cache. Runtime managed value update is done by a separate function called `lazilyReflectManagedConfigToOptions` and this is only called from local storage cache path.

I would like to know what's the reliable way to "update" the options from new managed config with Hybrid. From the HybridJS docs, it looks like I should use "store.sync" rather than "store.set" because I don't want to propagate values to the options storage. However, circular referencing models are quite confusing to understand without a solid understanding, especially about lifecycle. I saw the use of `await store.resolve(Options)` instead of `chrome.storage.local.get(['options'])` would resolve the risk of modifying options before initialisation via store.connect method in Options (which includes migration). The previous avoidance of circular referencing was to reduce the metal load of understanding lifecycle.

seia-soto:
Also, it might be helpful to have more logging points when we try to debug Enterprise customers with limited access despite the spec is quite standard under browser. I guess the current log points provide enough information for debugging but not for everyone.

smalluban (changes requested):
We don't need to update on the fly options. If we skip one SW runtime it's ok, as managed configuration doesn't change often. The critical is to wait for the first fetch if there is nothing in the cache. I think your solution is still too much complicated. It can fit inside of the connect method of the model. We just need to reverse the order and update local storage with data from managed when fetched.

seia-soto:
I changed the code flow as mentioned. However, I still want to make the code to have access up-to-date managed config as soon as possible even if it's not going to change frequently. Letting users restarting browser (exactly extension service worker here) twice to load the managed config properly sounds like just another bug for me, or confusion at least. I believe using `chrome.storage.managed.onChange` would solve the issue better.

src/store/managed-config.js
```diff
@@ -34,49 +71,58 @@ const ManagedConfig = {
...
-          // Don't await - we don't need to block on this
-          chrome.storage.local.set({ managedConfig });
-        } else {
-          // Try to get local version as fallback
-          const { managedConfig: fallbackConfig = {} } =
-            await chrome.storage.local.get('managedConfig');
-
-          managedConfig = fallbackConfig;
-        }
+  [store.connect]: !isManagedStorageAvailable
```
smalluban: This is microoptimization, but makes logic more complicated.

src/store/managed-config.js
```diff
@@ -15,6 +15,43 @@ import { isOpera, isWebkit } from '/utils/browser-info.js';
...
+      enabled: true,
+      filters: config.customFilters,
+    };
+  }
+
+  return config;
+}
+
+if (isManagedStorageAvailable) {
+  chrome.storage.managed.onChanged.addListener(function () {
```
smalluban: Don't use it - there is no guarantee that this API works as expected. We are ok with updating the managed state if it is cached already with another restart of SW.

src/store/managed-config.js
```diff
@@ -15,6 +15,43 @@ import { isOpera, isWebkit } from '/utils/browser-info.js';
 
 export const TRUSTED_DOMAINS_NONE_ID = '<none>';
 
+const MANAGED_CONFIG_STORAGE_ID = 'managedConfig';
+
+const isManagedStorageAvailable = __FIREFOX__ || (!isOpera() && !isWebkit());
+
+async function populateManagedConfigCache(config) {
```
smalluban: This is a simple function - we can avoid splitting logic into separate functions.

src/store/options.js
```diff
@@ -135,7 +135,7 @@ const Options = {
 
       // Apply managed options for supported platforms
       if (__FIREFOX__ || (!isSafari() && !isOpera())) {
-        return manage(options);
+        return applyManagedConfigToOptions(options, await store.resolve(ManagedConfig));
```
smalluban: I don't see a need to rename the function. The function should contain all of the required logic inside - don't pass config as a parameter here.

As you see, the managed config is only applied to options for supported platforms.

smalluban (changes requested):
Please focus ONLY on appliing the logic change to `connect` method of the managed config - to first fetch from localStorage, and also fetch (without await) from storage.managed, which should update localStorage. 

The whole PR should be more or less 25-50 lines changed.

seia-soto:
@smalluban Reflected last comment. The update was heavily delayed with taxation.

One thing would be I don't understand this comment. Does FF throw an error whenever there's an "empty" config or "no" config?

```
      // Firefox uses filesystem configuration on every platform
      // and throws if storage.managed is not available,
      // so we can skip fallback for it
```

smalluban (approved):
Now it is beautiful.

smalluban (changes requested):
You have changed the logic suddenly - in `__DEBUG__` on Firefox, the managed storage should be read as well. 

Please keep the logic as close to how it was before as possible.

seia-soto:
@smalluban I still don't understand this catch block, so I wouldn't know what's the accurate way. Please, check the comments in the below code block.

```ts
      // Firefox throws if storage.managed is not available unlike Chrome
      let managedConfigFromBackend = chrome.storage.managed
        .get()
        .catch(function (e) {
          // Allow fallback in debug mode (for e2e tests) <-- Does this actually help? What's the trigger if it should be called on FF?
          if (__DEBUG__) {
            return {};
          }
          throw e;
        })
        .then(function (managedConfig) {
          // Prevent local storage overriding in debug mode (for e2e tests) <-- This is required in both ways
          if (!__DEBUG__) {
            chrome.storage.local.set({ managedConfig });
          }
          return managedConfig;
        });
```

smalluban:
As we fetch from cache first, the previous condition with `__DEBUG__` you have already covered. However, your last commit breaks the logic on Firefox, as when running `npm run debug` it never tries to fetch from managed storage. Just remove the `catch()` block altogether. If Firefox provides the managed storage, it should be used.

seia-soto:
I get that as FF shouldn't be throwing when there's no managed config, which actually makes sense. I wondered what was that for. The new change dropped catch block and passing `npm run wdio -- --clean --target=firefox --spec ./tests/e2e/spec/managed.spec.js` from local machine. Now waiting for CI.

src/store/managed-config.js
```diff
@@ -38,31 +38,20 @@ const ManagedConfig = {
...
-          // Don't await - we don't need to block on this
-          chrome.storage.local.set({ managedConfig });
-        } else {
-          // Try to get local version as fallback
-          const { managedConfig: fallbackConfig = {} } =
-            await chrome.storage.local.get('managedConfig');
+      // Firefox throws if storage.managed is not available unlike Chrome
+      let managedConfigFromBackend = chrome.storage.managed.get().then(function (managedConfig) {
+        // Prevent local storage overriding in debug mode (for e2e tests)
+        if (!__DEBUG__) chrome.storage.local.set({ managedConfig });
```
smalluban: Previously, there was a condition `if (Object.keys(managedConfig).length) {` - your current code still breaks the logic, this time for Chrome.

smalluban (changes requested):

src/store/managed-config.js
seia-soto: It doesn't matter in the new code because previous code under that condition was: `chrome.storage.local.set({ managedConfig });` and it's now handled by `if (!__DEBUG__) chrome.storage.local.set({ managedConfig });`. Also, it should NOT write to `chrome.storage.local.set` under debug mode (especially about e2e tests) since e2e will write their config into local storage.

I don't get the previous comment in the source nor clear flow there. Please, suggest the changes directly or let me know why that's required.

seia-soto: What I can assume is that `Object.keys(managedConfig).length` check is to validate if the Chrome response "actually" has a value (not empty object) but it shouldn't be handled in that way with this approach. If we don't save empty object as a cache, it makes this pull request meaningless as it will always hit `chrome.storage.managed.get`.

seia-soto: What I only rely on while doing this is mostly e2e tests because I assume `chrome.managed.storage.get` and `chrome.managed.local.get` to work just fine (despite I saw some comments related to the methods).

I don't understand why it's broken because it seems like a non-used or invalid code path from my perspective. It's extremely difficult what should I do without knowing the method to validate.

smalluban: On Chrome, before the change, the local version was set only if the managedConfig did not have keys. 
Now, in your logic, the local storage is set regardless of the value of the `managedConfig` - it just must be resolved. 
You also removed the comment about Chrome, that if the managed mode is not supported, it returns an empty object, rather than throws an error (as Firefox does).
In `then()` callback, you should check if the `managedConfig` has keys, not checking `__DEBUG__` - the previous logic did not do that.

seia-soto: If we apply that, we're making this pull request meaningless. Assuming that Chrome will not throw and just return an empty object from `chrome.storage.managed.get`, we will be skipping the cache population whenever it's not available or empty (as we cannot distinguish). That means it will always fallback to `chrome.storage.managed.get` which might hang.

I'm assuming "not available" and "empty" are different here. If it's not, please correct it. Also, this can be skipped as service worker restarts frequently and the latest cache will always be populated lazily. What I can do instead is that making it not to clear out the existing cache. I guess that it's what you've meant.

seia-soto: I found this issue that clarifies the behaviour: https://github.com/w3c/webextensions/issues/497. According to the issue, it looks like "not available" in the comment includes the situation when the managed config is "empty".

Therefore the behaviour would be:

**Chrome**

- If N/A, throws
- If not found, `{}`
- If empty, `{}`

**Firefox**

- If N/A, throws
- If not found, throws
- If empty, probably `{}` (when there's a configuration but no actual value)

smalluban: Ok, we can change the logic, so in `__DEBUG__` we skip fetching from managed storage at all, regardless of the platform, so the model only tries to load from local storage.

seia-soto: If `__DEBUG__` is only for e2e, correct. I wasn't aware of `npm run debug` tho: https://github.com/ghostery/ghostery-extension/pull/3368#issuecomment-4555167812

smalluban: I wasn't, but from now on, running `npm run debug` will skip the managed config from the browser. It is a trade-off, but as you correctly pointed out, it is required for this PR to provide any gain in Chrome.

seia-soto: I updated the code to fully drop managed config in debug mode. Under debug mode (or e2e tests), the extension will always rely on local storage.

smalluban (changes requested):

src/store/managed-config.js
```diff
@@ -38,30 +38,27 @@ const ManagedConfig = {
...
+        managedConfig ??= {};
+      } else {
+        // Firefox throws if storage.managed is not found unlike Chrome
+        // returning empty object
+        const managedConfigFromBackend = chrome.storage.managed
+          .get()
+          .then(function (managedConfig) {
+            chrome.storage.local.set({ managedConfig });
+            return managedConfig;
+          });
```
smalluban: If we add here a catch block, the whole try/catch won't be needed.

smalluban (changes requested):

src/store/managed-config.js
seia-soto: Updated.

src/store/managed-config.js
```diff
@@ -37,45 +37,43 @@ const ManagedConfig = {
...
           chrome.storage.local.set({ managedConfig });
-        } else {
-          // Try to get local version as fallback
-          const { managedConfig: fallbackConfig = {} } =
-            await chrome.storage.local.get('managedConfig');
+          return managedConfig;
+        })
+        .catch(function (e) {
+          // `e.message` equals `Managed storage manifest not found` here
+          if (__FIREFOX__ && e?.message.includes('not found')) return {};
```
smalluban: We have to always catch any errors. It can also fail on Chrome for an unknown reason. If we don't catch that, the whole extension won't work, as the options will not be resolved correctly.

smalluban (changes requested):

seia-soto:
@smalluban I'd appreciate if you can update the status on this pull request.

smalluban (approved):

PR #3366, opened 2026-05-22
Release v10.5.44

smalluban:
6ab6bd5e fix(distractions): narrow Instagram reels filter to avoid breaking the page (#3374)
ff2307d0 chore(locales): update translations (#3373)
3aa501f3 fix(managed-config): add default value for fallback local storage lookup (#3371)
63dcb6af fix(context-menu): guard chrome.contextMenus for Firefox Android (#3372)
ed2bff37 chore(locales): update translations (#3360)
78f122f6 feat(distractions): add feature with basic options (#3350)
747d45af chore(deps): bump @duckduckgo/autoconsent from 14.82.0 to 14.84.1 (#3365)
a64423b3 chore(deps-dev): bump vite from 8.0.13 to 8.0.14 (#3364)
ba005603 chore(deps-dev): bump web-ext from 10.1.0 to 10.2.0 in the web-ext group (#3363)
47c7f5e4 chore(deps-dev): bump eslint from 10.3.0 to 10.4.0 in the eslint group (#3362)
f9ebd0ec chore(deps): bump the adblocker group with 4 updates (#3361)
81def406 fix(onboarding): wait for config sync before opening onboarding tab (#3359)
d245a698 fix(context-menu): add option to disable context menu (#3357)
a9b8e60f fix(utils/tabs): unify open tab logic (#3356)
6afff49d fix(ui): format large numbers with locale-aware Intl formatters (#3355)
87995a88 chore(scripts): add App Store Connect submission script (#3353)
d1f90391 ci(release): build and upload Safari extension zip (#3341)
fb50ecb7 fix(tests): make Firefox e2e setup work with BiDi parent-process restriction (#3354)

PR #3374, opened 2026-05-25
fix(distractions): narrow Instagram reels filter to avoid breaking the page

smalluban:
Reported in https://github.com/ghostery/ghostery-extension/pull/3366#pullrequestreview-4356324400: after opening a reel on Instagram, the whole page becomes unclickable when the "Short Video Feeds" distraction option is enabled.

The previous `article:has(a[href*="/reel/"])` filter was too broad — it also matched the reel viewer article when opened, causing it to be hidden and the page to become unresponsive. The filter is now scoped to `main article` to avoid matching the viewer overlay, and the href selector is changed from `*="/reel/"` (substring) to `^="/reel"` (starts-with) to also correctly match reel feed items whose anchors point to `/reels/` (the reels hub URL) rather than a specific reel URL. The same starts-with fix is applied to the `div[role="link"]` filter.


AdamGhst (approved):
Good!

PR #3376, opened 2026-05-26
fix(custom-filters): guard against DNR dynamic and regex rule limits

smalluban:
Chrome's `declarativeNetRequest` API enforces hard limits on dynamic rules (`MAX_NUMBER_OF_DYNAMIC_RULES`) and regex rules (`MAX_NUMBER_OF_REGEX_RULES`). Without guards, exceeding these limits causes silent failures when saving custom filters.

This adds runtime validation in `updateDNRRules()` before calling the API. Two constants derived from live API values are added to `src/utils/dnr.js`: `CUSTOM_FILTERS_MAX_DYNAMIC_RULES` (`MAX_NUMBER_OF_DYNAMIC_RULES − 5000`) reserves budget for all other dynamic rule consumers (fixes, exceptions, distractions, redirect protection, GPC, paused), and `CUSTOM_FILTERS_MAX_REGEX_RULES` (`MAX_NUMBER_OF_REGEX_RULES − 500`) reserves budget for regex rules from static rulesets (currently up to ~370). When either limit is exceeded, a descriptive error is thrown and caught in `updateCustomFilters()`, pushed into `result.errors`, and surfaced to the user via the existing error list in the settings page.

**Note on a related Edge/Chromium bug:** During this work, a reported issue was investigated where `getAvailableStaticRuleCount()` did not reset after manually uninstalling the extension on Edge on a clean profile — causing the subsequent reinstall to fail to enable static rulesets. The root cause is in `GlobalRulesTracker`: the global quota is tracked in-memory and re-initialized from prefs at browser start; if Edge wipes extension prefs *before* `OnExtensionUninstalled` fires, `ClearExtensionAllocation` cannot find the stored allocation and the in-memory counter is never decremented, persisting until the next browser restart. No specific upstream fix was found in Chromium source (`global_rules_tracker.cc` has had no functional changes since its 2020 introduction). The issue is no longer reproducible in current browser versions, suggesting it may have been silently fixed in a browser update, but we still receive error reports from affected users — likely on older Edge versions or in specific profile states.

### DIFF
--- a/src/background/custom-filters.js
+++ b/src/background/custom-filters.js
@@ -12,7 +12,12 @@
 import { store } from 'hybrids';
 import { parseFilters, FilterType } from '@ghostery/adblocker';
 
-import { CUSTOM_FILTERS_ID_RANGE, getDynamicRulesIds } from '/utils/dnr.js';
+import {
+  CUSTOM_FILTERS_ID_RANGE,
+  CUSTOM_FILTERS_MAX_DYNAMIC_RULES,
+  CUSTOM_FILTERS_MAX_REGEX_RULES,
+  getDynamicRulesIds,
+} from '/utils/dnr.js';
 import convert from '/utils/dnr-converter.js';
 import * as engines from '/utils/engines.js';
 import * as OptionsObserver from '/utils/options-observer.js';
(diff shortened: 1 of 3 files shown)


chrmod (approved):

chrmod (commented):
Something that we should test e2e maybe?

smalluban:
I had that in mind, but it needs to put 25k of lines in the textarea. I'll check if this won't be a bottleneck.

PR #3377, opened 2026-05-26
feat(adblocker): allow disabling individual filters from the logger

chrmod:


src/background/disabled-filters.js
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { STORAGE_KEY } from '/store/disabled-filters.js';
+
+let ids = new Set();
```
smalluban: This whole file is not required. The `chrome.storage.onChanged` event should be placed in the model definition file (see other models). It ensures that when the model is used (the file is imported), the listener is also registered. The `onChanged` should trigger `store.clear(DisabledFilters, false)`. 

EDIT: the model already have `onChanged` as I wanted, so just remove this file.

src/background/adblocker/cosmetics.js
```diff
@@ -228,7 +229,7 @@ async function injectCosmetics(details, config) {
     const scriptFilters = [];
 
     for (const { filter, exception } of matches) {
-      if (exception === undefined) {
+      if (exception === undefined && !isDisabled(filter.getId())) {
```
smalluban: use `disabledFilters = store.get(DisabledFilters)` and then `store.ready(disabledFilters) && disabledFilters.ids.includes(filter.getId())`

If the model is not ready, your current code will also return false. This is a pattern we use for sync functions in adblocker code, which requires access to async store models.

src/background/adblocker/network.js
```diff
@@ -82,30 +83,32 @@ if (__FIREFOX__) {
...
-          request.blocked = true;
-          // There's a possibility that redirecting to file URL can expose
-          // extension existence.
-          if (details.type !== 'xmlhttprequest') {
-            result = {
-              redirectUrl: chrome.runtime.getURL('rule_resources/redirects/' + redirect.filename),
-            };
-          } else {
-            result = { redirectUrl: redirect.dataUrl };
+        if (!filter || !isDisabled(filter.getId())) {
```
smalluban: Use the pattern I proposed for cosmetics.

src/background/index.js
```diff
@@ -18,6 +18,7 @@ import './adblocker/index.js';
 import './never-consent/index.js';
 import './cookies.js';
 import './custom-filters.js';
+import './disabled-filters.js';
```
smalluban: Remove the file

src/pages/logger/store/log.js
```diff
@@ -54,9 +56,9 @@ export default {
       return values;
     },
     list: ({ tabId, query, filterType }) => {
-      query = query && new RegExp(query.trim(), 'i');
+      const needle = query?.trim().toLowerCase();
```
smalluban, line 59: `needle` is a too generic name. I cannot connect the meaning of the variable to what it really does.

src/pages/logger/views/main.js
```diff
@@ -65,19 +66,66 @@ function disableEllipsis(host, event) {
...
+  },
   tabId: ({ tabs }, value) => {
     if (value !== undefined) return value;
     // Fallback to current active tab (default value)
     return store.ready(tabs) ? tabs.find((tab) => tab.active).id : '';
   },
   query: '',
-  filterType: 0,
-  render: ({ logs, tabs, tabId, filterType }) => html`
+  filterType: sessionPref('logger:filterType', (v) => Number(v) || 0),
```
smalluban: This is overkill, or you want to keep this preference after closing and opening the logger again. However, I think the logger should not mess with `sessionStorage`. Both `filterType` and `applied` can be local properties.

src/store/disabled-filters.js
```diff
@@ -0,0 +1,34 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+import { store } from 'hybrids';
+
+export const STORAGE_KEY = 'disabledFilterIds';
+
+const DisabledFilters = {
+  ids: [Number],
```
smalluban: There is a transform from array to Set in the view. As this is used as a dictionary, use `store.record(false)` - so you can access truthy keys: `disabledFilters.ids[filter.getId()]`

smalluban (changes requested):
The "disabling" network filters will work only for Firefox - It might be a good thing to show it somewhere in the UI.

src/pages/logger/views/main.js
```diff
@@ -65,19 +66,66 @@ function disableEllipsis(host, event) {
...
+function toggleState(log) {
+  if (!log.filterId) return 'none';
+  if (log.filterType === FilterType.COSMETIC) return 'toggleable';
+  if (log.filterType === FilterType.NETWORK) return __CHROMIUM__ ? 'unsupported' : 'toggleable';
+  return 'none';
+}
+
+async function toggleFilter(host, event, log) {
+  event.stopPropagation();
+  const resolved = await store.resolve(DisabledFilters);
```
smalluban: The host has the model already, use `host.disabledFilters`

src/pages/logger/views/main.js
```diff
@@ -65,19 +66,66 @@ function disableEllipsis(host, event) {
...
+  const ids = resolved.ids.includes(log.filterId)
+    ? resolved.ids.filter((id) => id !== log.filterId)
+    : [...resolved.ids, log.filterId];
+  await store.set(DisabledFilters, { ids });
```
smalluban: Using the dictionary, you can set/clear a specific key: `store.set(disabledFilters, { ids: { [id]: true || null }})`

src/pages/logger/store/log.js
chrmod: needle is a conventional name for such filtering algorithms - I strongly insists on keeping it.

chrmod:
> The "disabling" network filters will work only for Firefox - It might be a good thing to show it somewhere in the UI.

disable button is "disabled" on Chrome

smalluban (approved):

PR #3378, opened 2026-05-27
fix(managed-config): unify trustedDomains managed config model

smalluban:
Managed policy handling for trusted domains was using an array sentinel pattern that diverged from the structured managed config model and required special-case checks in consumers. This could make pause rule initialization harder to reason about on service worker restart.

This change refactors trusted domains in managed config to use an `{ enabled, domains }` shape, aligned with how custom filters are represented. It updates the managed storage translation step to convert legacy array input into the structured object and adjusts consumers in options and paused background logic to use the enabled/domains fields directly. The result removes sentinel-specific branching and keeps managed mode behavior consistent across startup and options application.

### DIFF
--- a/src/store/managed-config.js
+++ b/src/store/managed-config.js
@@ -13,17 +13,13 @@ import { store } from 'hybrids';
 
 import { isOpera, isWebkit } from '/utils/browser-info.js';
 
-export const TRUSTED_DOMAINS_NONE_ID = '<none>';
-
 const ManagedConfig = {
   disableOnboarding: false,
   disableUserControl: false,
   disableUserAccount: false,
   disableTrackersPreview: false,
 
-  // TODO: Update the structure of `trustedDomains` as is with `customFilters`
-  // to have `enabled` flag and `domains` array
-  trustedDomains: [TRUSTED_DOMAINS_NONE_ID],
+  trustedDomains: { enabled: false, domains: [String] },
   customFilters: { enabled: false, filters: [String] },
 
   disableNotifications: (config) => config.disableOnboarding || config.disableUserControl,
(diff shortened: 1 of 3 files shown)


src/store/managed-config.js
```diff
@@ -64,7 +60,14 @@ const ManagedConfig = {
         }
       }
 
-      // Translate `customFilters` storage.managed key (an array) to model structure
+      // Translate array storage.managed keys to model structure
+      if (managedConfig.trustedDomains) {
```
philipp-classen, line 64: Is it a problem if this code is executed more than once? That it keeps nesting the structure like here?

```
> managedConfig = { trustedDomains: { foo: 42 } }
{ trustedDomains: { foo: 42 } }

> if (managedConfig.trustedDomains) {
|   managedConfig.trustedDomains = {
|     enabled: true,
|     domains: managedConfig.trustedDomains,
|   };
| }
==>
{ enabled: true, domains: { foo: 42 } }

> managedConfig
{ trustedDomains: { enabled: true, domains: { foo: 42 } } }

> if (managedConfig.trustedDomains) {
|   managedConfig.trustedDomains = {
|     enabled: true,
|     domains: managedConfig.trustedDomains,
|   };
| }

==> { enabled: true, domains: { enabled: true, domains: { foo: 42 } } }


> if (managedConfig.trustedDomains) {
|   managedConfig.trustedDomains = {
|     enabled: true,
|     domains: managedConfig.trustedDomains,
|   };
| }

==>
{
  enabled: true,
  domains: { enabled: true, domains: { enabled: true, domains: [Object] } }
}
```

smalluban: This connect is a pure function, and `managedConfig` is always the raw data from managed storage or local cache. This is a transformation on the fly, and it is not stored. I am not sure if I understand your concerns.

philipp-classen: Okay, if it runs only ones, it is fine. That was my question. I have not read through the whole code. Was not aware that the data is only read and not written back.

chrmod (approved):

PR #3379, opened 2026-05-27
fix(config): update browser filter to use array instead of string

smalluban:
The `filter.browser` field in remote config was previously accepted as a string, while `filter.platform` already supports an array. Since the browser filter is not actively used on flags, this is a good moment to align both fields to the same array-based API.

This change updates the `filter()` function in `src/utils/config.js` to check `Array.isArray(item.filter.browser)` and use `.includes()` for matching, exactly mirroring the platform filter logic. Unit tests in `config.test.js` are updated accordingly to pass arrays instead of strings.

### DIFF
--- a/src/utils/config.js
+++ b/src/utils/config.js
@@ -61,8 +61,8 @@ export function filter(item) {
 
     // Browser check
     // Values from getBrowser() method
-    if (check && typeof item.filter.browser === 'string') {
-      check = getBrowser().name === item.filter.browser;
+    if (check && Array.isArray(item.filter.browser)) {
+      check = item.filter.browser.includes(getBrowser().name);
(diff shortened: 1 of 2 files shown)


chrmod (commented):
Isn't this breaking backwards compatibility? 
Cannot we have multiple rules one per each browser?

smalluban:
It does not break anything, as the current configuration does not use the `browser` filter option, and all "completed" flags are defined without filter options.

It is possible to create multiple instances of filter objects. However, it was a mistake in the definition, as `platform` accepts an array of strings. For parity, the `browser` should also accept an array of strings.

chrmod (approved):

PR #3381, opened 2026-05-27
fix(adblocker): inject scriptlets into their declared execution world

chrmod:
Some scriptlets (e.g. `set-attr`, `set-cookie`, `href-sanitizer`) declare that they must run in the **ISOLATED** execution world, out of the page's reach, so sites cannot tamper with the APIs they rely on. That declaration used to be ignored — every scriptlet ran in the page's MAIN world. Scriptlets are now routed to their declared world on both platforms:

- **Chromium** — injected with the declared `world`.
- **Firefox** — one content script is registered per declared world.

The minimum supported Firefox is raised from 115 to **128**, the first version where content-script registration supports `world` and `matchOriginAsFallback`. This also resolves the pre-existing breakage on Firefox 115–127, where scriptlet registration silently failed because it depended on those 128-only options — those versions simply stop receiving updates instead of carrying a separate injection fallback.

## Test plan

- With Ad-Blocking enabled, add a custom filter using an ISOLATED scriptlet (e.g. `<site>##+js(set-attr, #target, data-x, true)`) and confirm it takes effect on both Chrome and Firefox, even when the page overrides the API it relies on.
- Confirm MAIN-world scriptlets (e.g. `set`) still apply, and sites with existing cosmetic filters behave as before.


philipp-classen (approved):

seia-soto (approved):
Positive but not sure about the side-effects yet

Tristan-deAlwis (commented):
nice

.github/workflows/test.yml
```diff
@@ -86,3 +86,20 @@ jobs:
...
+      FIREFOX_VERSION: esr_115.0
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-node@v5
+        with:
+          node-version-file: '.tool-versions'
+      - name: Install dependencies
+        run: npm ci
+      - name: Run scriptlet execution-world tests
+        run: npm run wdio -- --target=firefox --spec tests/e2e/spec/adblocker.spec.js --mochaOpts.grep "Execution world"
```
seia-soto: I don't think we need to add a separate target because of this. I would rather document the minimum supported version or programmatically disable the execution world application depending on the version. From https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/ExecutionWorld, I see 125 is the minimum version. I might be wrong because that's just a constant.

seia-soto: In different words, there's no need to run the minimum supported version as the latest version will support it too.

src/background/adblocker/content-scripts.js
```diff
@@ -9,29 +9,91 @@
...
+
+    // A sentinel set in the page world (read via `wrappedJSObject`) means the
+    // inline script ran; otherwise the page CSP blocked it, so retry through a
+    // blob: URL, which some policies still allow when inline scripts are denied.
+    if (self.wrappedJSObject[sentinel]) {
+      delete self.wrappedJSObject[sentinel];
+      return;
+    }
+
+    const blob = new self.Blob([scriptlets], { type: 'text/javascript; charset=utf-8' });
```
seia-soto: Is `self` necessary here?

src/background/adblocker/content-scripts.js
```diff
@@ -9,29 +9,91 @@
...
+function injectMainWorldScriptlets(details) {
+  const doc = document;
+  const { sentinel, scriptlets } = details;
+  const parent = doc.head || doc.documentElement;
+
+  try {
+    const inline = doc.createElement('script');
+    inline.textContent = `self[${JSON.stringify(sentinel)}]=true;\n${scriptlets}`;
+    parent.appendChild(inline);
+    inline.remove();
```
seia-soto: As far as I understand, `injectMainWorldScriptlets` is a script body to be executed. I don't see a clear reason to execute main world scriptlets through this rather just using the API: executeScript or contentScripts.register.

src/background/adblocker/cosmetics.js
```diff
@@ -66,19 +66,25 @@ function injectScriptlets(filters, hostname, details) {
 
     const func = scriptlet.func;
     const args = [scriptletGlobals, ...parsed.args.map((arg) => decodeURIComponent(arg))];
+    const declaredWorld = scriptlet.world === 'ISOLATED' ? 'ISOLATED' : 'MAIN';
```
seia-soto, line 79: This is used by multiple code branches, please declare the new variable to mimic enum behaviour like:

```
const EXECUTION_WORLD = {
  ISOLATED: 'ISOLATED',
  MAIN: 'MAIN'
}
```

or just use `browser.scripting.ExecutionWorld`.

tests/e2e/page/adblocker/world.html
```diff
@@ -0,0 +1,25 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Execution world</title>
+    <script>
+      // Neutralize setAttribute in the page's MAIN world. A scriptlet injected
+      // into the MAIN world is defeated by this override; one injected into its
+      // declared ISOLATED world keeps a pristine setAttribute and still applies.
```
seia-soto, line 10: It's difficult to tell what could be the best way to test this feature. uBlock Origin scriptlets often rely on safe-self snippet which is responsible for these kinds of attack. The test might be unreliable in the future and it could result in cat and mouse game which we don't want.

src/background/adblocker/content-scripts.js
seia-soto: This brings me a concern of late scriptlet injection.

chrmod: The `world` option is the reason: neither `contentScripts.register` nor `scripting.executeScript` can target the MAIN world before Firefox 128, and we support back to 115 (the `ExecutionWorld` type shows up in MDN earlier, but only with `ISOLATED`; `MAIN` landed in 128). So on 115–127 the APIs alone cannot reach the page's world at all.

`executeScript` also has exactly the timing problem you raise: we can only fire it in reaction to a navigation event, so it races the page's own scripts. A registered content script with `runAt: 'document_start'` is the one primitive Firefox guarantees runs before any page script. The bootstrap preserves that guarantee for the MAIN world too: the content script synchronously appends an inline `<script>`, which executes immediately, still within the document_start window. Only the blob: fallback (used when the page CSP blocks inline scripts) is best-effort.

Once the minimum supported Firefox reaches 128, we can pass `world: 'MAIN'` to the registration and delete the bootstrap.

.github/workflows/test.yml
chrmod: The job isn't checking that 115 supports the feature — it exercises the fallback branches that only pre-128 Firefox takes, which the latest-Firefox job can never reach: the `matchOriginAsFallback` rejection/retry in `contentScripts.register`, and the MAIN-world `<script>` bootstrap running on the oldest engine we support.

There's also nothing to version-gate: on Firefox we never pass `world` at all — the same bootstrap runs identically from 115 through current, so there is no execution-world application to disable. What the job pins down is the floor: a change that accidentally starts relying on a 128+ API fails CI instead of silently breaking ESR users.

It's scoped to the single "Execution world" suite, so it's cheap, and it can be deleted together with the fallbacks once the minimum supported Firefox is 128.

tests/e2e/page/adblocker/world.html
chrmod: Agreed — the fixture isn't meant to model anti-adblock defenses. The `setAttribute` override isn't an attack the scriptlet has to defeat; it's a world-membership probe: it makes the two worlds observably different so the test can assert *where* the scriptlet ran (ISOLATED keeps pristine prototypes, MAIN sees the page's). Because the assertion is about routing rather than beating a hostile page, it holds however the scriptlets themselves evolve, safe-self included. Worst case a scriptlet changes its declared world upstream and we swap the probe — a fixture update, not a cat-and-mouse escalation.

philipp-classen (approved):

seia-soto (changes requested):
I just searched when was Firefox ESR 115 released and it seems like in 2023. Please, consider dropping support for ESR 115. The latest release is FF ESR 140.12. It over-complicates the whole logic for old Firefox release in 2023. I would rather not apply execution world separation unless it's fully supported by the APIs we need.

Also, we should not rely on blob and DOM related features to inject main world scriptlet. We don't want DOM to slow down the script execution. Also, it is more dangerous to expose contents on DOM. We do have "extension id" injection in the scriptlet payload and if the website is able to lever the observer, the extension would be "observable".

src/background/adblocker/content-scripts.js
seia-soto: Regarding timing problem, the assumption is wrong. There's no benefit of document_start once you rely on DOM to inject the script.

philipp-classen (approved):

PR #3384, opened 2026-05-28
fix(custom-filters): preserve scriptlet arguments at injection time

chrmod:
## Summary
- Encode scriptlet arguments for accepted custom cosmetic filters so they survive injection unchanged

## Test plan
- [ ] Add a custom cosmetic filter that injects a scriptlet with arguments containing commas, quotes, or other special characters
- [ ] Reload the affected page and confirm the scriptlet runs with the original argument values
- [ ] Verify existing custom filters without scriptlet arguments continue to work

### DIFF
--- a/src/background/custom-filters.js
+++ b/src/background/custom-filters.js
@@ -30,6 +30,21 @@ function isTrustedScriptInject(scriptName) {
   );
 }
 
+function encodeScriptletArguments(filter) {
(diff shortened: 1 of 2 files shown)


smalluban (changes requested):
The change looks fine, but can we provide e2e tests for it?

smalluban (approved):

PR #3392, opened 2026-05-29
Update uninstall URL

philipp-classen:
Skip the URL shortener and use the direct URL for now


chrmod (commented):
@philipp-classen how about we redirect via www.ghostery.com so we can control the redirect server side?

PR #3394, opened 2026-05-29
fix(telemetry): fresh uninstalls should still go via telemetry

chrmod:
Supersedes https://github.com/ghostery/ghostery-extension/pull/3392

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -79,8 +79,6 @@ OptionsObserver.addListener(async function telemetry({ terms, feedback }, lastOp
 
     runner.setUninstallUrl();
     if (feedback) runner.ping('active');
-  } else {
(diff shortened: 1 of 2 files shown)


src/background/telemetry/index.js
```diff
@@ -79,8 +79,6 @@ OptionsObserver.addListener(async function telemetry({ terms, feedback }, lastOp
 
     runner.setUninstallUrl();
     if (feedback) runner.ping('active');
-  } else {
```
smalluban, line 82: Now it is not set until the terms are accepted. The idea was to set the URL before a user enables Ghostery.

smalluban (changes requested):

src/background/telemetry/index.js
Tristan-deAlwis, line 82: I don't see this URL set anywhere else, so it should stay

smalluban (approved):

PR #3396, opened 2026-05-29
Release v10.5.45

smalluban:
d0fb32a4 fix(trackers-preview): add missing Mobile Google SERP selectors (#3397)
c27bfdc9 fix(custom-filters): guard against DNR dynamic and regex rule limits (#3376)
fa59bfd0 fix(logger): add missing translate="no" to prevent auto-translation (#3395)
c20802dd fix(custom-filters): preserve scriptlet arguments at injection time (#3384)
ef62a53a fix(managed-config): unify trustedDomains managed config model (#3378)
6e5b4b42 ci(release): build and upload automation extension zips (#3345)
31a99ce7 feat(adblocker): allow disabling individual filters from the logger (#3377)
46fee80a feat(config): update browser filter to use array instead of string (#3379)
b4819f64 chore(deps-dev): bump eslint-plugin-prettier (#3385)
faf54abb chore(deps): bump @sentry/browser from 10.53.1 to 10.55.0 (#3388)
3cef68a9 chore(deps): bump tldts-experimental from 7.0.30 to 7.4.0 (#3391)
34dcf067 chore(deps-dev): bump license-report from 6.8.2 to 6.8.5 (#3390)
0de4d9e8 chore(deps): bump @duckduckgo/autoconsent from 14.84.1 to 14.85.2 (#3389)
d52993d6 chore(deps-dev): bump web-ext from 10.2.0 to 10.3.0 in the web-ext group (#3387)
399eea5c chore(deps-dev): bump the wdio group with 5 updates (#3386)
c047f337 feat(errors): expand Sentry tags & update once parameter (#3383)
8b26b31b fix(logger): guard against undefined filter in filter-matched handler (#3375)
29bc812c chore(pause-assistant): remove FLAG_PAUSE_ASSISTANT flag check (#3380)


smalluban:
If you need to include still open PRs, follow the release procedure, section "Updating the Release".

AdamGhst (approved):
Smoke tests 10.5.45

All tests passed. All the above is ok. 
Tested on:
- macOS / iOS: Safari
- macOS: Chrome, Firefox, Edge, Brave (shields off)
- PC / Win10: Chrome, Firefox, Opera

The only problem is on YouTube: a loading loop or a long load time before the video starts. It lasts about as long as the background ad would, max 5-8s, usually 3-4s. So the expirience isn't bad. Not as smooth as competitors yet (copmpared with AdGuard), but clearly better than before the fix.

Once you sit through one loop (it plays an unseen ad in the background), the problem disappears and Ghostery works flawlessly; it only comes back after restarting the browser or clearing cookies. Personally, after an initial 2-10s loop I didn't hit it again.

The loop is much rarer on Firefox than on Chromium-based browsers – there it's basically random, and I only saw it on the first video load after a fresh browser start. But that's Windows only; on macOS Firefox is just as affected as Chromium.

A user can still occasionally hit an infinite loop where the video never starts. It's rare, and the easy fix is to switch to another tab and back, YouTube then plays immediately. Switching tabs also speeds up the start when there's just a long load time. This infinite loop is the only scenario where Ghostery actually blocks the experience.

Overall Ghostery works well and the release is awesome, we just need to take care of the YouTube problem.

PR #3410, opened 2026-06-09
Show excepted filters correctly in the Logger

chrmod:
The Logger rendered matched-but-excepted filters identically to applied ones, so an excepted filter (e.g. a scriptlet disabled by an exception) looked like it was still active. It also silently stopped logging once the main engine was rebuilt — for example after saving custom filters — because it stayed subscribed to the stale engine instance.

## Changes

- The Logger now flags exceptions: the filter text is dimmed and a pause icon with an explanatory tooltip is shown next to it, based on the exception state reported by the filter matching.
- `engines` exposes `addSaveListener()`, which notifies subscribers whenever an engine instance is replaced in memory. The background logger uses it to re-bind its `filter-matched` subscription to the new main engine, so logging survives engine rebuilds.

## Test plan

1. Open the Logger and visit a page where filters match — matched filters appear as before.
2. Add an exception for one of the matched filters (e.g. a `#@#` or scriptlet exception in custom filters) and reload the page — the excepted filter shows up dimmed with a pause icon instead of looking applied.
3. Keep the Logger open, save custom filters (this rebuilds the engine), then reload the page — new entries keep appearing in the Logger.

### DIFF
--- a/src/background/logger.js
+++ b/src/background/logger.js
@@ -120,7 +120,7 @@ chrome.runtime.onConnect.addListener(async (port) => {
 
       engine.on('filter-matched', logFilterMatched);
 
-      const offReplace = engines.onReplace(engines.MAIN_ENGINE, (nextEngine) => {
+      const unsubscribeSaveListener = engines.addSaveListener(engines.MAIN_ENGINE, (nextEngine) => {
         engine.unsubscribe('filter-matched', logFilterMatched);
         engine = nextEngine;
         engine.on('filter-matched', logFilterMatched);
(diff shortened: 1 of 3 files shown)


src/utils/engines.js
```diff
@@ -76,6 +77,23 @@ function loadFromMemory(name) {
...
   engines.set(name, engine);
+
+  const listeners = engineReplaceListeners.get(name);
+  if (listeners) {
+    for (const callback of listeners) callback(engine);
+  }
+}
+
+// The in-memory engine instance is replaced (not mutated) on reload, so holders must re-bind.
+export function onReplace(name, callback) {
```
smalluban: Change to `addSaveListener()`

src/background/logger.js
```diff
@@ -120,11 +120,18 @@ chrome.runtime.onConnect.addListener(async (port) => {
 
       engine.on('filter-matched', logFilterMatched);
 
+      const offReplace = engines.onReplace(engines.MAIN_ENGINE, (nextEngine) => {
```
smalluban: It should be `unsubscribeSaveListener` (as we will have `engines.addSaveListener()`)

smalluban (changes requested):

PR #3411, opened 2026-06-09
fix(adblocker): make scriptlet injection idempotent

chrmod:
A single navigation fires several injection triggers, so a scriptlet can run more than once in the same document — and duplicate runs (timers, observers, patched APIs) can break pages.

Each scriptlet is now wrapped at build time with a guard that runs it at most once per document and argument set. The guard hides its state behind a `Proxy` over `document.evaluate`, so it adds no page-enumerable property; it fails open on any error and claims a run only after success, so the worst case is a duplicate run, never a suppressed one.

All platforms share one `scripting.executeScript` injection path (Firefox ≥ 128, #3381). Because event-driven injection races page scripts, Firefox additionally keeps per-hostname `contentScripts.register` registrations — engine-level `document_start` injection on every navigation after the first — and the guard dedupes the overlap.

### Test plan

- With `example.com##+js(set-constant, foo, 1)`: value is set on first load, after reload, and in same-site iframes — on Chromium and Firefox.
- On Firefox, reload a page whose inline `<head>` script reads the constant — it is already set (document_start).
- Pause the site: scriptlet stops; resume: it returns. Edit custom filters: the change applies on next load.
- The site's own `document.evaluate` XPath calls still work.


philipp-classen (approved):

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,71 @@
...
+      `[generate-scriptlets] expected "func" to be a function for scriptlet "${name}", got ${typeof entry.func}`,
+    );
+  }
+
+  const lines = [`  aliases: ${JSON.stringify(entry.aliases || [])},`];
+  if (entry.world) lines.push(`  world: ${JSON.stringify(entry.world)},`);
+  lines.push(`  requiresTrust: ${!!entry.requiresTrust},`);
+  lines.push(`  func: ${wrapScriptletSource(entry.func.toString(), { debug, name })},`);
+
+  return `SCRIPTLETS[${JSON.stringify(name)}] = {\n${lines.join('\n')}\n};`;
```
seia-soto: I don't see any reason to merge multiple lines here. The comment could be seen as stylistic but it seems like we can just rely on JSON.stringify for the whole object.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,71 @@
...
+  func: function (scriptletGlobals, ...args) {
+    self[args[0]] = (self[args[0]] || 0) + 1;
+  },
+};
+
+const BANNER = `/* eslint-disable */
+// Generated by scripts/generate-scriptlets.js — DO NOT EDIT.
+// Each scriptlet func is wrapped with an in-document idempotency guard.`;
+
+function emitEntry(name, entry, debug) {
```
seia-soto: I don't get the use of `debug` here. The extension should be able to inject arguments through the filters.

scripts/utils/wrap-scriptlet.js
```diff
@@ -0,0 +1,43 @@
...
+    } catch (e) {}
+  }
+  try {
+    __orig.apply(this, __args);
+  } catch (e) {${logError}
+    return undefined;
+  }
+  if (__reg) __reg.add(__guardToken);
+  return true;
+}`;
```
seia-soto: This could be very risky even if we know that this will work. Please, make sure to rely on `Function.prototype.toString()` instead of typing everything in literal. We wouldn't be able to debug it easily if any kinds of chore syntax or type error happens. The function still could be safely converted into the string if it's self-containable.

src/background/adblocker/cosmetics.js
```diff
@@ -10,9 +10,11 @@
  */
 
 import { store } from 'hybrids';
-import scriptlets from '@ghostery/scriptlets';
 import { FLAG_SUBFRAME_SCRIPTING } from '@ghostery/config';
 
+// Generated at build time (gitignored), see scripts/generate-scriptlets.js.
+import scriptlets from './scriptlets.generated.js';
```
seia-soto, line 16: Should we do this in the `ghostery/scriptlets` repository instead?

scripts/utils/wrap-scriptlet.js
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+// Fail open: a broken guard must never suppress the scriptlet. Tokens are claimed
+// only after the original returns (throwing injections retry); in-world execution
+// is run-to-completion, so check-then-claim cannot interleave.
```
seia-soto: This method is fully detectable by the website. Also, I strongly suggest that most of fingerprinters or bot-detections will scan properties on window and more. To be more safe, we also need to override the defineProperty and defineProperties and more.

seia-soto (changes requested):
This method of guarding is not safe at all, especially about the main world scripts. We don't need to worry about the leaks under isolated world tho.

Also, it's questionable if we should have another "generated scriptlets" rather than placing under "ghostery/scriptlets" repository.

src/background/adblocker/cosmetics.js
chrmod: yes, it will land there in the next iteration

scripts/utils/wrap-scriptlet.js
chrmod: Reworked: the guard no longer writes to the page global. State now lives in a closure behind a `Proxy` over `document.evaluate`, so there is no property to enumerate and no need to trap `defineProperty`.

scripts/utils/wrap-scriptlet.js
chrmod: Done — the guard is now a real `guardRunOnce` function serialized via `Function.prototype.toString()`, not a source-string literal.

scripts/utils/scriptlets-module.js
chrmod: Real scriptlets are idempotent by design, so none can reveal a double-execution. This counter increments per run so the e2e suite can assert exactly-once injection; it stays behind `debug` so it never ships.

src/background/adblocker/cosmetics.js
```diff
@@ -48,17 +50,19 @@ const scriptletGlobals = {
...
-function injectScriptlets(filters, hostname, details) {
-  if (__FIREFOX__) {
-    if (filters.length === 0) {
-      contentScripts.unregister(hostname);
-      return;
-    }
-    if (contentScripts.isRegistered(hostname)) {
-      return;
-    }
+// Random so a page cannot forge the guard handshake; per-hostname so it is not a cross-site identifier.
```
philipp-classen: Only under the assumption that our code runs first (not guaranteed!).

Not a problem in itself, because in that scenario, there is not much that we can do anyway (the page can already neutralize the injected scriptlets). But the "cannot forge" claim overstates the guarantees. When taken literally, it may lead to wrong assumptions at some point.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+// patching the existing slot in place adds no page-enumerable property.
+function guardRunOnce(secret, token, orig, args, thisArg, logError) {
+  function run() {
+    try {
+      orig.apply(thisArg, args);
+      return true;
+    } catch (e) {
+      if (logError) logError(e);
+      return undefined;
+    }
```
seia-soto, line 22: Having this actually makes the implementation further from the original behaviour. It will throw if it's expected to throw and it will return value if it's expected to return the value. Please, do not wrap this in the try...catch block.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+
+  // Native `evaluate` throws on these arguments; only the guard proxy answers, echoing the secret.
+  function ask(key, claim) {
+    try {
+      var answer = document.evaluate(secret, key, claim);
+      return answer && typeof answer === 'object' && answer.s === secret ? answer : null;
+    } catch {
+      return null;
+    }
+  }
```
seia-soto: We can inline the whole thing. It's over-complicated to wrap everything in the function because Proxy's callbacks are designed to handle those directly.

seia-soto: Also, there's no point of having argument injection in this function like: `token`. I would rather just use the `token` variable directly since this function is already inlined.

src/background/adblocker/cosmetics.js
```diff
@@ -48,17 +50,19 @@ const scriptletGlobals = {
...
+const guardSecrets = new Map();
+function getGuardSecret(hostname) {
+  let secret = guardSecrets.get(hostname);
+  if (!secret) {
+    if (guardSecrets.size >= 1000) guardSecrets.clear();
+    secret = crypto.randomUUID();
+    guardSecrets.set(hostname, secret);
   }
+  return secret;
+}
```
seia-soto, line 68: We don't need to do this because they're disposable if we call `randomUUID` per `injectScriptlets` call. It's not designed to have many overlaps but it can still have overlaps. Other possible solution is just relying on `runtime.getURL` because it will generate random identifier per browser session. I don't think it's superior but it's acceptable imo.

src/background/adblocker/content-scripts.js
```diff
@@ -13,9 +13,15 @@ export const contentScripts = (() => {
   const map = new Map();
   return {
     async register(hostname, scriptletsByWorld) {
+      const key = JSON.stringify(scriptletsByWorld);
+      if (map.get(hostname)?.key === key) return;
+
+      if (!map.has(hostname) && map.size >= 1000) this.unregisterAll();
```
seia-soto: I'm not sure about adding this restriction. It'd be better not to store the whole key in the string if the purpose was to reduce the memory usage or pressure. We have decent hashing function (in 32-bit).

src/background/adblocker/content-scripts.js
```diff
@@ -13,9 +13,15 @@ export const contentScripts = (() => {
   const map = new Map();
   return {
     async register(hostname, scriptletsByWorld) {
+      const key = JSON.stringify(scriptletsByWorld);
+      if (map.get(hostname)?.key === key) return;
```
seia-soto: This `map.get(hostname)` is called multiple times, please put it in the variable.

src/background/adblocker/content-scripts.js
```diff
@@ -24,30 +30,27 @@ export const contentScripts = (() => {
...
               runAt: 'document_start',
               world,
             }),
           );
+
+          // A newer registration replaced this entry while awaiting.
+          if (map.get(hostname)?.registered !== registered) {
+            registered.pop().unregister();
+            return;
+          }
```
seia-soto, line 46: This probably cannot happen even considering the async task because another one is very likely to fail on the initial key check. Also, it's more reasonable to using `key` (JSON stringified full scriptlets) instead of `registered`.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+      while (owner && !Object.prototype.hasOwnProperty.call(owner, 'evaluate')) {
+        owner = Object.getPrototypeOf(owner);
+      }
+      if (!owner) return run();
+
+      // Keyed by document: in Firefox the patched prototype survives the initial about:blank swap.
+      var store = new WeakMap();
+      Object.defineProperty(owner, 'evaluate', {
+        value: new Proxy(evaluate, {
+          apply: function (target, that, callArgs) {
```
seia-soto: `that` is `thisArg`.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+            if (callArgs[0] !== secret || !that) return target.apply(that, callArgs);
```
seia-soto: You should rely on `Reflect.apply` instead.

seia-soto: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/apply

src/background/adblocker/cosmetics.js
```diff
@@ -77,17 +81,18 @@ function injectScriptlets(filters, hostname, details) {
     }
 
     const func = scriptlet.func;
-    const args = [scriptletGlobals, ...parsed.args.map((arg) => decodeURIComponent(arg))];
+    const token = `${scriptletName}\x1f${parsed.args.join('\x1f')}`;
```
seia-soto, line 89: I don't mind this but I personally prefer having a hashing function with JSON.stringify.

src/background/adblocker/cosmetics.js
```diff
@@ -336,18 +320,9 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
...
+  OptionsObserver.addListener('paused', function unregisterScriptletsOnPause() {
+    contentScripts.unregisterAll();
   });
-
-  chrome.webNavigation.onBeforeNavigate.addListener(
-    (details) => {
-      injectCosmetics(details, { bootstrap: true, scriptletsOnly: true });
-    },
-    { url: [{ urlPrefix: 'http://' }, { urlPrefix: 'https://' }] },
-  );
```
seia-soto, line 350: This is slightly dropped without a proper description. `scriptletsOnly` argument is for this.

tests/unit/content-scripts.test.js
```diff
@@ -0,0 +1,137 @@
...
+      await contentScripts.register('example.com', { MAIN: 'm;', ISOLATED: '' });
+    } finally {
+      console.warn = warn;
+    }
+
+    globalThis.browser.contentScripts.register = async (options) => {
+      const handle = { options, unregistered: false, unregister() {} };
+      registrations.push(handle);
+      return handle;
+    };
```
seia-soto: `globalThis.browser.contentScripts.register` is implemented twice, please swap the function instead.

src/background/adblocker/cosmetics.js
seia-soto: Also, please try to combine secret and token. I think they're doing the same job.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+  if (!status) {
+    try {
+      var evaluate = document.evaluate;
+      if (typeof evaluate !== 'function') return run();
+
+      var owner = document;
+      while (owner && !Object.prototype.hasOwnProperty.call(owner, 'evaluate')) {
+        owner = Object.getPrototypeOf(owner);
+      }
+      if (!owner) return run();
```
seia-soto, line 48: We don't need this if we use `Reflect`, see below: https://github.com/ghostery/ghostery-extension/pull/3411/files#r3524313512.

tests/e2e/spec/scriptlet-idempotency.spec.js
```diff
@@ -0,0 +1,84 @@
...
+
+function readMarker(id) {
+  return browser.execute((el) => document.getElementById(el)?.textContent, id);
+}
+
+async function expectRanExactlyOnce(id, expected) {
+  try {
+    await browser.waitUntil(async () => (await readMarker(id)) === expected, {
+      timeout: 10000,
+    });
```
seia-soto, line 29: We can reduce the timeout of 10 seconds.

tests/e2e/spec/scriptlet-idempotency.spec.js
```diff
@@ -0,0 +1,84 @@
...
+  after(disableCustomFilters);
+
+  it('injects each scriptlet and argument set exactly once per document, including after reload', async function () {
+    await browser.url(PAGE_URL);
+    await expectRanExactlyOnce('rpnt-a', 'aaa+');
+    await expectRanExactlyOnce('rpnt-b', 'bbb+');
+
+    await browser.refresh();
+    await expectRanExactlyOnce('rpnt-a', 'aaa+');
+  });
```
seia-soto, line 74: I don't think this test is explicitly not required, or it would be better to test both filters after reload not the single one.

tests/unit/content-scripts.test.js
```diff
@@ -0,0 +1,137 @@
...
+        this.unregistered = true;
+      },
+    }));
+    pending[0](handles[0]);
+    pending[1](handles[1]);
+    await Promise.all([first, second]);
+
+    assert.equal(handles[0].unregistered, true);
+    assert.equal(handles[1].unregistered, false);
+  });
```
seia-soto, line 100: This testing validates the behaviour but it doesn't prove that manual unregister call in the try...catch block in content-scripts.js:l49.

tests/unit/content-scripts.test.js
```diff
@@ -0,0 +1,137 @@
...
+      matchOriginAsFallback: true,
+      runAt: 'document_start',
+      world: 'MAIN',
+    });
+
+    await contentScripts.register('worlds.com', { MAIN: 'm;', ISOLATED: 'i;' });
+    assert.deepEqual(
+      registrations.slice(1).map((r) => r.options.world),
+      ['MAIN', 'ISOLATED'],
+    );
```
seia-soto, line 62: I don't think this has to be tested through array. Please, use `'x' in obj` check instead. Otherwise, any kinds of atomic checks are fine.

tests/unit/scriptlet-guard.test.js
```diff
@@ -0,0 +1,158 @@
...
+      assert.equal(counter(SECRET, 't'), true);
+      assert.equal(counter(SECRET, 't'), true);
+    } finally {
+      globalThis.Proxy = OriginalProxy;
+    }
+
+    globalThis.document = {};
+    assert.equal(counter(SECRET, 't'), true);
+    assert.equal(globalThis.self.__c, 5);
+  });
```
seia-soto, line 157: We don't need to describe this because not having Proxy in global means the page already hooked the Proxy object and our scriptlets are running too late or the environment is not perfect. Since we expect the extension to run recent versions of browsers, we can ensure that this behaviour test is not required.

src/background/adblocker/cosmetics.js
seia-soto: Leaving a note: `chrome.webNavigation.onBeforeNavigate` is NOT triggered on client-side navigation.

tests/unit/scriptlet-guard.test.js
```diff
@@ -0,0 +1,158 @@
...
+  it('dedups per document when the patched prototype survives a navigation', () => {
+    const proto = Object.getPrototypeOf(globalThis.document);
+
+    assert.equal(counter(SECRET, 't'), true);
+    assert.equal(counter(SECRET, 't'), false);
+
+    globalThis.document = Object.create(proto);
+    assert.equal(counter(SECRET, 't'), true);
+    assert.equal(globalThis.self.__c, 2);
+  });
```
seia-soto: Please, note that `chrome.webNavigation.onBeforeNavigate` is NOT triggered on client-side navigation. Also, we don't expect it to happen with the current code with my understanding. Please, provide why this test is required.

tests/unit/scriptlet-guard.test.js
```diff
@@ -0,0 +1,158 @@
...
+
+    const windowKeysBefore = Object.getOwnPropertyNames(globalThis).sort();
+    counter(SECRET, 't');
+
+    assert.deepEqual(Object.getOwnPropertyNames(document), []);
+    assert.deepEqual(Object.getOwnPropertyNames(proto), ['evaluate']);
+    assert.equal(Object.getOwnPropertyDescriptor(proto, 'evaluate').enumerable, false);
+    assert.deepEqual(Object.getOwnPropertyNames(globalThis).sort(), windowKeysBefore);
+    assert.deepEqual(document.evaluate('//div', document), { nativeXPathResult: true });
+  });
```
seia-soto, line 117: I think we don't need this as we explicitly specify `false` to `enumerable`.

seia-soto: I would recommend leaving a comment instead. I feel like it's testing browser spec not the feature.

tests/unit/scriptlet-guard.test.js
```diff
@@ -0,0 +1,158 @@
...
+
+  it('passes `this`, scriptletGlobals and args through to the original', () => {
+    const fn = build('function (g, ...a) { self.seen = { x: this.x, g: g, args: a }; }');
+
+    fn.call({ x: 42 }, SECRET, 't', { marker: 1 }, 'a', 'b');
+
+    assert.equal(globalThis.self.seen.x, 42);
+    assert.deepEqual(globalThis.self.seen.g, { marker: 1 });
+    assert.deepEqual(globalThis.self.seen.args, ['a', 'b']);
+  });
```
seia-soto, line 103: I don't think we need to describe how `Function.prototype.call` works.

tests/unit/scriptlet-guard.test.js
```diff
@@ -0,0 +1,158 @@
...
+
+    const calls = captureConsoleError(() => {
+      assert.equal(thrower(SECRET, 'boom'), undefined);
+      assert.equal(thrower(SECRET, 'boom'), undefined);
+    });
+
+    assert.equal(globalThis.self.__t, 2);
+    assert.equal(calls.length, 2);
+    assert.match(calls[0].join(' '), /broken\.js.*boom/);
+  });
```
seia-soto: I explicitly don't want this behaviour. If the scriptlet with specific argument fails, we see it failed not something else. Imagine it did some modification to the environment and crashed at the middle. We cannot ensure the environment is clean.

seia-soto (changes requested):
I was thinking about what criteria can be applied when choosing function to override or patch. From that, I would recommend using `Number.isSafeInteger` or `Number.isFinite`. These functions are (1) not throwing an error, (2) has simple call structure, (3) reasonable rarely used, (4) doesn't belong to prototype.

Not throwing an error means the attacker cannot see the stack-trace easily to detect the proxy, and having simple call structure and not having prototype mean we don't need to deal with complexity. Regarding `Object.is`, I believe it is used by Vue.js data processing, so I would like to avoid it.

I also left other comments.

tests/e2e/spec/scriptlet-idempotency.spec.js
chrmod: Reduced to 5s.

scripts/utils/scriptlets-module.js
chrmod: Kept the try/catch: on Firefox the scriptlets for one world are concatenated into a single content-script blob, so an uncaught throw would abort every scriptlet after it in the batch. Changed the claim instead — the token is now claimed before the run, so a scriptlet that throws runs exactly once and isn't retried against a mutated document.

tests/unit/scriptlet-guard.test.js
chrmod: Kept for batch isolation (a throw would otherwise abort the rest of the concatenated content-script blob on Firefox). Addressed the retry concern: the token is claimed before the run, so a throwing scriptlet executes exactly once. Test updated to assert run-once.

scripts/utils/scriptlets-module.js
chrmod: Dropped the argument — `ask` now reads `token` from the closure. Kept it as a function since it's called at three sites.

scripts/utils/scriptlets-module.js
chrmod: Done — the trap now uses `Reflect.apply(target, receiver, callArgs)`.

scripts/utils/scriptlets-module.js
chrmod: Renamed `that` → `receiver`.

src/background/adblocker/content-scripts.js
chrmod: Done — hoisted to `const existing = map.get(hostname)` before the await.

src/background/adblocker/content-scripts.js
chrmod: Kept the full-string key. It gates re-registration, so a hash collision would silently skip a needed re-register. The map is capped at 1000 entries, so memory is bounded.

src/background/adblocker/content-scripts.js
chrmod: It can happen: a register() with different code for the same host calls unregister() and sets a new entry while an earlier call is still awaiting browser.contentScripts.register — covered by the "discards a registration that resolves after being replaced" test. A same-key concurrent call returns at the initial key check, so reaching this line always means a different registration replaced this one; the `registered` identity check is exact.

src/background/adblocker/cosmetics.js
chrmod: The executeScript path and the persistent contentScripts.register path must share one secret per host, or their guards use different secrets and both run. A fresh UUID per injectScriptlets call would break that, so it's cached. The map is capped at 1000 and cleared on overflow.

chrmod: Different roles: secret authenticates the handshake (proves the answer came from our proxy, not the page); token identifies the (scriptlet, args) pair for dedup. Many tokens live under one secret, so they can't be merged.

scripts/utils/scriptlets-module.js
chrmod: Independent of Reflect.apply — the owner-walk selects which object in the prototype chain owns `evaluate`, so the patch lands on the prototype and survives Firefox's about:blank swap (per-document dedup test). Reflect.apply only changes how the trap forwards the native call.

tests/unit/content-scripts.test.js
chrmod: Done — the test now saves the beforeEach mock and restores it, instead of redefining register.

tests/unit/content-scripts.test.js
chrmod: Added "unregisters already-registered worlds when a later world fails": the second world's register throws and the test asserts the first world's handle is unregistered, exercising the catch block.

tests/e2e/spec/scriptlet-idempotency.spec.js
chrmod: Done — the reload now asserts both rpnt-a and rpnt-b.

tests/unit/scriptlet-guard.test.js
chrmod: Kept — the guard falls open to a plain run when Proxy or document.evaluate is unavailable, and the test pins that behaviour so a future change can't silently start suppressing scriptlets.

tests/unit/scriptlet-guard.test.js
chrmod: Renamed to name the trigger: Firefox's initial about:blank→document swap at document_start, where the prototype patch persists but the document instance changes — not client-side navigation. The store is a WeakMap keyed by document, so each document dedups independently, which is what the test checks.

tests/unit/scriptlet-guard.test.js
chrmod: Kept as a one-line assertion — it pins the no-page-enumerable-property invariant, so flipping enumerable to true (which would leak a detectable property) fails the test instead of passing silently.

tests/unit/scriptlet-guard.test.js
chrmod: Not testing `call` — it verifies the wrapper threads `this`, scriptletGlobals, and the args through to the original scriptlet in the right order, which could regress.

tests/unit/content-scripts.test.js
chrmod: Kept the array assertion — it checks in one place that both worlds register and in what order.

src/background/adblocker/cosmetics.js
chrmod: Replaced rather than dropped: on Firefox scriptlets are now registered as document_start content scripts (contentScripts.register); on Chromium injection is driven by webRequest.onResponseStarted. Both cover the early-injection case the onBeforeNavigate listener handled. Agreed it doesn't fire on client-side navigation.

src/background/adblocker/cosmetics.js
chrmod: Left as the delimiter-joined string — it's the dedup key, so exactness matters. The 0x1f unit separator can't appear in a scriptlet name and is vanishingly unlikely in args.

src/background/adblocker/cosmetics.js
chrmod: Reworded — the comment no longer claims the handshake can't be forged; it now notes a page running before us could observe the secret, so it hardens rather than guarantees.

scripts/utils/scriptlets-module.js
chrmod: The push-based line building was removed in a later refactor; entries are emitted directly now, and world/aliases already go through JSON.stringify. func stays templated because it's serialized source, not JSON data.

chrmod:
Pushed a5b1b8eb addressing the review.

**Behavior:** a scriptlet that throws now runs exactly once — the token is claimed before the run, so it isn't retried against an already-mutated document. Kept the per-scriptlet try/catch, since on Firefox a throw would otherwise abort the rest of the concatenated content-script blob.

**Cleanups:** `Reflect.apply` in the guard proxy, dropped the redundant `ask` argument, `that`→`receiver`, cached the pre-await `map.get`, softened the "cannot forge" comment. Tests: run-once assertion, new partial-unregister coverage, swap-not-redefine the register mock, both filters checked after reload, about:blank rename, 5s timeout.

**Left as-is (reasons inline):** the per-hostname secret cache (must be shared across the executeScript and content-script paths), full-string dedup keys, the replaced-during-await guard, the owner-walk (about:blank survival), and the fail-open / arg-plumbing tests.

chrmod:
> I would recommend using `Number.isSafeInteger` or `Number.isFinite`.

I tested the detectability assumptions in V8 before picking `document.evaluate`:

- `Function.prototype.toString.call(proxy)` returns `function () { [native code] }` — name dropped — for a proxy over any native (`Number.isSafeInteger.toString()` natively includes the name). So a one-line, throw-free `toString()` check detects the proxy on either candidate; not-throwing doesn't add stealth. The stack-trace vector is real, but it's a second way to see what `toString()` already reveals.
- A proxied `Number.isSafeInteger` benchmarks ~4× slower than native (~20 ns vs ~5 ns/call, more once V8 inlining is factored in). That's overhead on numeric hot paths and a clean timing oracle. On `document.evaluate` (µs-scale calls), the same trap cost is unmeasurable.
- Page code replaces or freezes `Number.*` (polyfills, SES `lockdown()`). On a frozen `Number`, `defineProperty` throws and the guard falls open — dedup is lost exactly where this PR fixes it. A replaced slot orphans the WeakMap store on re-patch. `Document.prototype.evaluate` sees no such traffic.
- The prototype placement is load-bearing: the receiver of `document.evaluate(secret, …)` is the document instance, which keys the WeakMap per document and makes the Firefox about:blank swap behave. With `Number.isSafeInteger` the receiver is the realm-singleton `Number`, so keying would need a closure-captured `document` — no net simplification.

On rarity: XPath is rarer in page code than numeric checks, so that criterion also favors `evaluate`. Agreed on avoiding `Object.is`.

src/background/adblocker/cosmetics.js
```diff
@@ -48,17 +50,19 @@ const scriptletGlobals = {
...
-function injectScriptlets(filters, hostname, details) {
-  if (__FIREFOX__) {
-    if (filters.length === 0) {
-      contentScripts.unregister(hostname);
-      return;
-    }
-    if (contentScripts.isRegistered(hostname)) {
-      return;
-    }
+// Unguessable per-hostname value: hardens the handshake (a page running before us could still observe it) and is not a cross-site identifier.
```
philipp-classen:
```suggestion
// crypto.randomUUID per hostname:
//  - random so forging the handshake by brute-force guessing is infeasible
//  - per-hostname, so the value cannot serve as a cross-site identifier
//
// Note that a page that runs before us can read the secret (via a hooked document.evaluate) and forge the handshake,
// but it can already neutralize scriptlets directly — so the guard is not the weak link.
// Also, hooking document.evaluate only works for scriptlets that are injected in MAIN, not ISOLATED.
```

I tried to clarify, since these are important parts of the threat-vector of this solution.

src/background/adblocker/cosmetics.js
```diff
@@ -48,17 +50,24 @@ const scriptletGlobals = {
...
-    }
-    if (contentScripts.isRegistered(hostname)) {
-      return;
-    }
+// crypto.randomUUID per hostname:
+//  - random so forging the handshake by brute-force guessing is infeasible
+//  - per-hostname, so the value cannot serve as a cross-site identifier
+//
+// Note that a page that runs before us can read the secret (via a hooked document.evaluate) and forge the handshake,
+// but it can already neutralize scriptlets directly — so the guard is not the weak link.
```
philipp-classen, line 58:
```suggestion
// but it can already neutralize scriptlets directly — so the guard is not the weak link.
// This document.evaluate attack vector only affects scriptlets that are injected into MAIN, not in ISOLATED.
```

seia-soto:
> > I would recommend using `Number.isSafeInteger` or `Number.isFinite`.
> 
> I tested the detectability assumptions in V8 before picking `document.evaluate`:
> 
>     * `Function.prototype.toString.call(proxy)` returns `function () { [native code] }` — name dropped — for a proxy over any native (`Number.isSafeInteger.toString()` natively includes the name). So a one-line, throw-free `toString()` check detects the proxy on either candidate; not-throwing doesn't add stealth. The stack-trace vector is real, but it's a second way to see what `toString()` already reveals.

Same restrictions apply to all hook candidates. `Function.prototype.toString.call(proxy)` should be handled generically and it is not a reason to keep `document.evaluate` specifically. A `document.evaluate` proxy has the same `toString` exposure as a `Number.isSafeInteger` proxy.

>     * A proxied `Number.isSafeInteger` benchmarks ~4× slower than native (~20 ns vs ~5 ns/call, more once V8 inlining is factored in). That's overhead on numeric hot paths and a clean timing oracle. On `document.evaluate` (µs-scale calls), the same trap cost is unmeasurable.


The performance argument also needs reproducible data. Without the benchmark script, engine/browser information, warm-up policy, iteration count, and variance, the `~4×` number is not enough to be the actual performance difference. It may be true that proxying `Number.isSafeInteger` is more likely to affect hot paths, but that should be demonstrated rather than asserted. On top of that, `Number.isSafeInteger` does not throw an error - also solving attack surface from stack trace. Also, it parses and evaluates XPath and can throw for invalid expressions, namespace issues, wrong-document contexts, or incompatible result types. That makes it a worse fit.

>     * Page code replaces or freezes `Number.*` (polyfills, SES `lockdown()`). On a frozen `Number`, `defineProperty` throws and the guard falls open — dedup is lost exactly where this PR fixes it. A replaced slot orphans the WeakMap store on re-patch. `Document.prototype.evaluate` sees no such traffic.

The freeze/replace concern is generic. That makes this not a correctness argument.

>     * The prototype placement is load-bearing: the receiver of `document.evaluate(secret, …)` is the document instance, which keys the WeakMap per document and makes the Firefox about:blank swap behave. With `Number.isSafeInteger` the receiver is the realm-singleton `Number`, so keying would need a closure-captured `document` — no net simplification.

The prototype-placement argument is false claim. There is no reason to store the value in the `document`. If the state is per frame, it can live in a closure. If subframes are involved, they should get their own scriptlet injection anyway. `document` is the value being replaced across the about:blank transition, so using the document receiver is convenient only if the design truly requires per-`Document` state. It is not required for preserving a value across the same reused execution context.

eslint.config.js
```diff
@@ -50,4 +50,5 @@ export default [
   {
     ignores: ['.wdio', 'dist/*', 'xcode/*', 'src/rule_resources*', 'web-ext-artifacts/*'],
   },
+  { ignores: ['**/*.generated.js'] },
```
seia-soto, line 53: nit: this doesn't need to be in a separate line.

scripts/utils/scriptlets-module.js
seia-soto: This is false claim. `try/catch` wouldn't stop other scriptlet runs if the initial run triggered via Proxy callback.

scripts/utils/scriptlets-module.js
seia-soto: Choosing `document.evaluate` brings all those complexity. I strongly suggest changing to atomic method.

Even if we're keeping `document.evaluate`, there's no reason to do this rather than relying on `Document.prototype.evaluate`. The approach is over-engineered.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+
+      // Keyed by document: in Firefox the patched prototype survives the initial about:blank swap.
+      var store = new WeakMap();
+      Object.defineProperty(owner, 'evaluate', {
+        value: new Proxy(evaluate, {
+          apply: function (target, receiver, callArgs) {
```
seia-soto, line 54: Please, keep `target, thisArg, args` as argument names. It's way more universal and direct. It's even used for many autocompletes and MDN documentations.

scripts/utils/scriptlets-module.js
```diff
@@ -0,0 +1,132 @@
...
+
+  if (status.has) return false;
+
+  // Claim before running so a scriptlet that throws mid-way is not re-run against a mutated document.
+  ask(true);
+  return run();
+}
+
+export function wrapScriptletSource(originalSource, name = 'scriptlet') {
+  const logError = `function (e) { console.error(${JSON.stringify(`[adblocker] ${name} failed:`)}, e); }`;
```
seia-soto, line 89: Passing the full function instead of `name` is not reasonable.

src/background/adblocker/content-scripts.js
```diff
@@ -24,30 +31,27 @@ export const contentScripts = (() => {
             await browser.contentScripts.register({
               js: [{ code }],
               allFrames: true,
-              matches: [`https://*.${hostname}/*`, `http://*.${hostname}/*`],
+              matches: [`https://${hostname}/*`, `http://${hostname}/*`],
```
seia-soto, line 34: Describe this changes. Often filters are scoped in the base domain.

src/background/adblocker/content-scripts.js
seia-soto: "discards a registration that resolves after being replaced" didn't trigger that code path. Please, clarify.

src/background/adblocker/cosmetics.js
seia-soto: This could be potentially unset after navigation completes/aborts rather keeping the 1000 entries as a ceiling. Also, `token` and `secret` can be a single value if we think that we just query per scriptlet per context. Different roles don't matter because they becomes 'pair'.

src/background/adblocker/cosmetics.js
seia-soto: Exactness is same for both cases.

src/background/adblocker/cosmetics.js
seia-soto: The only reason why this event handler was kept and not replaced by `webRequest.onResponseStarted` is because `webRequest.onResponseStarted` alone was not stable enough.

tests/unit/content-scripts.test.js
```diff
@@ -0,0 +1,160 @@
...
+        return handle;
+      };
+      await contentScripts.register('example.com', { MAIN: 'm;', ISOLATED: 'i;' });
+    } finally {
+      console.warn = warn;
+    }
+
+    assert.equal(handles.length, 1);
+    assert.equal(handles[0].unregistered, true);
+  });
```
seia-soto, line 145: I don't get the point of this test. The test specifies `unregisters already-registered worlds when a later world fails` but there's a sole register happening without follow-ups.

tests/unit/scriptlet-guard.test.js
seia-soto: I'd rather rely on behavioural test. This becomes just another maintenance whenever we modify the source code.

seia-soto: I suggest dropping the test.

tests/unit/scriptlet-guard.test.js
seia-soto: It's totally up to browser and it's more important that the source code specifies that. We should know what we're doing and we don't know the full list of what makes the function detectable. This test doesn't worth it.

tests/unit/scriptlet-guard.test.js
seia-soto: There's no case when Proxy or document.evaluate is unavailable.

tests/unit/scriptlet-guard.test.js
```diff
@@ -0,0 +1,158 @@
...
+
+    assert.equal(globalThis.self.__t, 1);
+    assert.equal(calls.length, 1);
+    assert.match(calls[0].join(' '), /broken\.js.*boom/);
+  });
+
+  it('passes `this`, scriptletGlobals and args through to the original', () => {
+    const fn = build('function (g, ...a) { self.seen = { x: this.x, g: g, args: a }; }');
+
+    fn.call({ x: 42 }, SECRET, 't', { marker: 1 }, 'a', 'b');
```
seia-soto, line 98: I'd rather just skip the `thisArg` and put `null` there. This can be more simplified.

seia-soto (changes requested):
`document.evaluate` is not a better approach than hooking other methods. It exposes more complexity with document swapping from `about:blank`'s initial navigation and attack surfaces via stack trace.

Also, `chrome.webNavigation.onBeforeNavigate` is still silently dropped.

tests/e2e/spec/scriptlet-idempotency.spec.js
seia-soto: 1 or 2s should be enough.

tests/e2e/spec/scriptlet-idempotency.spec.js
```diff
@@ -0,0 +1,85 @@
...
+async function ensureScriptletActive() {
+  await browser.waitUntil(
+    async () => {
+      await browser.url(PAGE_URL);
+      await browser.pause(300);
+      return ((await readMarker('rpnt-a')) || '').includes('+');
+    },
+    { timeout: 20000, interval: 500, timeoutMsg: 'scriptlet never became active' },
+  );
+}
```
seia-soto, line 51: This will be tested through adblocker test. Having idempotency test here would be enough.

tests/unit/content-scripts.test.js
```diff
@@ -0,0 +1,160 @@
...
+    assert.equal(registrations.length, 1);
+    assert.deepEqual(registrations[0].options, {
+      js: [{ code: 'm;' }],
+      allFrames: true,
+      matches: ['https://example.com/*', 'http://example.com/*'],
+      matchAboutBlank: true,
+      matchOriginAsFallback: true,
+      runAt: 'document_start',
+      world: 'MAIN',
+    });
```
seia-soto, line 56: This assert is not required. Please, drop. It just repeated what you've written in the source code.

src/background/adblocker/cosmetics.js
seia-soto: I don't know if this comment is required because late injection already means the extension doesn't have a control over the page script likely unless it's isolated.

src/background/adblocker/cosmetics.js
```diff
@@ -77,17 +86,18 @@ function injectScriptlets(filters, hostname, details) {
...
-      let code = '';
-      if (filter.hasSubframeConstraint()) {
-        code += `window.parent!==window&&`;
-      }
-      code += `(${func.toString()})(...${JSON.stringify(args)});\n`;
-      scriptletsByWorld[declaredWorld] += code;
-      continue;
+    // Subframe-constrained filters depend on live frame ancestry, so they stay executeScript-only.
+    if (__FIREFOX__ && !filter.hasSubframeConstraint()) {
+      scriptletsByWorld[declaredWorld] += `(${func.toString()})(...${JSON.stringify(args)});\n`;
```
seia-soto, line 100: `window.parent` is silently dropped.

chrmod:
closing in favor of https://github.com/ghostery/ghostery-extension/pull/3468

PR #3413, opened 2026-06-11
Show phantom activities in the panel

chrmod:
Screenshot 2026-06-11 at 12 48 00


seia-soto:
Just an idea, I would like to see phantom trackers as a grey color in the trackerdb wheel. They're not actually blocked but subsequently blocked, having less impact in actual.

PR #3423, opened 2026-06-16
feat(custom-filters): add filter lists

### LINKED ISSUE #2296: Adding user’s rules via https link
The feature request is an answer for lack of syncing options, — like custom filters, — in the extension. A message from a user:
> If I can suggest, it would be a solution, if user’s rules could be added via https link (simple txt file, like an AdGuard filter). It would be another great feature.

smalluban:
Adds support for **remote filter lists** — URLs that are fetched, cached, and automatically refreshed — on top of the existing local custom rules, plus a refactor of the whole custom-filters subsystem. Closes #2296.

**Background:** Flat `custom-filters.js` is split into `src/background/custom-filters/{index.js,filter-lists.js,dnr.js}` — `index.js` rebuilds the custom engine + DNR rules from the primary input and cached lists, `filter-lists.js` fetches/caches/parses/refreshes lists, `dnr.js` handles Chromium DNR conversion and limits.

**Store:** `CustomFilters` consolidates everything under a single `customFilters` key (migrated from `customFiltersInput`): primary `text`, a `filterLists` record (per-list `text`, `name`, `lastUpdatedAt`, `expires`, `error` fetch failure, `errors` compilation errors), top-level counts (`networkFilters`, `cosmeticFilters`, `dnrRules`), primary `errors`, and a `userScripts` flag for the state used at the last rebuild. `Options.customFilters` gains a synced `filterLists` record keyed by URL with `enabled` + `trustedScriptlets`.

**Lists:** `! Title:`/`! Expires:` parsed from the header, expiry clamped to 1h–30d (default 1d). Refresh is part of the periodic engine update; only changed lists trigger a rebuild; "Update now" forces a full cache-bypassing refresh; removing a list cleans its cache. The primary input stays authoritative — its errors surface to the user. Each list parses with its own trusted-scriptlets setting, and DNR conversion runs per source so errors attribute to the correct list (or primary), not lumped together.

**DNR (Chromium):** Accepted network filters convert to dynamic rules within the reserved ID range, enforcing dynamic- and regex-rule limits with a clear error when exceeded.

**User Scripts guard (Chromium):** Adds the `userScripts` manifest permission; remote lists only load when "Allow user scripts" is enabled. The UI dims/inerts the lists section with a warning linking to the extension settings, re-checks support on `window` focus, and excludes lists on the next rebuild if user scripts is later disabled (detected via the stored `userScripts` state).

**Settings UI:** Collapsible (`<details>`) error lists for primary input and per-list; "Add a filter list URL" form with invalid/duplicate validation; per-list row with name/URL, last-update (or "Pending update..."), fetch failure, compilation errors, per-list trusted-scriptlets toggle, and remove; usage counters (DNR rules X / max on Chromium, network filters on Firefox, cosmetic filters on both).

**Tests:** E2E test serves a list with a cosmetic + network rule from the local server and verifies adding applies both rules and removing reverts them.

Zrzut ekranu 2026-06-19 o 11 53 25


src/pages/settings/store/filter-list.js
```diff
@@ -0,0 +1,35 @@
...
+import Options from '/store/options.js';
+
+export default {
+  id: true,
+  url: '',
+  [store.connect]: {
+    async set(_, values) {
+      const url = values.url.trim();
+
+      if (!/^https?:/.test(url)) {
```
philipp-classen:
```suggestion
      if (!/^https?:/.test(url) || !URL.canParse(url)) {
```

src/background/custom-filters/filter-lists.js
```diff
@@ -0,0 +1,143 @@
...
+
+  for (const line of text.split(/\r?\n/, METADATA_LINES_LIMIT)) {
+    const match = line.match(/^[!#]\s*(Title|Expires):\s*(.+)$/i);
+    if (!match) continue;
+
+    if (match[1].toLowerCase() === 'title') {
+      result.name = match[2].trim();
+    } else {
+      // Supported formats: "5 days", "12 hours" or a bare number of days
+      const expires = match[2].match(/^(\d+)\s*(day|hour)?/i);
```
philipp-classen: I wonder if being defensive here and matching also week and minute (and either failing or handling it) would make sense. Looks like there is no clear standard.

Currently "1 week" would be parsed as 1 day.

uBlock Origin supports weeks ('w') and also minutes ('m'):
https://github.com/gorhill/uBlock/blob/a0ef1eb795340e66a8116b5c8e3cdfdfa2a229b7/src/js/assets.js#L53

 [I first thought "m" is months, but it is parsed as minutes. 1440 = 24 * 60, which a minimum of 4 hours (240 minutes). "Expires: 3 months" ==> 4 hours (or max(3, 240) seconds)]

In the eyeo documentation (https://developers.eyeo.com/create-a-filter-list), it is day and hours only as I read it (matching the current implementation):
> Expires: Number + optional text
> Update frequency in days (or hours if followed by text starting with 'h')
> ! Expires: 1 days or ! Expires: 7

src/background/custom-filters/filter-lists.js
```diff
@@ -0,0 +1,143 @@
...
+
+export function parseListMetadata(text) {
+  const result = { name: '', expires: DEFAULT_EXPIRES };
+
+  for (const line of text.split(/\r?\n/, METADATA_LINES_LIMIT)) {
+    const match = line.match(/^[!#]\s*(Title|Expires):\s*(.+)$/i);
+    if (!match) continue;
+
+    if (match[1].toLowerCase() === 'title') {
+      result.name = match[2].trim();
```
philipp-classen: To be safe if a custom filter accidentally puts a bad name there (we currently match everything till the new newline character). We could cap it like that:

match[2].trim().slice(0, 100).trim();

... maybe also filter out special characters. Something like that perhaps:

.replace(/[\p{Cc}\p{Cf}]/gu, '')

It is user-configured (i.g. we cannot verify the filter rules) but still untrusted data that we have to process here.

smalluban: Good catch; I will fix it.

src/background/custom-filters/filter-lists.js
smalluban: We can add support for months and minutes, this is no cost, and we spread the supported syntax. The default still goes to 1 day, if the value is not found in the filter list.

smalluban:
By mistake, I merged this PR instead of another one opened in a different tab. I recreated it here: https://github.com/ghostery/ghostery-extension/pull/3441

PR #2135, opened 2024-12-10
fix(telemetry): remove obsolete parameters

smalluban:
Clears out static (hardcoded) and unused parameters.

PR #3427, opened 2026-06-18
feat(telemetry): capture install reason

chrmod:
Tags telemetry pings with the `chrome.runtime.onInstalled` reason (`ir` param) so install events can be reconciled against Chrome Web Store counts, separating genuine fresh installs from storage-reset/bot anomalies.

## Test plan
- Install in a clean profile and complete onboarding.
- In Settings, reveal Developer tools (click the version 5+ times) and confirm **Install reason** shows `install`.

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -91,8 +91,6 @@ chrome.runtime.onInstalled.addListener((details) => {
     if (!runner) return;
 
     await runner.setInstallReason(details.reason);
-    // Refresh so the uninstall URL also carries the reason for churn analysis
-    runner.setUninstallUrl();
   })();
 });
 
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -85,6 +85,17 @@ OptionsObserver.addListener(async function telemetry({ terms, feedback }, lastOp
   }
 });
 
+chrome.runtime.onInstalled.addListener((details) => {
+  (async () => {
+    setup.pending && (await setup.pending);
+    if (!runner) return;
+
+    await runner.setInstallReason(details.reason);
+    // Refresh so the uninstall URL also carries the reason for churn analysis
+    runner.setUninstallUrl();
+  })();
+});
+
 export async function recordSerpVisit() {
   setup.pending && (await setup.pending);
   if (!runner) return;
--- a/src/background/telemetry/metrics.js
+++ b/src/background/telemetry/metrics.js
@@ -127,6 +127,13 @@ export default class Metrics {
     await this.saveStorage(this.storage);
   }
 
+  async setInstallReason(reason) {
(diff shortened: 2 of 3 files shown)


smalluban (approved):

src/background/telemetry/index.js
```diff
@@ -85,6 +85,17 @@ OptionsObserver.addListener(async function telemetry({ terms, feedback }, lastOp
...
 });
 
+chrome.runtime.onInstalled.addListener((details) => {
+  (async () => {
+    setup.pending && (await setup.pending);
+    if (!runner) return;
+
+    await runner.setInstallReason(details.reason);
+    // Refresh so the uninstall URL also carries the reason for churn analysis
+    runner.setUninstallUrl();
```
smalluban: Is the L77 still needed?

src/background/telemetry/metrics.js
```diff
@@ -202,7 +209,9 @@ export default class Metrics {
       // Onboarding complete
       buildQueryPair('oc', this.storage.install_complete_all ? '1' : '0') +
       // Feedback state
-      buildQueryPair('hw', conf.options.terms && conf.options.feedback ? '1' : '0');
+      buildQueryPair('hw', conf.options.terms && conf.options.feedback ? '1' : '0') +
+      // 'na' here flags an install ping with no fresh-install lifecycle event
+      buildQueryPair('ir', this.storage.installReason || 'na');
```
philipp-classen: `ir` has been already used in the past (`install_rand`):

```
install_rand: {
  field: "install_rand",
  alias: "ir",
...
}
```

We removed the parameter 2024-12-11 (https://github.com/ghostery/ghostery-extension/pull/2135). But since there are still old clients and there is old data, perhaps better to use a fresh name.

philipp-classen (approved):

src/background/telemetry/index.js
smalluban: Now I think this line is not required, as we update uninstall URL on every SW restart, as it contains fields that might change over time.

smalluban (changes requested):
We already update uninstall URL on every SW restart.

src/background/telemetry/index.js
smalluban, line 82: Are we sure that we will send this ping with the install reason set?

chrmod: On a genuine fresh install the user must complete onboarding before accepting terms, so `onInstalled` always fires and persists the reason well before `OptionsObserver` triggers the install ping. The guard in `setInstallReason` (`if (storage.install_all || ...)`) also prevents any later update events from overwriting a reason that was already pinged.

src/background/telemetry/index.js
chrmod: I've dropped it

smalluban (approved):

PR #3428, opened 2026-06-18
feat(devtools): add session toggles to disable individual filtering subsystems

chrmod:
Adds per-session dev-menu toggles to disable individual filtering subsystems: network filtering, CSS / scriptlet / extended-CSS cosmetics, anti-tracking, and autoconsent. Overrides are stored in `chrome.storage.session` and reset on browser restart.

## Test plan
- Open Settings → reveal the developer tools section (click the version number repeatedly).
- In the new "Filtering (this session)" card, turn off a toggle and confirm that capability stops applying on a reloaded page (e.g. disable cosmetics and verify hidden elements reappear).
- Restart the browser and confirm all toggles return to enabled.

### DIFF
--- a/src/background/adblocker/cosmetics.js
+++ b/src/background/adblocker/cosmetics.js
@@ -16,6 +16,7 @@ import { FLAG_SUBFRAME_SCRIPTING } from '@ghostery/config';
 import { resolveFlag } from '/store/config.js';
 import Options, { getPausedDetails } from '/store/options.js';
 import DisabledFilters from '/store/disabled-filters.js';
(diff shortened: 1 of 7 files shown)


smalluban (commented):
The logic looks correct. Should we deliver this in the normal build? With the change, there will be another dependency for the cosmetics/network/dnr. 

What about only adding this in DEBUG mode of the extension? So you can run it locally?

chrmod:
Yes, we need it in normal build for sake of debugging actual breakage.

PR #3449, opened 2026-06-26
fix(telemetry): report zap mode

chrmod:
Zap param is used to show special uninstall survey for zap users.


smalluban (approved):

smalluban:
One sentence in the description about why we change would be nice :)

philipp-classen (approved):

PR #3464, opened 2026-07-06
Guarantee document_start scriptlet injection on Chromium via userScripts

chrmod:
On Chromium, scriptlets are injected reactively with `scripting.executeScript`, which cannot guarantee they run before the page's own scripts — a scriptlet like `set-constant` can land too late.

When the "Allow user scripts" permission is available, scriptlets are now registered through `chrome.userScripts` — Chromium's equivalent of Firefox's `browser.contentScripts` — so the browser itself injects them at `document_start` on every navigation. The path mirrors the existing Firefox flow behind the same `contentScripts` interface. Without the permission, injection falls back to the current `executeScript` path unchanged. Firefox is untouched.

Two cases can't be served by a per-hostname registration and stay on per-frame `executeScript`: subframe (`>>`) filters, and local frames (`about:blank`, `srcdoc`), which `chrome.userScripts` cannot reach.

Registrations persist in the browser, so stale ones are dropped on startup, after extension updates, and whenever the filtering engine changes — engine save listeners now fire on every path that changes an engine, including incremental filter updates — and the next navigation re-registers from the current engine. A refresh purge skips hostnames re-registered while it was reading the browser state, so a concurrent navigation cannot lose its fresh registration. Each world's script is registered and removed with its own single-id call, since the batch `userScripts` operations are atomic and reject when any id is unknown — a hostname registered in one world could otherwise never be unregistered, e.g. when its site gets paused.

### Test plan

- On Chromium with "Allow user scripts" enabled, confirm a scriptlet runs before the page's own scripts and exactly once — on first load, after reload, in iframes, and in `about:blank`/`srcdoc` frames.
- With the toggle disabled, confirm scriptlets still inject via the legacy path.
- Pause a site and confirm its scriptlets stop being injected.
- Change or remove a custom scriptlet filter and confirm the first page load afterwards already reflects the change — the old scriptlet no longer runs.
- On Firefox, confirm scriptlet injection is unchanged.


smalluban:
I'll do a review, but this PR looks much riskier than #3467 - I prefer to merge it and release it after the upcoming release.

chrmod:
@smalluban with latest changes it should present no risk until userScripts api is present.

smalluban:
My concern is a scope of the changes in the files. Upcoming release already contains changes in the adblocking logic. It is not about the feature itself. 

This is opt-in feature, which almost no one will have, so I recommend to merge it after 10.5.51.

seia-soto (changes requested):
Other than details, I wonder if extending the Firefox content script approach would be better in style and readability aspect. From what I see, user script and content script APIs are very similar. I think they could be minimally decorated to be compatible in the upper layer and keep all the remaining Firefox code instead of having a fully separate implementation. In this way, we don't implement the scriptlet render step multiple times too.

It also gives me another thinking why Firefox implementation went with Map and Chromium one went with Set. I don't see if one is superior to another but I also don't see any reason to complicate it more. Other comments would be mostly stylistic aspect rather than the API use. For an instance, I don't see many variable names are clear enough. I would say "useSomething" pattern is often used to describe the function rather than the value.

If I refactor this, I would have a shared abstraction of content script and user script API instead of Chromium and Firefox registries. Then move out all the scriptlet compilation/rendering parts from the registry side.

src/background/adblocker/cosmetics.js
```diff
@@ -386,3 +398,11 @@ if (__FIREFOX__) {
...
   );
 }
+
+if (USER_SCRIPTS) {
+  // Registrations persist across restarts, so drop stale ones when the engine changes.
+  const refresh = () => contentScripts.unregisterAll();
+  chrome.runtime.onStartup.addListener(refresh);
+  chrome.runtime.onInstalled.addListener(refresh);
+  engines.addSaveListener(engines.MAIN_ENGINE, refresh);
+}
```
seia-soto, line 407: I am not sure if this path is safe. I see the save listener is only triggered via `saveToMemory` function but the update function in `engines.js` contains a path not calling `saveToMemory` but `saveToStorage`.

chrmod: Good catch — you're right that the diff path in `update()` saves without notifying. In practice the main engine never takes that path: `update()` only runs for the persistent engines, and every main-engine change funnels through `replace()`/`create()` in `reloadMainEngine()`, which do notify. But that safety hangs on the call graph rather than on `addSaveListener` itself, so `update()` now goes through `saveToMemory()` too (ab7d38a0) — listeners fire on every path that changes an engine.

Double-checking this surfaced a real race in the refresh itself: `unregisterAll()` snapshots registrations with `getScripts()`, and a navigation that re-registers a hostname in that window (likely, since a navigation is often what wakes the service worker) would get its fresh registration torn down while `isRegistered()` stays true — silently disabling scriptlets for that hostname for the rest of the worker's lifetime. The purge now skips hostnames re-registered while the snapshot was in flight (8144420d).

chrmod:
> Other than details, I wonder if extending the Firefox content script approach would be better in style and readability aspect. From what I see, user script and content script APIs are very similar. I think they could be minimally decorated to be compatible in the upper layer and keep all the remaining Firefox code instead of having a fully separate implementation. In this way, we don't implement the scriptlet render step multiple times too.
> 
> It also gives me another thinking why Firefox implementation went with Map and Chromium one went with Set. I don't see if one is superior to another but I also don't see any reason to complicate it more. Other comments would be mostly stylistic aspect rather than the API use. For an instance, I don't see many variable names are clear enough. I would say "useSomething" pattern is often used to describe the function rather than the value.
> 
> If I refactor this, I would have a shared abstraction of content script and user script API instead of Chromium and Firefox registries. Then move out all the scriptlet compilation/rendering parts from the registry side.

A full refactor make sense, but I want to do it in a separate PR for sake of a smaller diff. We've already made a lot of significant changes to this codebase in last week.

btw. chrome.userScripts and browser.contentScript, look similar but those are not identical, for example `register` returns different objects - that's why we have a difference in Set vs Map.

src/background/adblocker/content-scripts.js
```diff
@@ -58,3 +58,78 @@ export const contentScripts = (() => {
...
+            });
+        }
+      },
+      isRegistered(hostname) {
+        return registered.has(hostname);
+      },
+      unregister(hostname) {
+        registered.delete(hostname);
+        const ids = ['MAIN', 'ISOLATED'].map((world) => userScriptId(hostname, world));
+        chrome.userScripts.unregister({ ids }).catch(() => {});
```
philipp-classen: Errors are being discarded here, but I think chrome.userScripts.unregister will only deregister if all ids exist. (Otherwise it will throw)

So, if not both MAIN and ISOLATED got registered but only one, it would then not be able to remove it.

src/background/adblocker/cosmetics.js
philipp-classen: This part is unclear to me. In the latest version, this looks like a race:

On a service startup, the engine gets loaded, saved to memory, and then the callback here fires (unregistering all contentScripts).

So, when engine is ready, it will trigger later the registration of the scripts, but also unregisterAll. (Both in different paths).

src/background/adblocker/content-scripts.js
chrmod: right. on it

chrmod: Confirmed — I probed the API from inside the extension on Chrome 151: `unregister({ids})` rejects with `Nonexistent script ID` and removes nothing when any id is missing, and `register`/`update` are just as atomic (a duplicate or unknown id fails the whole batch). Since most hostnames register a single world, per-hostname unregistration was effectively broken — most visibly, pausing a site did not stop its scriptlets, and a hostname whose scriptlets moved between worlds could never refresh its registration.

Fixed in c88b7beb: each world is now registered, updated, and unregistered with its own single-id call, and `register()` also drops a world that no longer has code. The refresh e2e spec gained a pause case built on a single-world registration — it times out waiting for the registration drop without the fix.

src/background/adblocker/cosmetics.js
chrmod: Fixed in b1a53372. The listener also fired when the engine was only *loaded* from storage — that's what ran `unregisterAll` on every worker startup. Now loading from storage doesn't notify listeners, so the refresh happens only when the engine actually changes, and startup no longer unregisters anything.

philipp-classen (approved):

PR #3468, opened 2026-07-07
fix(adblocker): inject Chromium scriptlets once per document

chrmod:
Fixes the Chromium scriptlet double-injection that the tests on `idempotency-tests` document.

On Chromium both `webNavigation.onCommitted` and `webRequest.onResponseStarted` fire for the same document, and each triggers a scriptlet injection with nothing deduping them — so scriptlets ran twice per document. This tracks the documentIds already injected into (using the `documentId` that `chrome.scripting.executeScript` reports back as proof the injection landed, persisted in `chrome.storage.session`) and skips repeats.

Note: a separate, pre-existing race in the Firefox content-script registration path can also double-inject (flaky); it is out of scope here and not addressed by this change.

## Test plan

- Add a custom scriptlet filter and load a matching page on Chrome; the scriptlet applies exactly once (e.g. `rpnt`-rewritten text shows `aaa+`, not `aaa++`), including after a reload.
- The `scriptlet-idempotency` e2e spec passes on Chrome.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/src/background/adblocker/cosmetics.js
+++ b/src/background/adblocker/cosmetics.js
@@ -48,7 +48,32 @@ const scriptletGlobals = {
   warOrigin: chrome.runtime.getURL('/rule_resources/redirects/empty').slice(0, -6),
 };
 
-function injectScriptlets(filters, hostname, details) {
+// On Chromium both webNavigation.onCommitted and webRequest.onResponseStarted fire for the
+// same document, so scriptlets would run twice. We remember the documentIds we have already
+// injected into (proved by executeScript's result) and skip a repeat. The in-memory set gives
+// a synchronous check that closes the race between the two triggers; session storage keeps it
+// across service-worker restarts.
+const INJECTED_DOCUMENTS_KEY = 'injectedScriptletDocuments';
+const injectedDocuments = new Set();
+
+if (__CHROMIUM__) {
+  chrome.storage.session
+    .get(INJECTED_DOCUMENTS_KEY)
+    .then((stored) => {
+      for (const id of stored[INJECTED_DOCUMENTS_KEY] || []) injectedDocuments.add(id);
+    })
+    .catch(() => {});
+}
+
+function rememberInjectedDocument(documentId) {
+  injectedDocuments.add(documentId);
+  if (injectedDocuments.size > 1000) {
+    injectedDocuments.delete(injectedDocuments.values().next().value);
+  }
+  chrome.storage.session.set({ [INJECTED_DOCUMENTS_KEY]: [...injectedDocuments] }).catch(() => {});
+}
+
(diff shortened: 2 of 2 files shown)


seia-soto:
I thought that the reason we have multiple event listeners is because the flakiness of `executeScript` API. If we're going to do this, wouldn't removing the event listener be better? I thought that `onResponseStarted` was always faster.

chrmod:
> onResponseStarted

`onResponseStarted` is indeed always faster, but it does not fire when page is loaded from cache - we need to keep both listeners.

smalluban (commented):
Do we have to keep the ids in the session storage? The `onCommited` and `onResponseStarted` for the `main_frame` will always trigger within the same SW session. Keeping IDs only in memory should be sufficient.

seia-soto:
@smalluban I also wondered the same point.

I personally thought that it could be benefitial to the already opened tab, but we only inject the scriptlets once the navigation happens (which will always trigger the event listener).

I am not sure about other use cases.

seia-soto (commented):
We can change the logic into the below steps:

1. If `onResponseStarted` fires first and there's no `documentId` saved, execute script and save `documentId`.
2. If `onCommitted` fires first and there's no `documentId` saved, assume that `onResponseStarted` didn't fire at all and do not save `documentId`.
3. If `onCommitted` fires later (we observe the same `documentId` from the storage), assume `onResponseStarted` did the job and remove the `documentId` from the storage.

This way, the implementation could be more straightforward.

.github/workflows/test.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
```
philipp-classen, line 7: Is this removal intended?

chrmod: yes, that allow us to run tests on PRs that target different branch than `main`

src/background/adblocker/cosmetics.js
```diff
@@ -90,24 +124,36 @@ function injectScriptlets(filters, hostname, details) {
...
+
+  // The documentId echoed back by executeScript proves the injection landed; persist it so
+  // the dedup outlives a service-worker restart. If nothing landed, release the reservation.
+  const results = await Promise.all(injections);
+  const injectedDocumentId = results.flat().find((result) => result?.documentId)?.documentId;
+
+  if (injectedDocumentId) {
+    rememberInjectedDocument(injectedDocumentId);
+  } else if (details.documentId) {
+    injectedDocuments.delete(details.documentId);
```
philipp-classen, line 156: Looks like an error case on Chrome (injection failing). Perhaps we should issue an warning here?

(Chromium-only likely, because Firefox will always hit it, as I understand.)

*Update:* Though I see that we are already warning above (in the catch block). Maybe that is enough.

.github/workflows/test.yml
smalluban: It must have been because of the target branch of this PR. Can you clean it up, @chrmod?

chrmod: i think it should stay

philipp-classen: Ah, you mean examples like we had when we refined a PR by creating another PR on topic of it (instead of trying to explain as review comments)? Okay, then I'm fine with keeping the change.

smalluban (approved):

philipp-classen (approved):

PR #3471, opened 2026-07-08
chore(scripts): detect permission changes requiring user consent

smalluban:
When a published extension update adds a permission that triggers an install-time warning, Chrome disables the extension for every user until they manually re-approve it, silently breaking the extension for the whole user base. The `wdio:update` e2e test cannot catch this, because it loads an unpacked extension in developer mode where Chrome auto-grants new permissions on reload — the regression is only visible in the packed/Web Store update flow.

This adds a static `scripts/check-permissions.js` check, wired up as `lint:permissions` and included in the `lint` script. It diffs the `permissions` and `host_permissions` of the current `src/manifest.*.json` against the last released tag and fails if any newly added permission is known to require user consent (e.g. `bookmarks`, `history`, `downloads`, `management`), or if host permissions are broadened. Non-warning permission additions are reported informationally without failing the build.


philipp-classen (approved):

chrmod (approved):
I don't mind but this list changes in time so its usefulness may degrade over time. I would rather insist we test manually each time we change permission list. Which should be rare.

smalluban:
I know the limitations, but there is no way we can do it as the e2e test - the extension must be installed as a signed packaged version; otherwise, the prompt will not be shown. I expect that the list of permissions does not change often. 

We should check permissions manually, but this can give us a hint that we try to add risky code before we merge it into main.

PR #3472, opened 2026-07-08
Cross-platform subframe scriptlet injection

chrmod:
## Goal

Make subframe-targeting scriptlets (`domain>>##+js(...)`) behave identically on Firefox and Chromium (both with and without the userScripts permission). Previously they only reached **cross-origin** child frames on Chromium's `executeScript` path — on Firefox and the Chromium userScripts path they silently never ran, which meant enabling **"Allow user scripts"** on Chrome dropped subframe-scripting support that the fallback path had.

## What changes

Subframe scriptlet delivery is now split by the decision it actually requires:

- **Direct-domain** scriptlets keep going through the content-script registry (Firefox `contentScripts` / Chromium `userScripts`), preserving `document_start` injection.
- **Subframe-constrained** scriptlets are injected per-frame from the real frame ancestor chain on **every** platform. Only a per-frame decision can inject into a cross-origin child because *its parent* matched — a per-hostname registration structurally cannot.

Alongside the main fix:

- **Mixed direct/subframe filters** (`a.com,b.com>>##+js(...)`) now also run on the top `a.com` frame, which the old per-frame `window.parent` guard wrongly suppressed.
- **Firefox registration** now matches the exact hostname instead of `*.hostname`. The wildcard also covered subdomain child frames, which register under their own hostname too, so the same scriptlet was injected twice into a subdomain frame.

## Behavior notes

- On Firefox, same-origin subframe scriptlets shift from registration to per-frame injection, so their timing moves from `document_start` to just after the child frame commits. This is the trade-off for correct, uniform cross-origin behavior; the affected feature is gated behind the subframe-scripting flag.

## Test plan

With custom **trusted** scriptlet filters enabled, on Firefox and on Chrome (both with and without **Allow user scripts**):

- Add `page.localhost>>##+js(set, __probe__, true)` and open a page that embeds a **cross-origin** iframe. Confirm the scriptlet ran inside the iframe and did **not** run on the top page.
- Confirm the same scriptlet also runs inside a **same-origin** iframe.
- Add a direct-domain scriptlet (e.g. `page.localhost##+js(set-constant, ...)`) and open a page that embeds a **subdomain** iframe (e.g. `sub.page.localhost`). Confirm the scriptlet takes effect exactly once in that iframe (no doubled effect).

### DIFF
--- a/src/background/adblocker/content-scripts.js
+++ b/src/background/adblocker/content-scripts.js
@@ -24,7 +24,10 @@ const firefoxRegistry = (() => {
             await browser.contentScripts.register({
               js: [{ code }],
               allFrames: true,
-              matches: [`https://*.${hostname}/*`, `http://*.${hostname}/*`],
+              // Match the exact hostname (like the Chromium userScripts registry). A `*.` wildcard
+              // would also match subdomain frames, which self-register under their own hostname —
+              // the overlap injects the same scriptlet twice into a subdomain child.
+              matches: [`https://${hostname}/*`, `http://${hostname}/*`],
               matchAboutBlank: true,
               matchOriginAsFallback: true,
               runAt: 'document_start',
--- a/src/background/adblocker/content-scripts.js
+++ b/src/background/adblocker/content-scripts.js
@@ -24,9 +24,7 @@ const firefoxRegistry = (() => {
             await browser.contentScripts.register({
               js: [{ code }],
               allFrames: true,
-              // Match the exact hostname (like the Chromium userScripts registry). A `*.` wildcard
-              // would also match subdomain frames, which self-register under their own hostname —
-              // the overlap injects the same scriptlet twice into a subdomain child.
+              // Exact hostname: a `*.` wildcard would double-inject into subdomain frames that self-register.
               matches: [`https://${hostname}/*`, `http://${hostname}/*`],
               matchAboutBlank: true,
               matchOriginAsFallback: true,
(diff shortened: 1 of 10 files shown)


smalluban:
Does this conflict with https://github.com/ghostery/ghostery-extension/pull/3464?

src/background/adblocker/content-scripts.js
```diff
@@ -24,7 +30,8 @@ const firefoxRegistry = (() => {
             await browser.contentScripts.register({
               js: [{ code }],
               allFrames: true,
-              matches: [`https://*.${hostname}/*`, `http://*.${hostname}/*`],
+              // Exact hostname: a `*.` wildcard would double-inject into subdomain frames that self-register.
+              matches: [`https://${hostname}/*`, `http://${hostname}/*`],
```
seia-soto, line 34: Leaving a note here, this works because the main frame scripts can be pinned in content/user script api not the subframe scripts. However, I don't know if we should handle having same scriptlets over `domain.tld` and `domain.tld>>` matchers.

src/background/adblocker/content-scripts.js
```diff
@@ -60,15 +61,13 @@ const firefoxRegistry = (() => {
...
-// Registrations persist across service-worker restarts, so register() upserts by a stable id
-// and unregisterAll() queries the browser to purge them (the map is empty after a restart).
+// userScripts registrations persist across restarts; upsert by a stable id, purge via getScripts().
 const USER_SCRIPTS_NAMESPACE = 'ghostery-scriptlet';
 
 function userScriptId(hostname, world) {
   return `${USER_SCRIPTS_NAMESPACE}:${world}:${hostname}`;
 }
 
-// __CHROMIUM__ guard so chrome.userScripts is tree-shaken from the Firefox build.
```
seia-soto, line 71: Is there any reason to drop this comment?

seia-soto: This is not really relevant and seems like unnecessary comment from previous change tho.

chrmod:
Superseded: the direct/subframe split landed via #3476 (per-frame >> injection with real frame ancestors) and the userScripts-path integration was folded into the main merge on the base branch (474f351a), so >> filters keep the per-frame executeScript path on every platform. Branch kept for reference.

PR #3476, opened 2026-07-09
fix(adblocker): subframe scriptlet injection on Firefox

chrmod:
## What was wrong

On Firefox, subframe-targeting scriptlets (`domain>>##+js(...)`) were broken:

- They never ran in **cross-origin** child frames.
- Direct-domain scriptlets were injected **twice** into subdomain frames.

Both now behave correctly, matching Chromium. The Chromium path is unchanged.

## Test plan

With custom **trusted** scriptlet filters enabled, on Firefox:

- Add `page.localhost>>##+js(set, __probe__, true)` and open a page embedding a **cross-origin** iframe — the scriptlet runs inside the iframe and **not** on the top page.
- Confirm it also runs inside a **same-origin** iframe.
- Add `page.localhost##+js(set-constant, ...)` and open a page embedding a **subdomain** iframe (`sub.page.localhost`) — the scriptlet takes effect exactly once (no doubled effect).

### DIFF
--- a/src/background/adblocker/content-scripts.js
+++ b/src/background/adblocker/content-scripts.js
@@ -15,7 +15,11 @@ export const contentScripts = (() => {
     async register(hostname, scriptletsByWorld) {
       this.unregister(hostname);
 
+      // Reserve synchronously so a concurrent onCommitted register() short-circuits on
+      // isRegistered() instead of registering the hostname again and injecting twice.
       const registered = [];
+      map.set(hostname, registered);
+
       try {
         for (const [world, code] of Object.entries(scriptletsByWorld)) {
           if (!code) continue;
@@ -24,7 +28,8 @@ export const contentScripts = (() => {
             await browser.contentScripts.register({
               js: [{ code }],
               allFrames: true,
-              matches: [`https://*.${hostname}/*`, `http://*.${hostname}/*`],
+              // Subdomain frames register their own scriptlets, so match only this exact hostname.
+              matches: [`https://${hostname}/*`, `http://${hostname}/*`],
               matchAboutBlank: true,
               matchOriginAsFallback: true,
               runAt: 'document_start',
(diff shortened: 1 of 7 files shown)


src/background/adblocker/content-scripts.js
```diff
@@ -24,7 +28,8 @@ export const contentScripts = (() => {
             await browser.contentScripts.register({
               js: [{ code }],
               allFrames: true,
-              matches: [`https://*.${hostname}/*`, `http://*.${hostname}/*`],
+              // Exact hostname: a `*.` wildcard would double-inject into subdomain frames that self-register.
```
smalluban: This comment explains the change in the code, not the current status of the code below. I am not sure if this will be helpful for a reader in the future.

src/background/adblocker/cosmetics.js
```diff
@@ -114,16 +148,6 @@ async function injectScriptlets(filters, hostname, details) {
...
-      if (filter.hasSubframeConstraint()) {
-        code += `window.parent!==window&&`;
-      }
-      code += `(${func.toString()})(...${JSON.stringify(args)});\n`;
-      scriptletsByWorld[declaredWorld] += code;
-      continue;
-    }
-
     injections.push(
       chrome.scripting
```
smalluban, line 115: It looks like this code now also runs on Firefox - before, there was a `continue` statement, which finished the loop.

src/background/adblocker/cosmetics.js
```diff
@@ -140,14 +164,12 @@ async function injectScriptlets(filters, hostname, details) {
     );
   }
 
-  if (__FIREFOX__) {
-    contentScripts.register(hostname, scriptletsByWorld);
-    return;
-  }
+  const results = await Promise.all(injections);
+
+  if (__FIREFOX__) return;
```
smalluban: As Firefox uses the `contentScripts` API, I would split the registration and inject logic completely - and create a new high-level function; it can still be `injectScriptlets`, but for chromium would be `executeScripts`.

Returning in the middle of the function just because of the platform makes it hard to reason.

smalluban (changes requested):

smalluban (approved):

philipp-classen (approved):

PR #3490, opened 2026-07-13
GPC support: set "navigator.globalPrivacyControl" to true

philipp-classen:
GPC support: set "navigator.globalPrivacyControl" on Firefox and Chromium

Background:
https://w3c.github.io/gpc/#javascript-property-to-detect-preference

We already set the Sec-GPC signal, but the spec requires also to have navigator.globalPrivacyControl be truthy before the page runs.

Test pages:
- https://globalprivacycontrol.org/
- https://browserleaks.com/donottrack

### DIFF
--- a/src/content_scripts/gpc.js
+++ b/src/content_scripts/gpc.js
@@ -1,0 +1,30 @@
+/**
+ * Ghostery Browser Extension
+ * https://www.ghostery.com/
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+// Ensures that `navigator.globalPrivacyControl` is truthy (GPC support):
+// https://w3c.github.io/gpc/#javascript-property-to-detect-preference
+//
+// Injected into the page's MAIN world at document_start (including all frames).
+(function () {
+  try {
+    if (!window.navigator.globalPrivacyControl) {
+      Object.defineProperty(navigator, 'globalPrivacyControl', {
+        get() {
+          return true;
+        },
+        configurable: true,
+        enumerable: true,
+      });
+    }
+  } catch {
+    // ignore
+  }
+})();
--- a/src/manifest.chromium.json
+++ b/src/manifest.chromium.json
@@ -88,6 +88,14 @@
       "all_frames": true
     },
     {
+      "js": ["content_scripts/gpc.js"],
+      "matches": ["http://*/*", "https://*/*"],
+      "run_at": "document_start",
+      "match_about_blank": true,
+      "all_frames": true,
+      "world": "MAIN"
+    },
+    {
       "js": ["content_scripts/notifications.js"],
       "matches": ["http://*/*", "https://*/*"],
       "run_at": "document_start"
--- a/src/manifest.firefox.json
+++ b/src/manifest.firefox.json
@@ -61,6 +61,14 @@
       "all_frames": true
     },
     {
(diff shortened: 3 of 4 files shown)


src/content_scripts/gpc.js
```diff
@@ -0,0 +1,30 @@
...
+// Ensures that `navigator.globalPrivacyControl` is truthy (GPC support):
+// https://w3c.github.io/gpc/#javascript-property-to-detect-preference
+//
+// Injected into the page's MAIN world at document_start (including all frames).
+(function () {
+  try {
+    if (!window.navigator.globalPrivacyControl) {
+      Object.defineProperty(navigator, 'globalPrivacyControl', {
+        get() {
+          return true;
```
smalluban, line 21: The value is not always `true` - it should be taken from `options.autoconsent.gpc`

smalluban (changes requested):

src/content_scripts/gpc.js
```diff
@@ -0,0 +1,30 @@
+/**
```
smalluban, line 1: Use the already existing `content_script/autoconsent.js` instead of creating a new content script.

smalluban:
@philipp-classen In general, I like the idea - we can add support for JS land for GPC where it is not yet supported.

However, it won't behave as you expect. In Firefox, the value returns what the user set in the browser settings, not in the extension itself. If Chrome or Safari adds support, but the user doesn't set the value in the browser, it will be reverted to `false`.

src/content_scripts/gpc.js
philipp-classen: But if we have any roundtrip, we lose the property of running first. Then it will not work. Supporting out-out is not possible without breaking the functional, in my understanding.

My assumption is that if someone decides to install Ghostery, the intent is to be tracked and not have their personal data being sold. And the GPC is the formal way to express that.

philipp-classen:
> @philipp-classen In general, I like the idea - we can add support for JS land for GPC where it is not yet supported.
> 
> However, it won't behave as you expect. In Firefox, the value returns what the user set in the browser settings, not in the extension itself. If Chrome or Safari adds support, but the user doesn't set the value in the browser, it will be reverted to `false`.

Not according to my tests. One test page is these, which on Firefox and Chromium it is detected (both without Ghostery or with the current production version):
https://globalprivacycontrol.org/

Also, just debugging into the page with devtools shows the overwritten value. I do not see how Firefox reverts it. For me, it stays true independent of the GPC settings in Firefox.

src/content_scripts/gpc.js
smalluban: I understand the technical limitation on overwriting this API; however, we provide a way to enable or disable GPC, so we have to follow the user's choice. 

I prefer not to provide an overwrite on this rather than always setting it to `true`.

smalluban:
I missed that you overwrite if it is undefined OR FALSE. Still, I am not sure if we should do it that way. According to the MDN documentation and my testing, Chrome does not support it, as it also does not have an option for that in settings. `"globalPrivacyControl" in navigator` returns `false` for Chrome.

src/content_scripts/gpc.js
seia-soto: Or can we handle this like scriptlets (via content scripts or scripting API)?

philipp-classen: Interesting that you mention it, was just about to write, because Krzysztof also made me aware that content scripts can be used here:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/registerContentScripts

Switching to draft now. I think, that should still work and be configurable in the UI.

philipp-classen:
Closing in favor of https://github.com/ghostery/ghostery-extension/pull/3529

It now integrates with the existing UI toggle. The race issues is avoided by using chrome.scripting.registerContentScripts.

PR #3495, opened 2026-07-14
fix(telemetry): keep sending pings when the insights database is unavailable

### LINKED ISSUE #3494: Telemetry pings silently dropped on desktop Firefox since v10.5.42
## Summary

Since **v10.5.42**, `active` and `engaged` telemetry pings are silently dropped on a subset of **desktop Firefox** profiles. Telemetry shows ~7% of desktop-Firefox DAU (Windows/Mac/Linux, uniform) went silent from mid-May and never recovered; Chrome/Edge/Safari and Firefox-Android are unaffected. Tracker blocking is **not** affected — this is a metrics-only regression (affected users are still active, just uncounted).

## Cause

#3331 added a `pv` ping param, which put an unguarded IndexedDB read on the ping's critical path in `src/background/telemetry/index.js` (`getConf`):

```js
const [options, config, dailyStats] = await Promise.all([
  store.resolve(Options),
  store.resolve(Config),
  store.resolve(DailyStats, yesterdayId),   // rejects if the `insights` DB fails to open
]);
```

`store.resolve` **rejects** on a failed DB open. Long-lived (pre-v10) desktop-Firefox profiles hit a throwing branch in the `daily-stats.js` v31 `upgrade()` (`deleteObjectStore('search'/'tabs')`, or missing `daily` store). The rejection flows through `getConf` → `_buildMetricsUrl` and is swallowed by the `.catch` in `_recordActive`/`_recordEngaged`, so both pings are dropped — and since the timestamps are already advanced, there's no retry until the next window, where it fails again. Fresh installs create a clean v31 DB and are unaffected.

## Fix

Make the `DailyStats` read non-fatal: wrap it in try/catch and default `yesterdayPages` to `0` (→ `pv=0`) instead of letting it reject `getConf`. Logging the caught IndexedDB error name would confirm the exact failure on affected profiles.

chrmod:
## What

`getConf` (telemetry) turned yesterday's page-view bucket — the `pv` param added in #3331 — into a hard dependency of every ping by awaiting `store.resolve(DailyStats, …)` unguarded. When a profile's `insights` IndexedDB fails to open, that resolve rejects, propagates through `_buildMetricsUrl`, and is swallowed by the ping's `.catch`, so `active` and `engaged` pings are silently dropped and never retried.

This makes the `DailyStats` read non-fatal — on failure it falls back to `pages: 0` (→ `pv=0`) instead of rejecting `getConf`.

Fixes #3494.

## Impact

Restores telemetry pings for the affected profiles — primarily long-lived desktop-Firefox installs whose legacy `insights` DB hits a throwing branch in the v31 `upgrade()`. Tracker blocking is not affected.

## Test plan

- In Firefox, browse a few pages and confirm the telemetry ping to `d.ghostery.com` still fires (Network tab), now including the `pv` param.
- Force `store.resolve(DailyStats, …)` to reject (e.g. corrupt or close the `insights` DB): confirm `active` / `engaged` pings are still sent with `pv=0` rather than being dropped.

### DIFF
--- a/src/background/telemetry/index.js
+++ b/src/background/telemetry/index.js
@@ -42,7 +42,8 @@ const setup = asyncSetup('telemetry', [
         const [options, config, dailyStats] = await Promise.all([
           store.resolve(Options),
           store.resolve(Config),
-          store.resolve(DailyStats, yesterdayId),
+          // A failed `insights` DB read must not block the ping.
+          store.resolve(DailyStats, yesterdayId).catch(() => ({ pages: 0 })),
         ]);
 
         return {


src/background/telemetry/index.js
```diff
@@ -42,7 +42,8 @@ const setup = asyncSetup('telemetry', [
         const [options, config, dailyStats] = await Promise.all([
           store.resolve(Options),
           store.resolve(Config),
-          store.resolve(DailyStats, yesterdayId),
+          // A failed `insights` DB read must not block the ping.
+          store.resolve(DailyStats, yesterdayId).catch(() => ({ pages: 0 })),
```
smalluban, line 46: I was rather thinking of adding a guard inside the `DailyStats` model, so it returns default values if the DB connection fails.

The same fix should be here instead: https://github.com/ghostery/ghostery-extension/blob/main/src/store/daily-stats.js#L93

smalluban (changes requested):

src/background/telemetry/index.js
smalluban: After looking at other problems, it is better to add a guard here, so that we can detect it and set the value to `-1`. The https://github.com/ghostery/ghostery-extension/pull/3497 replaces this PR.

PR #3497, opened 2026-07-15
fix(telemetry): SERP counting with page views

smalluban:
The `se` (SERP visits) and `pv` (page views) telemetry signals were counted through separate paths, so they could diverge within a day — SERP visits were bumped in `serp.js` on navigation regardless of incognito state, while page views were counted only when flushing tab stats.

- SERP visit counting is moved out of `serp.js` into the tab-stats flush (`flushTabStatsToDailyStats`) via the shared, now-exported `SERP_URL_REGEXP`, so `se` and `pv` share the same timing, UTC-day attribution, and incognito exclusion.
- Incognito state is resolved once in the `onCommitted` listener while the tab is still alive and cached on the stats object, so it is reliable even when flushing on tab close.
- Page views are now counted regardless of whether trackers were seen (the patterns list is only extended when there are trackers).

### DIFF
--- a/src/background/serp.js
+++ b/src/background/serp.js
@@ -16,7 +16,6 @@ import trackersPreviewCSS from '/content_scripts/trackers-preview.css?raw';
 import Options, { isGloballyPaused } from '/store/options.js';
 import { getWTMStats } from '/utils/wtm-stats.js';
 import { parseWithCache } from '/utils/request.js';
-import { recordSerpVisit } from './telemetry/index.js';
 
 // Trackers preview messages
 chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
@@ -40,14 +39,12 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
   return false;
 });
 
-const SERP_URL_REGEXP =
+export const SERP_URL_REGEXP =
   /^https?:[/][/][^/]*[.](google|bing)[.][a-z]+([.][a-z]+)?([/][a-z]+)*[/]search/;
 
 // SERP tracking prevention and trackers preview content scripts
 chrome.webNavigation.onCommitted.addListener(async (details) => {
   if (details.url.match(SERP_URL_REGEXP)) {
-    recordSerpVisit();
-
     const options = await store.resolve(Options);
 
     if (options.wtmSerpReport) {
--- a/src/background/stats.js
+++ b/src/background/stats.js
@@ -25,7 +25,10 @@ import { getMetadata, getUnidentifiedTracker } from '/utils/trackerdb.js';
 import Request from '/utils/request.js';
 import { isOpera, isWebkit } from '/utils/browser-info.js';
 
+import { recordSerpVisit } from './telemetry/index.js';
+
 import * as logger from './logger.js';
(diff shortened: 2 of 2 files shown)


smalluban:
We should avoid querying the Tabs API for each navigation. We need to investigate other possibilities. In the meantime, the fix for the broken DB is here: https://github.com/ghostery/ghostery-extension/pull/3513

smalluban:
**Design note for reviewers:** why this PR resolves a tab's `incognito` state with `chrome.tabs.get()` in `onCommitted`, rather than caching it from `chrome.tabs.onUpdated`.

**Idea considered:** listen to `chrome.tabs.onUpdated`, capture each tab's `incognito` value into an in-memory cache, and read it synchronously when the per-tab stats object is created in `setupTabStats` (called from `webNavigation.onCommitted`). The `incognito` flag gates whether page views / SERP visits get written to daily stats, and since `incognito` is a per-tab constant it looks cacheable. In practice this approach has several issues:

1. **It doesn't actually reduce IPC.** The events are themselves IPC pushed from the browser process, and `onUpdated` is chatty — it fires for `status` (loading/complete), title, favicon, `audible`, `mutedInfo`, etc. That's typically *more* messages than a single targeted `chrome.tabs.get()` per navigation. In MV3 each event can also wake the service worker, so the listener isn't free. The only real gain is a synchronous read at point-of-use, not less IPC.

2. **Event ordering vs. `onCommitted` is not guaranteed.** The stats object is created on `webNavigation.onCommitted`. There's no ordering guarantee between `tabs.onUpdated {status:'loading'}` and `onCommitted`, so for a brand-new tab the cache may not be populated yet when the stats object is created → cache miss.

3. **MV3 service-worker restarts wipe the in-memory cache.** Chromium runs the background as a service worker that gets torn down/restarted. A plain `Map` is lost on restart; even a storage-backed map hydrates *asynchronously*, so it can still miss right after wake. On a miss, `incognito` is `undefined` (falsy), the daily-stats flush won't bail, and **incognito page views / SERP visits get recorded — a privacy leak.**

4. **You need a `chrome.tabs.get()` fallback anyway.** Because of #2 and #3, a cache miss must fall back to the authoritative async lookup. So the round-trip doesn't disappear — it just becomes less frequent — at the cost of more moving parts for the same worst-case behavior.

5. **tabId reuse.** Chrome reuses tab IDs after tabs close. Without cleanup on `chrome.tabs.onRemoved`, a reused ID can inherit a stale `incognito: true` from a previous incognito tab. Cleanup on removal is mandatory.

6. **SPA / history navigations don't line up.** `history.pushState` navigations fire `tabs.onUpdated` (with `changeInfo.url`) but not `webNavigation.onCommitted` (they fire `onHistoryStateUpdated`). The two event streams don't match the stats lifecycle.

**Bottom line:** `onUpdated` caching is a latency/critical-path micro-optimization (synchronous read, no await), not an IPC or correctness win. The cache can't be trusted alone (cold start after SW restart, ordering, tabId reuse), so a `chrome.tabs.get()` fallback is still required. Given the current `chrome.tabs.get()` on the once-per-navigation path is already correct and cheap, the added complexity isn't worth it.

chrmod (approved):

PR #3499, opened 2026-07-15
fix(adblocker): reliably inject subframe scriptlets into cross-origin frames on Firefox

smalluban:
A subframe scriptlet e2e test was flaky on Firefox CI: subframe-constrained (`>>`) scriptlets intermittently failed to inject into cross-origin child frames. On Firefox these frames are out-of-process (Fission), and the document present at `webNavigation.onCommitted` is provisional and gets replaced — so `chrome.scripting.executeScript` resolves successfully but the injected scriptlet is discarded with the throwaway document, permanently missing that frame.

This defers Firefox subframe (`>>`) scriptlet injection from `onCommitted` to the content script's `injectCosmetics` bootstrap message, which runs at `document_start` inside the frame's final document, so the injection reliably lands.

- On Firefox, scriptlets without a subframe constraint keep the direct-domain content-script registration; only `>>`-constrained ones are injected per-frame, and only once the bootstrap message arrives from the frame's final document.
- The bootstrap message sender carries no `parentFrameId`, so the `>>` constraint is evaluated against the frames hierarchy the background already tracks: on Firefox, `injectCosmetics` resolves the sender's ancestor chain with `ancestorsOf()` when the bootstrap message arrives.
- `FramesHierarchy` now separates writes from reads: `updateAncestry()` only maintains the hierarchy — replacing the previous `ancestors()` call that updated it and returned the chain in one step — and `ancestorsOf()` is the single way to resolve a chain on both platforms. The recovery behavior where a frame pointing at an unknown ancestor drops the whole tab is preserved and now covered by a unit test.
- Simplified the e2e probe back to a single iframe (removing the earlier retry/recreate workaround) now that first-frame injection is reliable.

Subframe filter evaluation stays behind the `SUBFRAME_SCRIPTING` flag and Chromium injection behavior is unchanged. Verified with 3 consecutive Firefox runs of the subframe-scripting spec — all 4 tests pass with 0 retries.

## Test plan

- On Firefox, open a page embedding a cross-origin iframe targeted by a `>>` scriptlet filter and confirm the scriptlet takes effect inside the iframe on first load.
- Reload the page several times and confirm the scriptlet keeps applying, with no intermittent misses.
- On Chromium, open the same page and confirm subframe scriptlets and styles behave as before.

### DIFF
--- a/src/background/adblocker/cosmetics.js
+++ b/src/background/adblocker/cosmetics.js
@@ -73,7 +73,7 @@ function rememberInjectedDocument(documentId) {
   chrome.storage.session.set({ [INJECTED_DOCUMENTS_KEY]: [...injectedDocuments] }).catch(() => {});
 }
 
-async function injectScriptlets(filters, hostname, details, frameReady = false) {
+async function injectScriptlets(filters, hostname, details) {
   if (__CHROMIUM__) {
     if (filters.length === 0) return;
     if (details.documentId) {
@@ -111,15 +111,6 @@ async function injectScriptlets(filters, hostname, details, frameReady = false) 
       continue;
     }
 
-    // On Firefox, injecting a subframe-constrained (>>) scriptlet at webNavigation.onCommitted is
-    // unreliable for out-of-process (cross-origin) child frames: the committed document is still
-    // provisional and gets replaced, so executeScript resolves but its effect is discarded. We
-    // defer these injections to the content script's bootstrap message, which runs inside the
-    // frame's final document (frameReady), guaranteeing the scriptlet lands.
-    if (__FIREFOX__ && !frameReady) {
-      continue;
-    }
-
     injections.push(
       chrome.scripting
         .executeScript({
@@ -137,17 +128,12 @@ async function injectScriptlets(filters, hostname, details, frameReady = false) 
   }
 
   if (__FIREFOX__) {
-    // Content-script registration for direct-domain scriptlets is handled at onCommitted
-    // (document_start). The frameReady path only performs the deferred subframe executeScript
-    // above, so it must not touch the per-hostname registration state.
-    if (!frameReady) {
-      if (scriptletsByWorld.MAIN || scriptletsByWorld.ISOLATED) {
-        if (!contentScripts.isRegistered(hostname)) {
-          contentScripts.register(hostname, scriptletsByWorld);
-        }
-      } else {
-        contentScripts.unregister(hostname);
+    if (scriptletsByWorld.MAIN || scriptletsByWorld.ISOLATED) {
+      if (!contentScripts.isRegistered(hostname)) {
+        contentScripts.register(hostname, scriptletsByWorld);
       }
+    } else {
+      contentScripts.unregister(hostname);
     }
     return;
   }
@@ -206,7 +192,7 @@ framesHierarchy.handleWebextensionEvents();
  * (like registering content scripts for scriptlet filters on Firefox)
  */
 async function injectCosmetics(details, config) {
-  const { bootstrap: isBootstrap = false, scriptletsOnly, frameReady = false } = config;
+  const { bootstrap: isBootstrap = false, scriptletsOnly } = config;
 
   try {
     setup.pending && (await setup.pending);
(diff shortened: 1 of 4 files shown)


chrmod (approved):

src/background/adblocker/cosmetics.js
```diff
@@ -231,12 +231,17 @@ async function injectCosmetics(details, config) {
...
-    ancestors = framesHierarchy.ancestors(
+    const chain = framesHierarchy.ancestors(
       { tabId, frameId, parentFrameId, documentId },
       { domain, hostname },
     );
+
+    // On Firefox the chain is consumed by the bootstrap-message pass instead (see ancestorsOf).
+    if (!__FIREFOX__) {
+      ancestors = chain;
+    }
```
philipp-classen: This code is confusingly written. Why not something like that?

```suggestion
    if (!__FIREFOX__) {
      ancestors = framesHierarchy.ancestors(
        { tabId, frameId, parentFrameId, documentId },
        { domain, hostname },
      );
    }
```

chrmod: Good catch on the clarity — but the call can't move inside the `!__FIREFOX__` block. `framesHierarchy.ancestors()` isn't a pure getter: it also *records* the current frame in the hierarchy (registers/updates it in `this.tabs`). Firefox depends on that side effect running, because it reads the ancestor chain back later from the bootstrap message via `ancestorsOf()` (line ~366). Skipping the call on Firefox would leave the frame unregistered and break subframe ancestor resolution there.

Only the *return value* is platform-specific, so that's the part guarded by `!__FIREFOX__`. I've reworded the comment to make the side-effect intent explicit instead.

seia-soto:
See WHATWG HTML5 standards: 7.3.2 Browsing contexts: https://html.spec.whatwg.org/#windows, it mentions: `In general, there is a 1-to-1 mapping from the Window object to the Document object, as long as the Document object has a non-null browsing context. There is one exception. A Window can be reused for the presentation of a second Document in the same browsing context, such that the mapping is then 1-to-2. This occurs when a browsing context is navigated from the initial about:blank Document to another, which will be done with replacement.`.

If the `window` is reused, scriptlet context would not be invalidated after document replacement. It's not valid that all scriptlets are discarded unlike how it's demonstrated. For an instance, `set-constant` and all scriptlets not depending on document would run just fine regardless.

seia-soto:
Update: I could not see any critical sign that the scriptlet injection in subframes are delayed. I used the below test materials:

```
    <script>
      function subframeFunc() {
        __ghostery_subframe__ = true;
      }

      subframeFunc();
    </script>
```

```
      it.only('subframe of the page', async function () {
        await setCustomFilters([`${PAGE_DOMAIN}>>##+js(set, subframeFunc, trueFunc)`]);
        await browser.url(PAGE_URL);
        await switchFrame($('#iframe-static'));
        const val = await browser.execute(function () {
          return window.__ghostery_subframe__;
        });

        await browser.debug();

        await expect(val).not.toBe(true);
        await browser.switchFrame(null);
      });
```

seia-soto:
I picked up 5b6389afa1f3a9a3d8977fdd18178c00d8ecb6e3 this commit before running any E2E tests.

seia-soto:
I also suggest adding a non-DOM related scriptlet testing apart from rpnt.

philipp-classen (approved):

chrmod:
**Window reuse:** reuse from the initial `about:blank` requires the incoming document to be same-origin with it (it inherits the parent's origin). A cross-origin child fails that, so it gets a fresh `Window` — and a separate process under Fission. Nothing for `set` to have written to survives. That said, you're right the description overstates it: the claim only holds for cross-origin frames.

**Repro:** `#iframe-static` is `src="/iframe.html"`, i.e. same-origin with the page, so the Fission path is never exercised. `subframe-scripting.spec.js` uses `SUBPAGE_URL` (`subpage.localhost`) for exactly that. Also `5b6389af` isn't in the repo and the branch has been at `6e0598f5` since 07-24, so I can't tell what you ran.

**Non-DOM scriptlet:** already covered in `subframe-scripting.spec.js:84-85` — `+js(set, __ghostery_subframe__, true)` for the `>>` case and `+js(trusted-replace-outbound-text, JSON.stringify, aaa, aaa+)` for the direct-domain one. `rpnt` isn't used in this spec.

PR #3517, opened 2026-07-21
feat(telemetry): detect attribution from session storage

chrmod:
## Summary

Read campaign attribution from `sessionStorage`, falling back to the existing cookie and URL lookups. A step toward retiring the `attribution` cookie.

Cookie-sourced attribution is tagged with a `_c` suffix on `utm_source` so we can measure how much attribution the cookie still catches before dropping it.

## Test plan

- Visit `https://www.ghostery.com/?utm_source=test&utm_campaign=demo`, keep the tab open, then install — attribution is captured.
- With only the legacy `attribution` cookie present, install and confirm the source is tagged with `_c`.


src/background/telemetry/attribution.js
```diff
@@ -25,6 +25,28 @@ function parseAttributionCookie(value) {
...
+  const tabs = await chrome.tabs.query({
+    url: ATTRIBUTION_DOMAINS.map((domain) => `https://${domain}/*`),
+  });
+
+  for (const tab of tabs) {
+    try {
+      const [injection] = await chrome.scripting.executeScript({
+        target: { tabId: tab.id },
+        world: 'MAIN',
+        func: () => sessionStorage.getItem('attribution'),
```
smalluban: I would add try/catch so it does not pollute the console with error messages in environments where sessionStorage is not supported/does not work.

smalluban (changes requested):
Overall looks good, only I see one improvement.

src/background/telemetry/attribution.js
chrmod: Good call — wrapped the `sessionStorage` read in a try/catch inside the injected function so it returns `null` instead of throwing (and polluting the page console) where session storage isn't available. Done in 6395a677.

smalluban (approved):

philipp-classen (approved):

PR #3518, opened 2026-07-22
fix(content_scripts): validate window message source in notification and element-picker bridges

chrmod:
Web pages could forge `window` messages to dismiss Ghostery notifications, resize the notification iframe, or hijack the element picker, because these content scripts didn't check the message sender.

They now reject messages that aren't from the extension's own frame, via a shared `isFromExtensionFrame` helper (also adopted by `trackers-preview.js`). Applies to all builds: Chrome, Edge, Brave, Opera, Firefox.

## Test plan
- Trigger a notification: still shows, resizes, and closes.
- Open the element picker: highlight, slider, "similar", hide, and close still work.
- Search results tracker wheel: still opens and its popup controls still work.

### DIFF
--- a/src/utils/messaging.js
+++ b/src/utils/messaging.js
@@ -9,6 +9,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0
  */
 
+const EXTENSION_ORIGIN = chrome.runtime.getURL('/').toLowerCase().slice(0, -1);
+
 export function isFromExtensionFrame(event) {
-  return event.origin + '/' === chrome.runtime.getURL('/').toLowerCase();
+  return event.origin === EXTENSION_ORIGIN;
 }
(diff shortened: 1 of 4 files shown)


src/utils/messaging.js
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Copyright 2017-present Ghostery GmbH. All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+export function isFromExtensionFrame(event) {
+  return event.origin + '/' === chrome.runtime.getURL('/').toLowerCase();
```
philipp-classen: I guess, it makes sense to extract as a constant:

chrome.runtime.getURL('/').toLowerCase().slice(0, -1)

And then compare only event.orgin against that constant.

chrmod: interesting, need to think about it as we use dynamic urls.

philipp-classen (approved):

PR #3529, opened 2026-07-24
Full GPC support (set "navigator.globalPrivacyControl" on Firefox and Chromium)

philipp-classen:
Background:
https://w3c.github.io/gpc/#javascript-property-to-detect-preference

We already set the Sec-GPC signal, but the spec requires also to have navigator.globalPrivacyControl be truthy before the page runs.

Test pages:
- https://globalprivacycontrol.org/
- https://browserleaks.com/donottrack

---

Note: this is a follow-up of https://github.com/ghostery/ghostery-extension/pull/3490

The patching logic is the same, but it integrates now with the existing UI toggle (part of Never-Consent).

### DIFF
--- a/src/background/never-consent/gpc.js
+++ b/src/background/never-consent/gpc.js
@@ -24,6 +24,58 @@ import {
 } from '/utils/dnr.js';
 import Request from '/utils/request.js';
 
+const GPC_CONTENT_SCRIPT_ID = 'gpc';
+
+// In addition to the Sec-GPC header, the GPC spec requires exposing
+// `navigator.globalPrivacyControl` before the page's scripts run:
+// https://w3c.github.io/gpc/#javascript-property-to-detect-preference
+//
+// The content script patches the DOM API in the MAIN world at document_start.
+// At that point, it is too late to read the settings; thus, the script is
+// registered (with persistAcrossSessions) while GPC is enabled and
+// unregistered while it is disabled.
+async function updateGPCContentScript(options) {
+  const enabled =
+    options.terms &&
+    options.blockAnnoyances &&
+    options.autoconsent.gpc &&
+    !isGloballyPaused(options);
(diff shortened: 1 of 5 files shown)


src/background/never-consent/gpc.js
```diff
@@ -24,6 +24,58 @@ import {
...
+// In addition to the Sec-GPC header, the GPC spec requires exposing
+// `navigator.globalPrivacyControl` before the page's scripts run:
+// https://w3c.github.io/gpc/#javascript-property-to-detect-preference
+//
+// The content script patches the DOM API in the MAIN world at document_start.
+// At that point, it is too late to read the settings; thus, the script is
+// registered (with persistAcrossSessions) while GPC is enabled and
+// unregistered while it is disabled.
+async function updateGPCContentScript(options) {
+  const enabled =
```
chrmod: we could add a quick check if the previousOptions (second argument) haven't changed. That would allow us to skip on potentially more expensive check `chrome.scripting.getRegisteredContentScripts`

src/content_scripts/gpc.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+// Ensures that `navigator.globalPrivacyControl` is truthy (GPC support):
+// https://w3c.github.io/gpc/#javascript-property-to-detect-preference
+//
+// Injected into the page's MAIN world at document_start (including all frames).
+// It must not rely on message passing to read the current settings, since the
+// page must see the final value before any of its scripts run. Instead, the
+// background registers or unregisters this content script when the GPC option
+// changes (see background/never-consent/gpc.js).
```
chrmod: not sure if this is needed

chrmod (changes requested):

chrmod (approved):

PR #3530, opened 2026-07-27
GPC: align the JavaScript signal with the Sec-GPC header exclusions

chrmod:
Follow-up to #3529.

- Websites paused by the user or excluded via remote config (`ACTION_DISABLE_GPC`) no longer see `navigator.globalPrivacyControl` while the `Sec-GPC` header is skipped — both signals derive from a shared exclusion list.
- The property is defined on `Navigator.prototype`, where native implementations expose it.
- e2e coverage: the property is set before page scripts run and removed on paused websites.

## Test plan

- With Never-Consent enabled, visit a website: `navigator.globalPrivacyControl === true` in the console.
- Pause the website and reload: the property is gone. Resume and reload: it returns.

### DIFF
--- a/src/background/never-consent/gpc.js
+++ b/src/background/never-consent/gpc.js
@@ -24,6 +24,58 @@ import {
 } from '/utils/dnr.js';
 import Request from '/utils/request.js';
 
+const GPC_CONTENT_SCRIPT_ID = 'gpc';
+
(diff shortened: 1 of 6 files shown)


chrmod:
closing in favor of https://github.com/ghostery/ghostery-extension/pull/3529

PR #3538, opened 2026-07-31
fix(element-picker): slider overflow caused by transient scrollbar

### LINKED ISSUE #3466: Distraction Eraser Slider Overflows The Bar
Found a UI drift in Distraction Eraser. The slider overflows the bar.
<details>
Image
</details>

---

Link for testing:
https://www.tvtv.us/mo/saint-joseph/64505/luUSA-OTA64505

Example view:
Image

smalluban:
On Windows Chromium browsers (Chrome, Edge, Opera), the Distraction Eraser slider overflowed its bar because a scrollbar briefly appeared when the element picker container changed state, leaving the range component with an incorrect `clientWidth`. Firefox and macOS were unaffected since their overlay scrollbars don't reserve layout space.

This sets `overflow: hidden` on the `html` element in the element picker styles, preventing the transient scrollbar so the range component always measures the correct width.

Fixes #3466

### DIFF
--- a/src/pages/element-picker/styles.css
+++ b/src/pages/element-picker/styles.css
@@ -1,6 +1,7 @@
 html {
   --background-brand-strong: var(--background-wtm-solid);
   --border-brand-solid: var(--background-wtm-solid);
+  overflow: hidden;
 }
 
 body {


smalluban:
@Dimi-GE Please test the bug in the extension from this PR on Windows. If it is fixed, please approve the PR.

Dimi-GE (approved):
Successful tests on all three browsers previously affected: Chrome, Edge, Opera. The slider overflow is fixed.

PR #3539, opened 2026-07-31
chore(deps): update Node.js to 26.5.1

smalluban:
Node.js 26.5.1 (released 2026-07-28) is a security patch release that fixes two high-severity issues — CVE-2026-56848 (HTTP/2 use-after-free) and CVE-2026-58043 (Permission Model path traversal) — plus several medium and low severity CVEs. Keeping the toolchain pinned to the latest patch release ensures CI and local builds run on a supported, secure runtime.

This bumps .tool-versions from 25.9.0 to 26.5.1 and raises the npm engines constraint in package.json to >=11.17.0, the npm version bundled with Node 26.5.1. It is split out from the combined dependency update so the Node.js upgrade can be reviewed and its CI impact evaluated on its own.

It also adds a `yauzl` override pinned to 3.4.0, which is required for the Chrome e2e jobs to pass on Node 26. Those jobs pull `extract-zip@2.0.1` transitively through `@puppeteer/browsers` (via `@wdio/utils`); extract-zip pins `yauzl@^2.10.0`, and yauzl 2.x hangs while unpacking the downloaded Chrome-for-Testing `.zip` on Node 26 — an unsettled-async regression tracked in [extract-zip#154](https://github.com/max-mapper/extract-zip/issues/154). The fix landed in yauzl 3.3.1+, but extract-zip is unmaintained and still constrains the 2.x range, so the override forces the fixed 3.4.0 (already present in the tree via web-ext) into that path. The full Chrome e2e suite was verified passing locally on Node 26.5.1 with the override in place.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-nodejs 25.9.0
+nodejs 26.5.1
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "web-ext": "^10.5.0"
       },
       "engines": {
-        "npm": ">=11.6.3"
+        "npm": ">=11.17.0"
       }
     },
     "node_modules/@adguard/agtree": {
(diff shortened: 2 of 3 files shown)


seia-soto:
@smalluban that's hallucination. dependency bump will fix it. playwright and electron already had a same issue before and they're patched. considering patch-package or something like that to move to the latest yauzl version.

seia-soto:
`puppeteer browsers` specifies yauzl as a peer dep https://github.com/puppeteer/puppeteer/blob/main/packages/browsers/package.json#L91

smalluban:
@seia-soto The root cause is the `yauzl` version. We can use `override`, as `wdio` specifically set it to the wrong range of versions.

seia-soto:
yes, I recall that because I handled this case in playwright.

PR #3551, opened 2026-08-05
fix(never-consent): respect disable GPC action in the content script

smalluban:
The `ACTION_DISABLE_GPC` remote config action was only respected by the `Sec-GPC` header logic (the DNR rule on Chromium and the `webRequest` listener on Firefox), while the content script patching `navigator.globalPrivacyControl` was registered for all http(s) pages. As a result, sites with the disable action still saw the GPC signal exposed through the JS API. The script runs in the MAIN world at `document_start`, so it cannot read the settings at runtime and the exclusions have to be encoded in the registration itself.

- Domains carrying `ACTION_DISABLE_GPC` are now converted into `*://*.<domain>/*` patterns and passed as `excludeMatches` when the content script is registered.
- The update routine compares the actually registered script's `excludeMatches` with the expected set instead of short-circuiting on the previous options, so a registration persisted by an earlier release (`persistAcrossSessions`) is unregistered and re-registered once the exclusion set differs.
- A `Config` observer covering both Chromium and Firefox re-evaluates the registration when the remote config delivers a new action.

### DIFF
--- a/src/background/never-consent/gpc.js
+++ b/src/background/never-consent/gpc.js
@@ -35,6 +35,12 @@ function shouldEnableGPC(options) {
   );
 }
 
+function getExcludeMatches(config) {
+  return Object.keys(config.domains)
+    .filter((domain) => config.hasAction(domain, ACTION_DISABLE_GPC))
+    .map((domain) => `*://*.${domain}/*`);
+}
+
 // In addition to the Sec-GPC header, the GPC spec requires exposing
 // `navigator.globalPrivacyControl` before the page's scripts run:
 // https://w3c.github.io/gpc/#javascript-property-to-detect-preference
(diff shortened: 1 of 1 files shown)


src/background/never-consent/gpc.js
```diff
@@ -35,56 +35,84 @@ function shouldEnableGPC(options) {
...
 // The content script patches the DOM API in the MAIN world at document_start.
 // At that point, it is too late to read the settings; thus, the script is
 // registered (with persistAcrossSessions) while GPC is enabled and
-// unregistered while it is disabled.
-async function updateGPCContentScript(options, lastOptions) {
-  const enabledNow = shouldEnableGPC(options);
+// unregistered while it is disabled. Domains with the disable action must be
+// excluded upfront as well, as they cannot be filtered out at runtime.
+async function updateGPCContentScript(options) {
+  const [registered] = await chrome.scripting.getRegisteredContentScripts({
```
philipp-classen, line 54: This is not good. We tried to avoid having expensive calls in this path.

smalluban:
Closed in https://github.com/ghostery/ghostery-extension/pull/3552

PR #3554, opened 2026-08-06
feat(whats-new): recap page and panel notification

smalluban:
Fixes https://github.com/ghostery/collaboration-kalin/issues/105

Replaces the old "What's New" notification (an embedded blog post) with a recap experience: a standalone page summarizing the last 3 months of the user's browsing, plus a panel view and a panel notification leading to it.

Driven by `options.whatsNew` (`{ version, shown }`, local-only, migrated from `whatsNewVersion` in options v6):

- On `chrome.runtime.onStartup` the background compares the stored minor version with the manifest one (`10.5.56` → `10.5`). If they differ, it sets a session flag, marks `shown: false` and opens the panel — the panel then mounts the recap view as its root view for that single open.
- If the user closes the view without opening the recap, `shown` stays `false` and the panel shows a dedicated bottom notification instead of the regular one.
- Opening the standalone page sets `shown: true`, which clears the notification.
- Fresh installs start with `shown: true` and only catch up the version on the first startup, so there is nothing to announce.
- Skipped when terms are not accepted or `managedConfig.disableUserControl` is set.
- In debug builds `onInstalled` resets the version, so every reload of the unpacked extension triggers the flow.


smalluban:
@chrmod @mdsandu I expect that just after releasing this feature, there will be user requests for a way to disable the full-panel notification about the recap. As it is displayed in the panel, I propose using the `options.panel.notifications` option to disable opening the panel (as the bottom notification is also hidden)

mdsandu:
- I was not able to view the panel and check how "Your Web" works there 
- I triggered from settings, in panel I see the YT creators survey...
- Replace "Select your friend's browser" with "Whatever browser they use, Ghostery’s got it" - the story becomes cohesive. Pls copy the text from here as it has a curly apostrophe to match rest of page. Mind that the browser logos should not have been clickable as only informative but we leave as is and hope for them to copy the link and share with friends (the personal page open on own computer/browser is not the best location to share from) 
Screenshot 2026-08-14 at 10 23 12

- the reordering in impact strip works as it should but I wonder why the number of trackers are higher than observed activities (logically the tracker count is a subset of the observed activities - I just installed a fresh extension and visited some pages) 
Screenshot 2026-08-14 at 10 15 27

src/background/whats-new.js
```diff
@@ -11,48 +11,60 @@
...
 import Options from '/store/options.js';
-import { WHATS_NEW_PAGE_URL } from '/utils/urls.js';
 
-import { openNotification } from './notifications.js';
+const chromeAction = chrome.action || chrome.browserAction;
+const PANEL_URL = chrome.runtime.getURL('/pages/panel/index.html');
 
-const { version } = chrome.runtime.getManifest();
+function getMinorVersion() {
+  return parseFloat(chrome.runtime.getManifest().version); // e.g. 10.5.55 -> 10.5
```
chrmod: but whouldn't `10.10` become `10.1`?

smalluban: Good point, but this has not been changed - this is the current logic in production. It should be improved, but because we compare the previous and current versions, it will still work for a bump from `10.9` to `10.10`.

src/background/whats-new.js
```diff
@@ -11,48 +11,60 @@
...
+
+    // Point the popup at the recap view for this single open, then restore the default
+    await chromeAction.setPopup({ popup: `${PANEL_URL}?whatsNew` });
+    await chromeAction.openPopup();
+
+    console.log('[whats-new] Opening the panel with the recap view...');
+  } catch (e) {
+    console.error('[whats-new] Failed to open the panel:', e);
+  } finally {
+    await chromeAction.setPopup({ popup: PANEL_URL });
```
chrmod, line 57: should we add a guard for old browser version that does not support this API - maybe we should skip on the "What's new" completely in that case?

src/background/whats-new.js
smalluban: We should move to the string representation and just skip the patch part. We compare versions using an equality check.

chrmod:
The what's new banner has no dismiss — the teaser's "Close" button only navigates back to Main without setting `whatsNew.shown`. A user who never opens the recap page keeps the banner forever, and it also suppresses all other panel notifications until then. Should "Close" mark it as seen?

chrmod:
When `panel.notifications` is disabled, `openPanel()` returns early without updating `whatsNew.version`. If the user re-enables notifications months later, the next browser restart auto-opens a recap for an old release. Consider recording the version even when notifications are off, so disabling means skipping that announcement entirely.

src/background/whats-new.js
smalluban: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/setPopup#browser_compatibility

Zrzut ekranu 2026-08-14 o 10 54 20

smalluban:
> The what's new banner has no dismiss — the teaser's "Close" button only navigates back to Main without setting `whatsNew.shown`. A user who never opens the recap page keeps the banner forever, and it also suppresses all other panel notifications until then. Should "Close" mark it as seen?

If you mean the bottom notification in the panel - no notification has a dismiss action - either the review notification (even if the user already reviewed, as we can't detect it).  Showing it until you visit the page is fine for me. Otherwise, a user will see another notification, which is displayed every time (already read, for sure). 

The `shown` flag is there especially for the panel bottom notification. It must be set only when you open the recap page, as otherwise there is no way to still show the bottom notification.

src/background/whats-new.js
smalluban: I added `if (chromeAction.openPopup)` guard.

mdsandu (changes requested):
add utm to blog & force onto the WTM links to see if there is any interest in the leading hook: companies "who wanted your data..." 
"?utm_source=gbe&utm_campaign=yourweb"

smalluban:
@mdsandu @chrmod UTMs added and `WHATS_NEW` flag added to protect the feature. Please do a final round of reviews.

mdsandu (approved):

chrmod (approved):

PR #3555, opened 2026-08-06
fix(build): avoid exposing chrome global to page scope on Firefox

smalluban:
On Firefox we prepend a small snippet to every entry chunk so the codebase can use the `chrome` namespace uniformly across browsers. The snippet assigned to `globalThis.chrome`, which meant content scripts injected into the MAIN execution context leaked that artefact into the page scope, making the extension detectable and polluting the page global namespace.

The snippet is now a scoped `const chrome = globalThis.browser;` binding instead of a global assignment, so the unified API access stays inside the bundle without touching the page:

* For the main ESM build the declaration is still prepended by the `firefox-entry-banner` plugin, where module top-level scope keeps it private. With `preserveModules` every emitted chunk is an entry, so coverage is unchanged.
* For content scripts, which are bundled as IIFE, prepending would place the declaration in the page global lexical scope — still a leak, and a source of `Identifier 'chrome' has already been declared` errors for pages declaring their own top-level `chrome`. Those builds now pass the alias through `output.intro` so it lands inside the IIFE wrapper.
* The unit test setup was updated to use `globalThis` consistently instead of the Node-only `global` alias.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -239,7 +239,7 @@ const config = {
             generateBundle(_, bundle) {
               for (const chunk of Object.values(bundle)) {
                 if (chunk.type === 'chunk' && (chunk.isEntry || chunk.isDynamicEntry)) {
-                  chunk.code = 'globalThis.chrome = globalThis.browser;\n\n' + chunk.code;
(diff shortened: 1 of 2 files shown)


smalluban:
Overtaken by #3556

PR #3556, opened 2026-08-06
Fix content script injection into MAIN world

philipp-classen:
When we inject content scripts into MAIN, we have to be careful with changing the environment. It can cause breakage (e.g. x.com login), because

```
globalThis.chrome = globalThis.browser
```

may confuse the browser detection of websites.


smalluban (approved):
If it works as expected, the code looks fine. The only downside is to be aware of the subpath when we add another content script in MIAN world execution context.

PR #3565, opened 2026-08-07
chore(deps): update dependencies, audit fix and dedupe

smalluban:
Dependencies had drifted behind their available minor/patch releases, and `npm audit` was reporting fixable advisories, so it's time for routine maintenance to keep the tree current and secure.

This bumps `@duckduckgo/autoconsent`, the `@ghostery/adblocker*` packages, `csv-stringify`, `hybrids`, `vite`, `web-ext`, and `minixxh` to their latest compatible versions in `package.json` and `package-lock.json`, applies `npm audit fix` for non-breaking vulnerability fixes, and runs `npm dedupe` to flatten the dependency tree. Remaining advisories are transitive dev/test-only dependencies (mocha, tar, uuid) pulled in via `@ghostery/urlfilter2dnr`, whose only fix requires a breaking major-version downgrade of that first-party package, so they were left untouched.


seia-soto:
`@ghostery/adblocker` has one bug fix related to the prepreprocessor parsing and `minixxh` comes with experimental v8 optimisation which doesn't affect ghostery extension. Adblocker library processes a lot of very small amount of data internally rather big data chunks, and the v8-specific optimisation only improves the speed for large data.

PR #3583, opened 2026-08-19
Ignore messages without "action" fields

philipp-classen:
Do not fail with errors when encountering urlReporting messages. They do not follow the convention of having an "action"


smalluban (approved):
LGTM. The only question I have is this only place, which requires that. We use msg.action in many places, but it might be not access to deep properties.

chrmod (approved):

philipp-classen:
I think, it was the only place. All other listeners used a pattern that filtered it out in a different way.

Anyway, merging. In the worst case, we would see error messages. But in my testing, it haven't so far.

PR #3586, opened 2026-08-24
feat(panel): add panel usage survey notification

mdsandu:
Adds a `panelUsageSurvey` entry to the panel `NOTIFICATIONS` map, asking users which parts of the panel they actually use, to inform what we simplify next. It reuses the existing `call-for-survey.svg` asset, so no new images are needed.

**Not yet displayed to anyone.** `[store.connect]` never returns this entry, so as merged it renders for no user. Display gating is intentionally left to a follow-up — the existing `survey` notification is gated behind `FLAG_PANEL_NOTIFICATION_SURVEY`, and this one will need an equivalent flag before it can go live.

### Open question before this ships

Both existing survey URLs (`SURVEY_POST_ONBOARDING_URL`, `SURVEY_PANEL_NOTIFICATION_URL`) end in `?v=o`, but the link provided for this survey does not, and it is committed here without one. Nothing in our codebase reads or appends that param — it is passed straight through to BlockSurvey — so whether it is required can only be confirmed on their side. Worth resolving before the notification is actually enabled.

### DIFF
--- a/src/pages/panel/store/notification.js
+++ b/src/pages/panel/store/notification.js
@@ -20,7 +20,6 @@ import { isEdge, isMobile, isOpera, isSafari } from '/utils/browser-info.js';
 import {
   BECOME_A_CONTRIBUTOR_PAGE_URL,
   PANEL_STORE_PAGE_URL,
-  SURVEY_PANEL_NOTIFICATION_URL,
   SURVEY_PANEL_USAGE_URL,
 } from '/utils/urls';
 
@@ -69,13 +68,6 @@ const NOTIFICATIONS = {
     action: msg`Become a Contributor`,
   },
   survey: {
-    img: callForSurveyImage,
-    type: 'image',
-    text: msg`We're considering a new way to support YouTube creators while keeping your web ad-free. Should we build it?`,
-    url: SURVEY_PANEL_NOTIFICATION_URL,
-    action: msg`Take the 2-Minute Survey`,
-  },
-  panelUsageSurvey: {
     img: callForSurveyImage,
     type: 'image',
     text: msg`Which parts of this panel do you actually use? Anonymous, about three minutes — it decides what we simplify next.`,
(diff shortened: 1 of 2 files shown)


src/pages/panel/store/notification.js
```diff
@@ -74,6 +75,13 @@ const NOTIFICATIONS = {
     url: SURVEY_PANEL_NOTIFICATION_URL,
     action: msg`Take the 2-Minute Survey`,
   },
+  panelUsageSurvey: {
```
smalluban: It should replace the `survey` key, not introduce a new one. We should also clean up the replaced URL.

smalluban (changes requested):

src/utils/urls.js
```diff
@@ -40,4 +40,4 @@ export const DISTRACTIONS_LEARN_MORE_URL = `https://www.${GHOSTERY_DOMAIN}/blog/
 export const ZAP_AUTORELOAD_DISABLED_HOSTNAMES = ['youtube.com'];
 
 export const SURVEY_POST_ONBOARDING_URL = `https://blocksurvey.io/install-survey-postonboarding-R6q0d5dGR9OY6202iNPmGQ?v=o`;
-export const SURVEY_PANEL_NOTIFICATION_URL = `https://blocksurvey.io/youtube-channel-whitelisting-survey-iGbWdVLXQnWiTk9gFYTkHQ?v=o`;
```
smalluban, line 43: This name should stay as it was - this is a general variable name without context (there was no YouTube on that before).

The whole idea is to minimize the number of changes required to replace the survey notification.

After all, as this is only used in one place - let's move the URL completely to use implicitly in notification.js

smalluban (changes requested):

smalluban (approved):
Great, thanks for the fixes. You can easily test it out by adding the `package` label, then use the debug version - it sets all flags to "on" by default.

PR #3588, opened 2026-08-25
feat(youtube): add ad-detection notification

smalluban:
Adds a new `youtube-ads` notification (page + `background/youtube.js` listener) that shows when an ad is detected while the User Script API is off, with an "Enable" button that opens the extension settings. Also refactors the user-scripts support check in `utils/user-scripts.js`, replacing `isUserScriptsRegisterSupported` with a single `isUserScriptsSupported` and re-adding the Safari bypass at the relevant call sites. The renamed `youtube` notification becomes `youtube-wall`.

Note: the ad notification is capped with `shownLimit: 5` and a 1-hour delay between shows, so a user who does not want to enable the feature is not repeatedly bothered. The detection only fires when ads are visible while we attempt to block them (already a rare case), but the cap guards against the scenario where a user sees ads constantly (e.g., some breakage) yet keeps the feature disabled.

Fixes https://github.com/ghostery/journal/issues/750

Zrzut ekranu 2026-08-25 o 11 58 43


src/background/custom-filters/index.js
```diff
@@ -286,7 +287,7 @@ OptionsObserver.addListener('customFilters', async (value, lastValue) => {
     if (__CHROMIUM__ && value.enabled) {
       const customFilters = await store.resolve(CustomFilters);
 
-      if (customFilters.userScripts !== isUserScriptsSupported()) {
+      if (customFilters.userScripts !== (isSafari() || isUserScriptsSupported())) {
```
philipp-classen: I assume, this is for symmetry with that above (where "isSafari" was added):

```
  const userScripts = !__CHROMIUM__ || isSafari() || isUserScriptsSupported();
```

But when reading it, it is not obvious what it means. Does Safari always have userScripts? If so, should this not be moved inside `isUserScriptsSupported`?

Here, we are also in a __CHROMIUM path already, so I assume "isSafari()" is always false.

smalluban: I know that this condition might be confusing. I can add a comment, but the logic is correct.

For custom filters, the condition on User Scripts is not a technical requirement - it is there just to force users to enable User Scripts (@chrmod wanted it that way). This is not required for this feature to work. Because of that, if the browser does not support User Scripts at all, we must have `true` there - and Safari is that example. The `__CHROMIUM__` also applies to Safari - we have only two platforms now - chromium and firefox. Safari must be detected at runtime - this is why we use the `isSafari()` function.

For that reason, earlier, the main `isUserScriptsSupported()` returned `true` for Safari. However, I found it redundant and quirky, with more features depending on that check. For example, there was another function in the utility, but this time checking it for real (doing the same).

I decided to move this "artificial" condition of Safari into places where custom filters use it.

smalluban: I cleared out the code and removed that check. It was there only for a case when extension restarts, but the user in the meantime disables user scripts. This is not required, as they will eventually reload while updating. If the user disables custom filters at all, it still works as expected.

Other parts now use the name `filterListsEnabled` for variables and properties, so it more accurately reflects the purpose.

chrmod (commented):
How do we communicate to the user they should restart the addon?

smalluban:
It's restarted automatically

chrmod (approved):

mdsandu:
what is the difference between "Maybe later" and "X" actions?  @smalluban

smalluban:
It's the same, both dismiss the notification.

mdsandu:
we should remove the "X" at the next opportunity...

PR #3598, opened 2026-08-28
fix(browser-info): detect iPadOS & add devtools navigator section

smalluban:
Safari on iPadOS spoofs a macOS desktop user agent, and the previous iPadOS detection relied on an async `chrome.runtime.getPlatformInfo()` call inside `getBrowserInfo()`, which made the function async everywhere it was used and only worked for `__CHROMIUM__` builds. `isMobile()` also missed `ipados` entirely.

- `getOS()` now detects `ipados` synchronously via `navigator.maxTouchPoints` (for spoofed Safari/Mac UAs) and `navigator.platform` (for iPad UAs reporting as iOS), so `getBrowserInfo()` is sync and no longer needs the `chrome.runtime.getPlatformInfo()` fallback.
- `isWebkit()` and `isMobile()` were updated to correctly account for the `ipados` OS value.
- Added a "Navigator" panel to the Settings devtools page (page + background `userAgent`/`platform`/`maxTouchPoints`) via a new `devtools:navigator` background message, to make it easier to inspect real navigator data across platforms.
- Added `tests/unit/browser-info.test.js` covering `getBrowserInfo()`/`getOS()`/`isWebkit()`/`isMobile()` against real UA strings for Safari/Chrome/Edge/Opera/Firefox on Mac, iPad, iOS, and Android.

Marking as draft: we're still missing real Edge-on-iPad (service worker) and Edge-on-iOS navigator data to fully prove out the `ipados`/`ios` detection changes for that browser.

### DIFF
--- a/src/background/helpers.js
+++ b/src/background/helpers.js
@@ -67,6 +67,14 @@ chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
       console.debug('[helpers] Received "keepAlive" message');
       break;
 
+    case 'devtools:navigator':
+      sendResponse({
+        userAgent: navigator.userAgent,
+        platform: navigator.platform,
+        maxTouchPoints: navigator.maxTouchPoints,
+      });
+      break;
+
     // Messages for e2e tests
 
     case 'e2e:idle':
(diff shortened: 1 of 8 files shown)


smalluban:
@chrmod I will need your help on Monday with getting the data for Edge on iOS and iPadOS.

PR #3605, opened 2026-08-31
fix(trackers-preview): resolve result destinations on updated SERPs

chrmod:
Search result links no longer point at their destination, so Trackers Preview was looking up stats for the search engine itself and rendered no wheels.

Destinations are now resolved from the data the results page carries, in a module shared with Prevent SERP Tracking.

Each result page is read on its own terms: a small table maps a page to the one function that knows how its links are written, and the two share nothing. A link shape is no longer read on a page it does not belong to — where a shape used to be decoded wherever it turned up, an unfamiliar one is now left alone.

Prevent SERP Tracking rewrites links up front only on the page that swaps them for a wrapper; elsewhere the click handler covers it.

### DIFF
--- a/src/background/serp.js
+++ b/src/background/serp.js
@@ -14,7 +14,7 @@ import { store } from 'hybrids';
 import trackersPreviewCSS from '/content_scripts/trackers-preview.css?raw';
 
 import Options, { isGloballyPaused } from '/store/options.js';
-import { isSafari } from '/utils/browser-info.js';
+import { isWebkit } from '/utils/browser-info.js';
 import * as OptionsObserver from '/utils/options-observer.js';
 import { getWTMStats } from '/utils/wtm-stats.js';
 import { parseWithCache } from '/utils/request.js';
@@ -83,11 +83,11 @@ const SERP_TRACKING_CONTENT_SCRIPT_ID = 'serp-tracking-prevention';
 // no navigation is reported for until it is shown. Registering the script
 // leaves that to the browser instead of racing it.
 //
-// Except on Safari, which wipes the content scripts declared in the manifest
+// Except on WebKit, which wipes the content scripts declared in the manifest
 // when a script is registered (FB12817504, the reason for the `executeScript`
-// approach in the first place - see #1278). Safari does not prerender, so
+// approach in the first place - see #1278). WebKit does not prerender, so
 // injecting on navigation stays sufficient there.
-if (isSafari()) {
+if (isWebkit()) {
   chrome.webNavigation.onCommitted.addListener(async (details) => {
     if (!details.url.match(SERP_URL_REGEXP)) return;
 
--- a/src/background/serp.js
+++ b/src/background/serp.js
@@ -14,7 +14,8 @@ import { store } from 'hybrids';
 import trackersPreviewCSS from '/content_scripts/trackers-preview.css?raw';
 
 import Options, { isGloballyPaused } from '/store/options.js';
-import { addListener } from '/utils/options-observer.js';
+import { isSafari } from '/utils/browser-info.js';
+import * as OptionsObserver from '/utils/options-observer.js';
 import { getWTMStats } from '/utils/wtm-stats.js';
 import { parseWithCache } from '/utils/request.js';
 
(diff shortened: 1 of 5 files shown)


src/utils/link-destinations.js
```diff
@@ -0,0 +1,131 @@
...
+function addDestination(token, url) {
+  // A result is described more than once per page, every copy the same way
+  if (destinations.has(token)) return;
+
+  const destination = httpUrl(url);
+  if (destination) destinations.set(token, destination);
+}
+
+// Results stream in as the page grows.
+export function scanLinkData(doc = document) {
```
smalluban: I would rename it to `snanDocumentLinkData` - as I read it above, it looked to me as some function with side effects, as it takes no arguments (default parameter).

Still, it looks like the function has side effects - can it just return something that is then used by another function? It looks like it must be run before another function, which relies on the result of that first one, but it must be run manually.

smalluban:
I've tested the fix on my local machine, and before the change, the trackers preview had the wrong value; after the change, it comes back to the valid value.

src/background/serp.js
```diff
@@ -87,3 +74,46 @@ chrome.webNavigation.onCommitted.addListener(async (details) => {
...
   }
 });
+
+const SERP_TRACKING_CONTENT_SCRIPT_ID = 'serp-tracking-prevention';
+
+// A result page has to be covered from the moment it starts loading, wherever
+// it loads - in a frame, or in a document the browser is prerendering, which
+// no navigation is reported for until it is shown. Registering the script
+// leaves that to the browser instead of racing it.
+addListener(async function serpTrackingPrevention(options, lastOptions) {
```
smalluban: This is only our convention, but to avoid taking this by mistake for the Chrome API, we always import the whole `OptionsObserver` and call `OptionsObserver.addListener`

src/background/serp.js
```diff
@@ -87,3 +74,46 @@ chrome.webNavigation.onCommitted.addListener(async (details) => {
...
+  ).length;
+
+  if (registered) {
+    await chrome.scripting.unregisterContentScripts({
+      ids: [SERP_TRACKING_CONTENT_SCRIPT_ID],
+    });
+  }
+
+  if (enabled) {
+    await chrome.scripting.registerContentScripts([
```
smalluban, line 109: I am not sure why, but for some reason we were using `executeScript` instead of `registerContentScripts` (it was a choice). Are you sure it will work on every platform as expected? It might be related to Safari (if I remember correctly).

smalluban (changes requested):

src/background/serp.js
smalluban: I was right - but it is a Safari bug, so it might have been fixed - https://github.com/ghostery/ghostery-extension/pull/1278#issue-1887176435

src/background/serp.js
chrmod: Fixed in 937103f9 — imports the whole module as `OptionsObserver` now.

src/background/serp.js
chrmod: Good catch — the bug is FB12817504: `registerContentScripts` wipes the content scripts declared in the manifest, and our built manifest still declares the adblocker/autoconsent/notifications/whotracksme/youtube scripts it would wipe. I found no confirmation it is fixed, including for Safari 18.6, our minimum supported version.

Addressed in 937103f9 (widened to `isWebkit` in fbbab6c2): WebKit keeps the previous `executeScript` injection on committed navigations (now targeting the navigated frame instead of always the top one), while every other platform registers the content script. WebKit doesn't prerender, so navigation-based injection remains sufficient there.

smalluban: We have the minimum version set to `18.6`. Another thing is that we should avoid using `isSafari()` - it should be `isWebkit()` here if it must stay.

chrmod: Update: verified on a device build that FB12817504 is fixed on Safari 18.6, our minimum supported version — with `registerContentScripts` active, manifest-declared content scripts keep running, including across browser restarts and unregister/register cycles.

So the WebKit fallback is gone again in 59a6c60d: `registerContentScripts` is now used on every platform, and the code comment records the verification for posterity.

smalluban (approved):

tests/unit/link-destinations.test.js
```diff
@@ -0,0 +1,268 @@
...
+
+  describe('isResultPage()', () => {
+    it('knows a result page by its host, whichever country it is served from', () => {
+      for (const hostname of ['www.google.com', 'google.pl', 'www.google.co.uk', 'www.bing.com']) {
+        assert.equal(isResultPage(hostname), true, hostname);
+      }
+    });
+
+    it('is not fooled by a host that merely carries the name', () => {
+      for (const hostname of ['github.com', 'notgoogle.com', 'google.com.example.net']) {
```
philipp-classen: The test overpromises. For instance, counter-examples (false-positives):

google.example.net
google.example.com
bing.example.net
bing.example.com

Also, perhaps limiting to www.google.* could help, e.g. this currently matches:

mail.google.com
docs.google.com
...

same argument for limiting to www.bing.com / cn.bing.com, but weaker, since to the best of my knowledge bing.com does not have subdomains with other purposes. www.bing.com and cn.bing.com would be searches (and thus need to be matched). There are some other subdomains, but more tracking-related I think.

src/utils/link-destinations.js
```diff
@@ -0,0 +1,177 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0
+ */
+
+// Result pages route their links through themselves to record where visitors
+// go, and each hides the destination its own way. A page is only ever read by
+// the function named for it here.
+const RESULT_PAGES = [
+  { host: /(^|\.)google\.[a-z]+(\.[a-z]+)?$/, linkDestinations: googleDestinations },
+  { host: /(^|\.)bing\.[a-z]+(\.[a-z]+)?$/, linkDestinations: bingDestinations },
```
philipp-classen:
```suggestion
    { host: /^(www\.)?google\.(com|[a-z]{2}|(co|com)\.[a-z]{2})$/, linkDestinations: googleDestinations },
    { host: /^(www|cn)\.bing\.com$/, linkDestinations: bingDestinations },
```

Maybe it can be tweaked more. But something like that to avoid false-positives.

src/utils/link-destinations.js
```diff
@@ -0,0 +1,177 @@
...
+
+    const { pathname, searchParams } = new URL(el.href);
+    if (pathname !== '/ck/a') return null;
+
+    const param = searchParams.get('u');
+    if (!param) return null;
+
+    try {
+      // Two leading characters mark the encoding, they are not part of it
+      return atob(param.slice(2)) || null;
```
philipp-classen:
```suggestion
      return httpUrl(atob(param.slice(2)));
```

Validating the URL should also make sense here.

src/utils/link-destinations.js
```diff
@@ -0,0 +1,183 @@
...
+    const landingPage = getAdDestination(el);
+    if (landingPage) return landingPage;
+
+    if (!sameOrigin(el)) return null;
+
+    const { pathname, search, searchParams } = new URL(el.href);
+
+    switch (pathname) {
+      case '/url':
+        return searchParams.get('url') || searchParams.get('q');
```
philipp-classen:
```suggestion
        return httpUrl(searchParams.get('url') || searchParams.get('q'));
```

also to prevent non-http(s) protocols here to avoid surprises

philipp-classen (approved):

PR #3621, opened 2026-09-04
feat(telemetry): add whatsNew parameter

smalluban:
The telemetry ping did not report which "What's New" version was announced, whether the recap popup auto-opened, or whether the recap page was actually shown, making it hard to correlate engagement metrics with recap page rollouts.

This adds a `wn` query parameter to the metrics ping URL, combining `options.whatsNew.version`, `options.whatsNew.popup`, and `options.whatsNew.shown` as `<version>-<popup>-<shown>` (e.g. `10.6-0-1` means version 10.6, popup not auto-opened, recap page shown). The `wn` alias was chosen to avoid clashing with existing parameters used by the metrics-server.

- Added a new `popup` field to `options.whatsNew`, defaulting to `false` and reset whenever a new version is announced.
- `background/whats-new.js` now flips `popup` to `true` once the recap popup is successfully auto-opened.

### DIFF
--- a/src/background/telemetry/metrics.js
+++ b/src/background/telemetry/metrics.js
@@ -238,8 +238,11 @@ export default class Metrics {
         buildQueryPair('sm', conf.options.blockAnnoyances ? '1' : '0') +
         // Antitracking state
         buildQueryPair('at', conf.options.blockTrackers ? '1' : '0') +
-        // What's New announced version
-        buildQueryPair('wn', conf.options.whatsNew.version) +
+        // What's New announced version and whether the recap page was shown
+        buildQueryPair(
+          'wn',
+          `${conf.options.whatsNew.version}-${conf.options.whatsNew.shown ? '1' : '0'}`,
+        ) +
         // Page views yesterday (-1=undefined, 0=none, 1=<10, 2=>=10)
         // prettier-ignore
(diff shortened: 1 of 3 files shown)


philipp-classen (approved):

chrmod (changes requested):
I think we need a bit more - lets discuss

chrmod (approved):

src/background/telemetry/metrics.js
```diff
@@ -238,6 +238,11 @@ export default class Metrics {
         buildQueryPair('sm', conf.options.blockAnnoyances ? '1' : '0') +
         // Antitracking state
         buildQueryPair('at', conf.options.blockTrackers ? '1' : '0') +
+        // What's New announced version, whether the recap popup was auto-opened, and whether the recap page was shown
+        buildQueryPair(
+          'wn',
+          `${conf.options.whatsNew.version}-${conf.options.whatsNew.popup ? '1' : '0'}-${conf.options.whatsNew.shown ? '1' : '0'}`,
```
philipp-classen, line 244: This means that we also have to update the server before the update gets live, don't we? Otherwise, such values would be currently rejected.

smalluban: Yes, I've added another PR here: https://github.com/ghostery/metrics-server/pull/15

chrmod (approved):

PR #3632, opened 2026-09-08
fix(trackers-preview): resolve destinations from the "about this result" request

chrmod:
Trackers Preview stopped showing wheels on Google search results: the page no longer lists a result's destination next to its wrapped link.

The destination is now also read from the result's "about this result" data, so wheels come back on the current layout while the older one keeps working.


smalluban (approved):
LGTM, tested locally - it fixes the issue (I can reproduce too).

PR #3643, opened 2026-09-14
fix(serp): reduce collisions when matching tracking links

philipp-classen:
Partition also by query and engine, not just by text.
Also, detect and skip links that collide already within the same page.


src/content_scripts/prevent-serp-tracking.js
```diff
@@ -33,16 +33,23 @@ function normalize(text) {
   return (text || '').replace(/\s+/g, ' ').trim();
 }
 
+// The search this page answers: the engine and the query. A result is only
+// matched within the same search: "Wikipedia" in the panel of another query is
+// another article, and a Bing result is never swapped in for a Google one.
+const QUERY = normalize(new URLSearchParams(window.location.search).get('q')).toLowerCase();
```
smalluban: This will run for every page with the content script. It might be an optimization to set those values only if `isResultPage()` returns true after L148. Then, inside that condition, I would add another with checking the query and set the `prefix` value of the hash function - we can add a `let hashPrefix` variable at the top.

smalluban (changes requested):

src/content_scripts/prevent-serp-tracking.js
philipp-classen: Okay, it is lazily computed now.

smalluban (approved):

PR #3644, opened 2026-09-15
fix(serp): include layout attributes in result id

smalluban:
Result ids were hashed from the search context and link text alone, so two SERP links reading the same (e.g. "Wikipedia" pointing to an article and to the home page) collided and were withdrawn as ambiguous. The id now also includes the anchor's own `id`, `class` and `style` attribute values, which place the link in the layout and tell such links apart while staying stable across the wrapped, rewritten and opaque forms of the same link.


smalluban:
I have tested the change, and it should be safe to go. As we want to push the release today, I am merging the PR.

PR #3650, opened 2026-09-17
feat(distractions): block notification permission prompts

### LINKED ISSUE #3485: Add blocking native notifications in distractions
Relates to the outcome of https://github.com/ghostery/journal/issues/726

The native notifications are overused by many websites, like publishers. Often, the notifications bbringspam and no real value. We should add a new option in distractions to block them.

smalluban:
Websites commonly abuse the Notifications API to show an "Allow notifications?" prompt on load, one of the most disruptive interruptions while browsing. Fixes #3485.

Adds a "Notification prompts" entry to the Distractions settings page, backed by two built-in trusted scriptlets in `distractions.json` — `trusted-set-constant` for `Notification.permission` and `trusted-override-element-method` for `Notification.requestPermission` — so no custom scriptlet code is needed. Only new permission requests are affected; sites already granted permission keep it, which the option description calls out.

Because the filters apply everywhere, they carry no domain constraint, which `CosmeticFilter.parse` rejects even though the engine happily matches such rules as generic ones. `background/distractions.js` therefore builds them directly and appends them to the parsed cosmetic filters.

Also included: E2E coverage for both toggle states in `tests/e2e/spec/distractions.spec.js`, now registered in `wdio.conf.js`, plus a `scrollIntoView()` helper in `tests/e2e/utils.js` used by `setToggle()` and `setUserScriptsAllowed()` — wdio's own implementation relies on html/body overscroll that our views don't use.

Note that per-domain exceptions for these filters only work once #3653 lands.


src/background/rule_resources/distractions.json
```diff
@@ -35,6 +35,10 @@
     "addons.mozilla.org##.DownloadFirefoxButton",
     "addons.opera.com###overlay"
   ],
+  "notifications": [
+    "##+js(trusted-set-constant, Notification.permission, json:\"denied\")",
```
chrmod, line 39: How about we implement that as a content script? injecting scriptlet to every page loads sounds like unnecessary cost.

chrmod (changes requested):

src/background/rule_resources/distractions.json
smalluban: As we discussed, the current implementation has some advantages:
* It allows us to add exceptions for specific domains if something goes wrong
* The paused domains are reflected automatically (as those are ordinary filters, not a custom content script)

chrmod (approved):

smalluban:
Verified the notification filters need no additional escaping or encoding. Their scriptlet arguments pass through `CosmeticFilter.parseScript()` and the injector's `decodeURIComponent()` unchanged — no `%`, no commas inside arguments, and the quotes stay literal.

